PR #21, opened 2024-07-02
ip: send an icmp message when TTL is expired

christophefontaine:
Instead of dropping silently the packet, forge and send back an ICMP message "Time To live exceeded".

### DIFF
--- a/modules/ip/datapath/gr_ip4_datapath.h
+++ b/modules/ip/datapath/gr_ip4_datapath.h
@@ -4,6 +4,7 @@
 #ifndef _GR_IP4_DATAPATH_H
 #define _GR_IP4_DATAPATH_H
 
+#include <gr_iface.h>
 #include <gr_ip4_control.h>
 #include <gr_mbuf.h>
 #include <gr_net_types.h>
@@ -14,7 +15,10 @@
 
 #include <stdint.h>
 
-GR_MBUF_PRIV_DATA_TYPE(ip_output_mbuf_data, { struct nexthop *nh; });
+GR_MBUF_PRIV_DATA_TYPE(ip_output_mbuf_data, {
+	struct nexthop *nh;
+	const struct iface *input_iface;
+});
 
 GR_MBUF_PRIV_DATA_TYPE(arp_mbuf_data, {
 	struct nexthop *local;
@@ -49,5 +53,7 @@ static inline void ip_set_fields(struct rte_ipv4_hdr *ip, struct ip_local_mbuf_d
 	ip->hdr_checksum = 0;
 	ip->hdr_checksum = rte_ipv4_cksum(ip);
 }
+
+#define GR_IP_ICMP_TTL_EXCEEDED 11
 
 #endif
--- a/modules/ip/datapath/ip_forward.c
+++ b/modules/ip/datapath/ip_forward.c
@@ -56,5 +56,3 @@ static struct gr_node_info info = {
 };
 
 GR_NODE_REGISTER(info);
-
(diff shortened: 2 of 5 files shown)


modules/ip/datapath/ip_ttl_exceeded.c
```diff
@@ -0,0 +1,82 @@
...
+
+		rte_node_enqueue_x1(graph, node, next, mbuf);
+	}
+
+	return nb_objs;
+}
+
+struct rte_node_register ip_ttl_exceeded_node = {
+	.name = "ip_ttl_exceeded",
+	.process = ip_ttl_exceeded,
```
rjarry, line 71: To be consistent with the other nodes, can you add `_process` to this function name?

modules/ip/datapath/ip_ttl_exceeded.c
```diff
@@ -0,0 +1,82 @@
...
+	for (uint16_t i = 0; i < nb_objs; i++) {
+		mbuf = objs[i];
+
+		ip = rte_pktmbuf_mtod(mbuf, struct rte_ipv4_hdr *);
+		// No need to check headroom
+		icmp = (struct rte_icmp_hdr *)
+			rte_pktmbuf_prepend(mbuf, sizeof(struct rte_icmp_hdr));
+
+		// Get the local router IP address from the input iface
+		input_iface = ip_output_mbuf_data(mbuf)->input_iface;
```
rjarry, line 42: It looks like you forgot to commit something else :) Nothing sets this input_iface value.

rjarry (changes requested):

modules/ip/datapath/ip_ttl_exceeded.c
```diff
@@ -0,0 +1,82 @@
...
+		ip4_addr_t local_ip = ip4_addr_get(input_iface->id)->ip;
+
+		ip_data = ip_local_mbuf_data(mbuf);
+		ip_data->vrf_id = vrf_id;
+		ip_data->src = local_ip;
+		ip_data->dst = ip->src_addr;
+
+		// RFC792 payload size: ip header + 64 bits of original datagram
+		ip_data->len = sizeof(struct rte_icmp_hdr)
+			+ RTE_MIN(sizeof(struct rte_ipv4_hdr) + 8,
```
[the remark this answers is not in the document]
christophefontaine: according to the rfc, we have to include the original IP header + 64 bits, within the udp/tcp src/dst ports are defined.

modules/ip/datapath/ip_ttl_exceeded.c
christophefontaine: will do.

modules/ip/datapath/ip_ttl_exceeded.c
christophefontaine: indeed !

modules/ip/datapath/ip_ttl_exceeded.c
```diff
@@ -0,0 +1,82 @@
...
+#include <gr_ip4_datapath.h>
+#include <gr_log.h>
+#include <gr_mbuf.h>
+
+#include <rte_common.h>
+#include <rte_graph_worker.h>
+#include <rte_icmp.h>
+#include <rte_ip.h>
+
+#define GR_IP_ICMP_TTL_EXCEEDED 11
```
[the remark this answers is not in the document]
rjarry: Can you put this constant in gr_ip4_datapath?

rjarry (changes requested):

rjarry (approved):

PR #24, opened 2024-07-04
trace: allow enabling packet trace at runtime

rjarry:
If the -t, --test-mode flag is used, enable packet tracing instead of
relying on an undocumented compilation flag. The performance impact
should be minimal when it is not enabled.


david-marchand:
On the principle, it looks good, but I suspect it has a non negligeable impact on performance.

rjarry:
Out of curiosity, I ran a simple IP forwarding performance test, before and after this patch.

```
before: 23.56 MPPS

GROUT: main: starting grout version v0.1-1-gbed15f4ef48e

NODE        CALLS  PACKETS  PKTS/CALL  CYCLES/CALL  CYCLES/PKT
ip_input     1988   315693      158.8       8367.9        52.7
port_rx      2176   315693      145.1       5573.3        38.4
port_tx      1988   315693      158.8       4400.9        27.7
eth_input    2176   315693      145.1       3263.6        22.5
ip_output    1988   315693      158.8       3168.4        20.0
eth_output   1988   315693      158.8       3135.6        19.7
ip_forward   1988   315693      158.8       1741.1        11.0

after: 23.56 MPPS

GROUT: main: starting grout version v0.1-2-ge3133c7c1c80

NODE        CALLS  PACKETS  PKTS/CALL  CYCLES/CALL  CYCLES/PKT
ip_input     2354   315196      133.9       7055.5        52.7
port_rx      2624   315196      120.1       4626.7        38.5
port_tx      2354   315196      133.9       3678.2        27.5
eth_input    2624   315196      120.1       2708.8        22.6
ip_output    2354   315196      133.9       2681.7        20.0
eth_output   2354   315196      133.9       2621.4        19.6
ip_forward   2354   315196      133.9       1455.5        10.9
```

If this patch has any impact, it is not noticeable in this particular test.

main/main.c
```diff
@@ -75,6 +77,7 @@ static int parse_args(int argc, char **argv) {
 			break;
 		case 't':
 			args.test_mode = true;
+			packet_trace_enabled = true;
```
christophefontaine, line 80: We may want to enable packet tracing with a real pci device, so test mode and packet tracing should be 2 independent options.

rjarry: Good point, I added `-x`, `--trace-packets`.

main/main.c
```diff
@@ -30,13 +30,14 @@
 #include <unistd.h>
 
 static void usage(const char *prog) {
-	printf("Usage: %s [-h] [-v] [-t] [-c FILE] [-s PATH]\n", prog);
```
david-marchand, line 33: I prefer alphabetical order. It is probably stupid, but it leaves no place for interpretation to "Where should I put my new option?".

That's a global comment on how the options parsing is done in this file.

modules/infra/datapath/rx.c
```diff
@@ -52,10 +53,15 @@ rx_process(struct rte_graph *graph, struct rte_node *node, void **objs, uint16_t
...
 			continue;
 		}
-		for (int r = count; r < count + rx; r++) {
+		for (r = count; r < count + rx; r++) {
 			eth_input_mbuf_data(node->objs[r])->iface = iface;
-			trace_packet("rx", iface->name, node->objs[r]);
 		}
+		if (unlikely(packet_trace_enabled)) {
+			for (r = count; r < count + rx; r++) {
+				trace_packet("rx", iface->name, node->objs[r]);
```
david-marchand, line 61: Nit: could we have trace_packets() that takes an array of packets?

modules/infra/datapath/trace.c
```diff
@@ -0,0 +1,149 @@
...
+	);
+
+	if (ether_type == RTE_ETHER_TYPE_VLAN) {
+		const struct rte_vlan_hdr *vlan;
+		uint16_t vlan_id;
+
+		vlan = rte_pktmbuf_mtod_offset(m, const struct rte_vlan_hdr *, offset);
+		offset += sizeof(*vlan);
+		vlan_id = rte_be_to_cpu_16(vlan->vlan_tci) & 0xfff;
+		ether_type = rte_be_to_cpu_16(vlan->eth_proto);
```
david-marchand, line 44: No QinQ support ?

main/main.c
rjarry: Ack, I'll change that in a separate commit. I will also update the man page stub.

modules/infra/datapath/trace.c
rjarry: not yet.

modules/infra/datapath/rx.c
rjarry: I guess we could, but it would be used only in rx.c. In eth_output, packets are printed one by one.

main/main.c
rjarry: The -h --help flag should always be first though. Don't you think?

modules/infra/datapath/rx.c
david-marchand: Ok, let's wait for another user to appear.

modules/infra/datapath/trace.c
david-marchand: It seems stacking vlan interfaces is properly checked in the cli, so we should be good.
So ok for me.

main/main.c
david-marchand: No exception, plus -h is quasi standard and a user should not be looking for it (hopefully).

rjarry: Allez, d'accord.

PR #25, opened 2024-07-04
modules: introduce LLDP support

christophefontaine:
Interfaces must accept multicast frames:
`grcli set interface port <port0> allmulti on
`
LLDP is configured globally
```
grcli set lldp [rx|tx|both|off]
grcli set lldp ttl [10..600]
```

```
# grcli show lldp config
lldp: rx: on tx: on ttl: 60

# grcli show lldp neighbors
LLDP Neighbors: 2
Age  TTL  ifname        Sys name
  4 120s   dpdk-p0      crs317
  6 120s   p1   compute04.mgmt.foobar.space
```


david-marchand:
> Interfaces must accept multicast frames: `grcli set interface port <port0> allmulti on ` LLDP is configured globally
> 

I would prefer a list of interfaces with LLDP enabled is set in the lldp module.

This way, the LLDP module request this allmulti stuff (better from a user pov).
And I would filter unwanted LLDP frames if received from an iface not in the list.

modules/lldp/cli.c
```diff
@@ -0,0 +1,209 @@
...
+			return CMD_ERROR;
+
+		printf("%3ld ", age);
+		decode_lldp(
+			resp->neighbors[i].n_tlv_data,
+			resp->neighbors[i].tlv_data,
+			iface.name,
+			brief
+		);
+	}
```
rjarry, line 151: Could you use libsmartcols to print the table here? See `show ip nexthop` for inspiration.

https://github.com/rjarry/grout/blob/bed15f4ef48ebf770d84fbb9f52be736c742ac06/modules/ip/cli/nexthop.c#L79-L126

modules/lldp/cli.c
```diff
@@ -0,0 +1,209 @@
...
+		with_help("Disable", ec_node_str("off", "off"))
+	);
+
+	ret = CLI_COMMAND(
+		CLI_CONTEXT(root, CTX_SET, CTX_ARG("lldp", "Modify LLDP configuration")),
+		"ttl TTL",
+		lldp_set_ttl,
+		"Set LLDP TTL",
+		with_help("Interval in seconds, 10 < ttl < 600", ec_node_uint("TTL", 10, 600, 10))
+	);
```
rjarry, line 176: Can you merge these two commands into a single one and use a set_attrs field in the API message to determine which settings are to be considered?

https://github.com/rjarry/grout/blob/bed15f4ef48ebf770d84fbb9f52be736c742ac06/modules/infra/api/gr_infra.h#L142

modules/lldp/sink.c
```diff
@@ -0,0 +1,79 @@
...
+	rte_edge_t edge = gr_node_attach_parent("eth_input", "lldp_sink");
+	if (edge == RTE_EDGE_ID_INVALID)
+		ABORT("gr_node_attach_parent(eth_input, lldp_input) failed");
+	gr_eth_input_add_type(rte_cpu_to_be_16(RTE_ETHER_TYPE_LLDP), edge);
+}
+
+static void lldp_sink_unregister(void) { }
+
+static struct rte_node_register node = {
+	.name = "lldp_sink",
```
rjarry, line 65: I'm not familiar with LLDP. Is there a reason why this node is called lldp_sink instead of lldp_input?

modules/lldp/announce.c
```diff
@@ -0,0 +1,191 @@
...
+}
+
+static void lldp_announce_unregister(void) {
+	rte_ring_free(lldp_ring);
+	rte_mempool_free(lldp_pool);
+}
+
+static struct rte_node_register lldp_announce_node = {
+	.flags = RTE_NODE_SOURCE_F,
+	.name = "lldp_announce",
```
rjarry, line 177: Same remark, I think it would be more consistent with the rest of the nodes if it was called lldp_output.

modules/lldp/sink.c
```diff
@@ -0,0 +1,79 @@
...
+#include <rte_graph_worker.h>
+
+#include <time.h>
+
+enum {
+	LLDP_REPLY = 0,
+	DROP_IN,
+	EDGE_COUNT,
+};
+extern struct gr_lldp_node_data lldp_ctx;
```
rjarry, line 25: Is this variable still used?

modules/lldp/control.c
```diff
@@ -0,0 +1,180 @@
...
+#include <gr_worker.h>
+
+#include <event2/event.h>
+#include <rte_common.h>
+#include <rte_graph_worker.h>
+
+int lldp_output_request_solicit(const struct iface *iface);
+
+extern struct gr_lldp_node_data lldp_ctx;
+extern struct gr_lldp_neigh neighbors[];
```
rjarry, line 20: I didn't find this declared in any header. Maybe it would be best to keep the symbol private to `sink.c` and expose it via functions.

rjarry (changes requested):

modules/lldp/cli.c
christophefontaine: I'm rewriting this part indeed: for a verbose output, I intend to switch to an output similar to `lldpcli show neigh`:
```
-------------------------------------------------------------------------------
Interface:    ens1f0, via: LLDP, RID: 1, Time: 0 day, 00:00:09
  Chassis:
    ChassisID:    mac dc:2c:6e:47:35:77
    SysName:      crs317
    SysDescr:     MikroTik RouterOS 7.12.1 (stable) Nov/17/2023 11:38:45 CRS317-1G-16S+
    MgmtIP:       192.168.1.3
    MgmtIP:       2a01:cb19:141e:f200::3
    MgmtIP:       fe80::de2c:6eff:fe47:3587
    Capability:   Bridge, on
    Capability:   Router, on
  Port:
    PortID:       ifname bridge/sfp-sfpplus6
    TTL:          120
------------------------------------------------------------------------------- 

```

For the brief output, I'll indeed use libsmartcol.

modules/lldp/cli.c
christophefontaine: Thanks ! That's what I was looking for :)

modules/lldp/announce.c
christophefontaine: I chose `lldp_announce` as it is a source node, it felt weird to have an `*_output` node as a source node.
WDYT ?

modules/lldp/sink.c
christophefontaine: A placeholder now: it will hold additional info, such as the system name, mgmt address, etc.

modules/lldp/control.c
christophefontaine: ack.

modules/lldp/announce.c
rjarry: I agree that it feels weird but it would be consistent with arp_output_request. Keep in mind that node names show up in the software stats. Users don't see that it is a source node unless they dump the graph and convert it to an image.

modules/lldp/sink.c
christophefontaine: No related to lldp, but I wanted to express that the packets won't be forwarded.
Won't `*_input`  imply that there will be subsequent nodes ?

christophefontaine:
> > Interfaces must accept multicast frames: `grcli set interface port <port0> allmulti on ` LLDP is configured globally
> 
> I would prefer a list of interfaces with LLDP enabled is set in the lldp module.
> 
> This way, the LLDP module request this allmulti stuff (better from a user pov). And I would filter unwanted LLDP frames if received from an iface not in the list.

Shouldn't `allmulti` be enabled by default, as it is required for IPv6 ?
And then we'd need to have a rx/tx configuration knob for each interface.

modules/lldp/sink.c
rjarry: OK, I would still prefer if we kept the input/output terminology to be consistent with other nodes. Since the node names used when displaying software stats, it will be less confusing.

david-marchand:
> > > Interfaces must accept multicast frames: `grcli set interface port <port0> allmulti on ` LLDP is configured globally
> > 
> > 
> > I would prefer a list of interfaces with LLDP enabled is set in the lldp module.
> > This way, the LLDP module request this allmulti stuff (better from a user pov). And I would filter unwanted LLDP frames if received from an iface not in the list.
> 
> Shouldn't `allmulti` be enabled by default, as it is required for IPv6 ? And then we'd need to have a rx/tx configuration knob for each interface.

Ideally, no.

Each module requiring multicast traffic should register its set of mcast addresses.
If adding those addresses is not supported (or a limit, like a max number of mac is reached), then allmulti should be enabled.

For the latter case, if the user enables allmulti for debugging, or whatever else, allmulti should still be maintained when the user disables it.
In the Linux kernel, there is a counter of allmulti / promiscuous users to handle this case.

I think grout will need a similar infrastructure code, for adding unicast / mcast address filtering.

david-marchand:
I created two issues https://github.com/rjarry/grout/issues/27 and https://github.com/rjarry/grout/issues/28.

modules/lldp/input.c
```diff
@@ -0,0 +1,89 @@
...
+	return neighbors;
+}
+
+static uint16_t
+lldp_input_process(struct rte_graph *graph, struct rte_node *node, void **objs, uint16_t nb_objs) {
+	struct iface_info_port *port;
+	const struct iface *iface;
+	const uint8_t *payload;
+	struct rte_mbuf *mbuf;
+	clock_t now = clock();
```
rjarry, line 41: Is this ok to call in the datapath? Maybe rte_rdtsc() would be better?

modules/lldp/output.c
```diff
@@ -0,0 +1,172 @@
...
+#include <rte_version.h>
+
+enum {
+	OUTPUT = 0,
+	ERROR,
+	EDGE_COUNT,
+};
+
+extern struct gr_lldp_conf_common_data lldp_ctx;
+extern struct gr_lldp_conf_iface_data lldp_iface_ctx[RTE_MAX_ETHPORTS];
```
rjarry, line 28: move to `lldp_priv.h` ?

modules/lldp/output.c
```diff
@@ -0,0 +1,172 @@
...
+
+const struct rte_ether_addr lldp_dst = {
+	.addr_bytes = {0x01, 0x80, 0xc2, 0x00, 0x00, 0x0e},
+};
```
rjarry, line 32: This is already defined in `control.c` as `lldp_dst2`. I'm OK with duplicating but at least add `static`.

modules/lldp/control.c
```diff
@@ -0,0 +1,274 @@
...
+static struct rte_ether_addr lldp_dst0 = {
+	.addr_bytes = {0x01, 0x80, 0xc2, 0x00, 0x00, 0x00},
+};
+static struct rte_ether_addr lldp_dst1 = {
+	.addr_bytes = {0x01, 0x80, 0xc2, 0x00, 0x00, 0x03},
+};
+static struct rte_ether_addr lldp_dst2 = {
+	.addr_bytes = {0x01, 0x80, 0xc2, 0x00, 0x00, 0x0e},
+};
+
```
rjarry, line 30: If these are only defined to be called in `lldp_iface_event`, it would be better to have them as stack variables.

modules/lldp/output.c
```diff
@@ -0,0 +1,172 @@
...
+	if (ret < 0)
+		return errno_set(-ret);
+
+	return 0;
+}
+
+static uint8_t *lldp_tlv_append(
+	struct rte_mbuf *mbuf,
+	uint8_t type,
+	uint8_t subtype,
```
rjarry, line 51: see proper naming/sizing of enums in `gr_lldp.h`.

modules/lldp/output.c
```diff
@@ -0,0 +1,172 @@
...
+		ttl_value = rte_cpu_to_be_16(lldp_ctx.ttl);
+	} else {
+		// Specific case: TX is disabled, but we want to
+		// send a last message with a TTL set to 0
+		ttl_value = 0;
+	}
+
+	// Only T_CHASSIS_ID T_PORT_ID T_TTL and T_END are mandatory
+	lldp_tlv_append(
+		mbuf, T_CHASSIS_ID, T_CHASSIS_MAC_ADDRESS, RTE_ETHER_ADDR_LEN, src_addr.addr_bytes
```
rjarry, line 93: minor nit pick:

```c
mbuf, T_CHASSIS_ID, T_CHASSIS_MAC_ADDRESS, sizeof(src_addr), &src_addr
```

modules/lldp/output.c
```diff
@@ -0,0 +1,172 @@
...
+	);
+	lldp_tlv_append(
+		mbuf, T_CHASSIS_ID, T_CHASSIS_IF_ALIAS, strlen(lldp_ctx.sys_name), lldp_ctx.sys_name
+	);
+	lldp_tlv_append(
+		mbuf, T_PORT_ID, T_PORT_MAC_ADDRESS, RTE_ETHER_ADDR_LEN, src_addr.addr_bytes
```
rjarry, line 99: Same:

```c
mbuf, T_CHASSIS_ID, T_CHASSIS_MAC_ADDRESS, sizeof(src_addr), &src_addr
```

rjarry (changes requested):

modules/lldp/cli.c
```diff
@@ -0,0 +1,387 @@
...
+#include <stdio.h>
+
+static cmd_status_t lldp_global_config(const struct gr_api_client *c, const struct ec_pnode *p) {
+	struct gr_lldp_set_global_conf_req req = {.set_attrs = 0};
+	void *resp = NULL;
+	uint16_t ttl = 0;
+	const char *str;
+
+	arg_u16(p, "TTL", &ttl);
+	if (ttl > 0) {
```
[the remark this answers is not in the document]
christophefontaine: TTL value should be already checked when it is parsed (min 10s, max 600s ) in the CLI.
```with_help("Interval in seconds, 10 < ttl < 600", ec_node_uint("TTL", 10, 600, 10)),```
Here, we only check for existence of the parameter.

File: modules/lldp/cli.c [hunk 15]
```diff
@@ -0,0 +1,387 @@
...
+		strcpy(req.sys_name, str);
+		req.set_attrs |= GR_LLDP_SET_NAME;
+	}
+	str = arg_str(p, "SYSDESC");
+	if (str) {
+		strcpy(req.sys_descr, str);
+		req.set_attrs |= GR_LLDP_SET_DESC;
+	}
+
+	free(resp);
```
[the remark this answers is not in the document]
christophefontaine: indeed, cleaned up.

modules/lldp/cli.c
```diff
@@ -0,0 +1,387 @@
...
+				);
+				break;
+			case T_CHASSIS_NET_ADDRESS:
+				uint8_t afi = data[offset];
+				if (afi == AFI_IP_4) {
+					ip4_net_format(
+						(const struct ip4_net *)data, buf, sizeof(buf)
+					);
+				}
+				snprintf(n->chassis_id, 512, "ip %s", buf);
```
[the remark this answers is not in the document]
christophefontaine: That's why `buf` is initialised to "".

modules/lldp/cli.c
```diff
@@ -0,0 +1,387 @@
...
+			case SUBTYPE_MANAGEMENT_ADDRESS_IPV4:
+				if (addr_str_len != 4)
+					break;
+				inet_ntop(AF_INET, &data[offset], n->mgmt_addr[0], INET_ADDRSTRLEN);
+				break;
+			case SUBTYPE_MANAGEMENT_ADDRESS_IPV6:
+				if (addr_str_len != 16)
+					break;
+				// gr_ip6_net_format((const struct ip6_net *)data, n->mgmt_addr[1], sizeof(n->mgmt_addr[1]));
+				break;
```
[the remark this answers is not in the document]
christophefontaine: Now we have IPv6 support, I'll use `ip6_net_format`.

modules/lldp/input.c
```diff
@@ -0,0 +1,89 @@
...
+
+			rte_pktmbuf_free(mbuf);
+		}
+	}
+
+	return nb_objs;
+}
+
+static void lldp_input_register(void) {
+	gr_eth_input_add_type(rte_cpu_to_be_16(RTE_ETHER_TYPE_LLDP), "lldp_input");
```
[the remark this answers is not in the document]
christophefontaine: done.

vjardin:
Why should a complete LLDP stack be redesigned ? What's about all the LLDP options ?

This stack https://github.com/lldpd/lldpd would be a good fit, but I don't understand the datamodel that could be used for lldpd to leverage grout.

Moreover, it would be a proof point that the integration of a signaling stack is achievable.

cc @vincentbernat

christophefontaine:
Le 6 oct. 2024 à 19:54, Vincent JARDIN ***@***.***> a écrit :﻿
Why should a complete LLDP stack be redesigned ? What's about all the LLDP options ?The initial idea wasn’t to implement all options, but to have rather a minimal implementation, to see what needs to be added to grout. This stack https://github.com/lldpd/lldpd would be a good fit, but I don't understand the datamodel that could be used for lldpd to leverage grout.
Moreover, it would be a proof point that the integration of a signaling stack is achievable.Indeed, we may want to revisit it as the infra is here, to investigate how we can integrate external stacks to grout. 
cc @vincentbernat

christophefontaine:
Not enough time to work on that, closing to rework properly on it later.

PR #29, opened 2024-07-06
Slow path

christophefontaine:
Add "slow_path_input" node, and modify existing arp_request_output to use this infrastructure.

### DIFF
--- a/modules/infra/datapath/control_input.c
+++ b/modules/infra/datapath/control_input.c
@@ -1,0 +1,122 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2024 Christophe Fontaine
+
+#include "gr_control_input.h"
+
+#include <gr_eth_input.h>
+#include <gr_eth_output.h>
+#include <gr_graph.h>
+#include <gr_log.h>
+#include <gr_mbuf.h>
+
+#include <rte_ether.h>
+#include <rte_graph_worker.h>
+#include <rte_malloc.h>
+#include <rte_version.h>
+
+enum {
+	UNKNOWN_CONTROL_INPUT_TYPE,
+	EDGE_COUNT,
+};
+
+struct gr_control_input_msg {
+	uint16_t type;
+	void *data;
+};
+
+static struct rte_ring *control_input_ring;
+static struct rte_mempool *control_input_pool;
+
+static rte_edge_t control_input_edges[1 << 8] = {UNKNOWN_CONTROL_INPUT_TYPE};
+
+void gr_control_input_add_handler(control_input_t type, const char *node_name) {
+	rte_edge_t edge = gr_node_attach_parent("control_input", node_name);
+	if (edge == RTE_EDGE_ID_INVALID)
+		ABORT("gr_node_attach_parent(control_input, lldp_output) failed");
+	if (control_input_edges[type] != UNKNOWN_CONTROL_INPUT_TYPE) {
+		ABORT("control_input: type=0x%04x -> already set to edge %u", type, edge);
+	}
+	control_input_edges[type] = edge;
+	LOG(DEBUG, "control_input: type=0x%04x -> edge %u", type, edge);
+}
+
+int post_to_stack(control_input_t type, void *data) {
+	struct gr_control_input_msg msg = {.type = type, .data = data};
+	int ret;
+
+	ret = rte_ring_enqueue_elem(control_input_ring, &msg, sizeof(msg));
+	if (ret < 0)
+		return errno_set(-ret);
+
+	return 0;
+}
+
+static uint16_t
+control_input_process(struct rte_graph *graph, struct rte_node *node, void **, uint16_t) {
+	struct gr_control_input_msg msg[RTE_GRAPH_BURST_SIZE];
+	struct rte_mbuf *mbuf;
+	uint16_t n;
+
+	n = rte_ring_dequeue_burst_elem(
+		control_input_ring,
+		msg,
+		sizeof(struct gr_control_input_msg),
+		RTE_GRAPH_BURST_SIZE,
+		NULL
+	);
+
+	for (unsigned i = 0; i < n; i++) {
+		mbuf = rte_pktmbuf_alloc(control_input_pool);
+		if (mbuf) {
+			control_input_mbuf_data(mbuf)->data = msg[i].data;
+			rte_node_enqueue_x1(graph, node, control_input_edges[msg[i].type], mbuf);
+		}
+	}
+
+	return n;
+}
+
+static void control_input_register(void) {
+	control_input_ring = rte_ring_create(
+		"control_input",
+		RTE_GRAPH_BURST_SIZE * 4,
+		SOCKET_ID_ANY,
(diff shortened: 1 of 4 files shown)


modules/ip/datapath/arp_output_request.c
```diff
@@ -115,34 +113,10 @@ arp_output_request_process(struct rte_graph *graph, struct rte_node *node, void
...
-		SOCKET_ID_ANY
-	);
-	if (arp_pool == NULL)
-		ABORT("rte_pktmbuf_pool_create(arp_output_request): %s", rte_strerror(rte_errno));
-}
-
-static void arp_output_request_unregister(void) {
-	rte_ring_free(nexthop_ring);
-	rte_mempool_free(arp_pool);
+	gr_slow_path_input_add_handler(RTE_ETHER_TYPE_IPV4, "arp_output_request");
```
christophefontaine, line 116: @rjarry Not sure if I can use directly `RTE_ETHER_TYPE_IPV4`, add an enum in `gr_slow_path.h`, or create an API to fetch the next available slot ?
`arp_slow_path_type = gr_slow_path_input_add_handler ("arp_output_request");`
And then `slow_path_post(arp_slow_path_type, ... )`

rjarry: You will probably need to do it in two steps. First append the node to slow_path next nodes, and then, add the node id indexed with the slow path "type", in this case `ARP_SOLICIT`.

Have a look at how dynamic nodes are registered after ip_local or after eth_input.

Also, I think we need a local enum here to reference the slow path type. Not reuse `RTE_ETHER_TYPE_IPV4` which is an ether type value for IPv4, not for ARP (maybe a typo?).

modules/infra/datapath/slow_path_input.c
```diff
@@ -0,0 +1,123 @@
...
+
+struct gr_slow_path_msg {
+	uint16_t type;
+	void *data;
+};
+
+static struct rte_ring *slow_path_ring;
+static struct rte_mempool *slow_path_pool;
+
+static rte_edge_t slow_path_edges[1 << 16] = {UNKNOWN_SLOW_PATH_TYPE};
```
rjarry, line 30: You can make this array a lot smaller if slow path type is uint8_t.

modules/infra/datapath/gr_slow_path.h
```diff
@@ -0,0 +1,16 @@
...
+#define _GR_INFRA_SLOW_PATH
+
+#include "gr_mbuf.h"
+
+#include <rte_graph.h>
+
+GR_MBUF_PRIV_DATA_TYPE(slow_path_input_mbuf_data, { void *data; });
+
+void gr_slow_path_input_add_handler(uint16_t type, const char *node_name);
+int slow_path_post(uint16_t type, void *data);
```
rjarry, line 14: Maybe restrict slow path type to uint8_t ? Will we ever need more than 255 slow path handlers?

modules/infra/datapath/slow_path_input.c
```diff
@@ -0,0 +1,123 @@
...
+
+void gr_slow_path_input_add_handler(uint16_t type, const char *node_name) {
+	rte_edge_t edge = gr_node_attach_parent("slow_path_input", node_name);
+	if (edge == RTE_EDGE_ID_INVALID)
+		ABORT("gr_node_attach_parent(slow_path_input, lldp_output) failed");
+	if (slow_path_edges[type] != UNKNOWN_SLOW_PATH_TYPE) {
+		LOG(WARNING, "slow_path_input: type=0x%04x -> already set to edge %u", type, edge);
```
rjarry, line 37: Maybe `ABORT()` in this case?

modules/ip/datapath/arp_output_request.c
```diff
@@ -26,53 +27,47 @@ enum {
...
 
-static struct rte_ring *nexthop_ring;
-
 int arp_output_request_solicit(struct nexthop *nh) {
 	int ret;
 	if (nh == NULL)
 		return errno_set(EINVAL);
 	ip4_nexthop_incref(nh);
-	ret = rte_ring_enqueue(nexthop_ring, nh);
+	ret = slow_path_post(RTE_ETHER_TYPE_IPV4, (void *)nh);
```
rjarry, line 35: I don't think `RTE_ETHER_TYPE_IPV4` is suitable here. You should define your own value `SLOW_PATH_ARP_SOLICIT`.

modules/ip/datapath/arp_output_request.c
```diff
@@ -26,53 +27,47 @@ enum {
...
+	for (unsigned i = 0; i < n_objs; i++) {
+		mbuf = objs[i];
+		nh = (struct nexthop *)slow_path_input_mbuf_data(mbuf)->data;
 		local = ip4_addr_get(nh->iface_id);
 		if (local == NULL) {
 			next = ERROR;
 			goto next;
 		}
 
+		rte_pktmbuf_reset(mbuf);
```
rjarry, line 69: Not required, the mbuf just got allocated in the parent node.

rjarry (changes requested):

modules/infra/datapath/slow_path_input.c
christophefontaine: fixed.

modules/ip/datapath/arp_output_request.c
christophefontaine: fixed

modules/ip/datapath/arp_output_request.c
christophefontaine: This is done in `gr_slow_path_input_add_handler`, so we don't have to append explicitly the next node.
As we abort anyway if this fails, I didn't see the value of asking the user to attach manually the next node.

Maybe we can also simplify the API for `gr_eth_input_add_type`, so it takes a node name instead of an edge ?
`void gr_eth_input_add_type(rte_be16_t eth_type, char * node_name)`

rjarry:
Or even simpler: `control_output` (and `control_input` for exception packets if we ever need them).

modules/infra/datapath/control_input.c
```diff
@@ -0,0 +1,127 @@
...
+		RING_F_MP_RTS_ENQ | RING_F_MC_RTS_DEQ
+	);
+	if (control_input_ring == NULL)
+		ABORT("rte_ring_create(arp_output_request): %s", rte_strerror(rte_errno));
+
+	control_input_pool = rte_pktmbuf_pool_create(
+		"control_input",
+		RTE_GRAPH_BURST_SIZE * 4,
+		256, // cache_size
+		0, // priv_size
```
rjarry, line 98: Since commit https://github.com/rjarry/grout/commit/a12f55fb1cad5fe9620146fd2d50bb4b61a356bf you will need to specify the correct priv_size.

We need to move this into a single function to make sure we don't forget it.

rjarry (changes requested):

modules/infra/datapath/control_input.c
david-marchand: A helper for mempool creation is probably better.

PR #30, opened 2024-07-08
infra/port: implement multicast l2 filtering

### LINKED ISSUE #27: implement unicast/multicast l2 filtering
A number of modules may require traffic from some unicast or multicast addresses that are filtered by hw when in non promisc / non allmulti mode.

Always running in promisc / allmulti is certainly possible, though it means receiving unwanted traffic.
But relying on promisc / allmulti may hide some network misconfiguration, or some other network equipment misbehaving, as grout would handle traffic that it should not in theory.

The request here is to add per iface lists of unicast and mcast addresses for which traffic is expected.

For a dpdk port backed interface, such lists will be associated with calls to rte_eth_dev_mac_addr_add, rte_eth_dev_mac_addr_del for unicast and rte_eth_dev_set_mc_addr_list. If adding a unicast/mcast address fails, then fallback would be to enable promisc / allmulti.

Associated with this, there is a need to track if enabling promisc/allmulti is a consequence of those unicast/mcast addresses or a request from the user (think of the case when the user enables promisc/allmulti via grcli, and later disables it).
The Linux kernel solution is to maintain a count of promisc / allmulti users (reflected by the "promiscuity" / "allmulti" counters one can find with "ip -s -d link").

christophefontaine:
- wrappers to call rte_eth_dev_mac_addr_[add|remove]
- Add API to filter multicast addresses on ports

resolves #27

### DIFF
--- a/modules/infra/cli/port.c
+++ b/modules/infra/cli/port.c
@@ -11,6 +11,7 @@
 
 #include <ecoli.h>
 #include <libsmartcols.h>
+#include <rte_ether.h>
 
 #include <errno.h>
 #include <sys/queue.h>
@@ -20,7 +21,7 @@ static void port_show(const struct gr_api_client *, const struct gr_iface *iface
 
 	printf("devargs: %s\n", port->devargs);
 	printf("driver:  %s\n", port->driver_name);
-	printf("mac: " ETH_ADDR_FMT "\n", ETH_BYTES_SPLIT(port->mac.bytes));
+	printf("mac: " ETH_ADDR_FMT "\n", ETH_ADDR_SPLIT(&port->mac));
 	printf("n_rxq: %u\n", port->n_rxq);
 	printf("n_txq: %u\n", port->n_txq);
 	printf("rxq_size: %u\n", port->rxq_size);
@@ -31,11 +32,7 @@ static void
 port_list_info(const struct gr_api_client *, const struct gr_iface *iface, char *buf, size_t len) {
 	const struct gr_iface_info_port *port = (const struct gr_iface_info_port *)iface->info;
 	snprintf(
-		buf,
-		len,
-		"devargs=%s mac=" ETH_ADDR_FMT,
-		port->devargs,
-		ETH_BYTES_SPLIT(port->mac.bytes)
+		buf, len, "devargs=%s mac=" ETH_ADDR_FMT, port->devargs, ETH_ADDR_SPLIT(&port->mac)
 	);
 }
 
@@ -65,7 +62,7 @@ static uint64_t parse_port_args(
 		}
 		memccpy(port->devargs, devargs, 0, sizeof(port->devargs));
 	}
-	if (eth_addr_parse(arg_str(p, "MAC"), &port->mac) == 0)
+	if (arg_eth_addr(p, "MAC", &port->mac) == 0)
 		set_attrs |= GR_PORT_SET_MAC;
 
 	if (arg_u16(p, "N_RXQ", &port->n_rxq) == 0)
--- a/modules/infra/control/port.c
+++ b/modules/infra/control/port.c
@@ -425,9 +425,182 @@ const struct iface *port_get_iface(uint16_t port_id) {
 	return port_ifaces[port_id];
 }
 
-static int iface_port_get_eth_addr(const struct iface *iface, struct rte_ether_addr *mac) {
-	const struct iface_info_port *port = (const struct iface_info_port *)iface->info;
+static int port_mac_get(const struct iface *iface, struct rte_ether_addr *mac) {
+	struct iface_info_port *port = (struct iface_info_port *)iface->info;
 	rte_ether_addr_copy(&port->mac, mac);
+	return 0;
+}
+
+static int port_mac_add(struct iface *iface, const struct rte_ether_addr *mac) {
+	struct iface_info_port *port = (struct iface_info_port *)iface->info;
+	struct mac_filter *filter;
+	const char *mac_type;
+	bool multicast;
+	uint8_t i;
+	int ret;
+
+	if (mac == NULL || !(rte_is_multicast_ether_addr(mac) || rte_is_unicast_ether_addr(mac)))
+		return errno_set(EINVAL);
+
+	if (rte_is_same_ether_addr(mac, &port->mac))
+		return 0;
+
+	multicast = rte_is_multicast_ether_addr(mac);
+	if (multicast) {
+		mac_type = "multicast";
+		filter = &port->mcast_filter;
+	} else {
+		mac_type = "unicast";
+		filter = &port->ucast_filter;
+	}
+
+	for (i = 0; i < filter->count; i++) {
+		if (rte_is_same_ether_addr(&filter->mac[i], mac)) {
+			LOG(INFO,
+			    "%s: %s mac " ETH_ADDR_FMT " already filtered (refs=%u)",
+			    iface->name,
+			    mac_type,
+			    ETH_ADDR_SPLIT(mac),
+			    filter->refcnt[i]++);
+			return 0;
+		}
+	}
+	if (i == ARRAY_DIM(filter->mac))
+		return errno_log(ENOSPC, mac_type);
+
+	rte_ether_addr_copy(mac, &filter->mac[i]);
+	filter->refcnt[i] = 1;
+	filter->count++;
+
+	LOG(NOTICE,
+	    "%s: enabling %s " ETH_ADDR_FMT " mac filter",
+	    iface->name,
+	    mac_type,
+	    ETH_ADDR_SPLIT(mac));
+
+	if (filter->flags & MAC_FILTER_F_ALL)
+		return 0;
+
+	if (multicast)
+		ret = rte_eth_dev_set_mc_addr_list(port->port_id, filter->mac, filter->count);
+	else
+		ret = rte_eth_dev_mac_addr_add(port->port_id, &filter->mac[i], 0);
+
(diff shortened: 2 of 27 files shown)


christophefontaine:
draft PR, to get some feedback on mcast filtering, before adding unicast filtering.

modules/infra/control/port.c
```diff
@@ -457,6 +463,72 @@ static struct iface_type iface_type_port = {
...
+static int mcast_count(struct iface_info_port *p) {
+	int count = 0;
+	for (int i = 0; i < GR_PORT_MCAST_ADDR_MAX; i++) {
+		if (p->mcast_refcnt[i])
+			count++;
+	}
+	return count;
+}
+
+static void mcast_sort(struct iface_info_port *p) {
```
david-marchand, line 475: qsort() ?

modules/infra/control/port.c
```diff
@@ -457,6 +463,72 @@ static struct iface_type iface_type_port = {
...
+		}
+	}
+}
+
+int gr_port_mcast_addr_add(struct iface_info_port *p, struct rte_ether_addr *mcast_addr) {
+	int i;
+	for (i = 0; i < GR_PORT_MCAST_ADDR_MAX; i++) {
+		if (rte_is_same_ether_addr(mcast_addr, &p->mcast[i])) {
+			LOG(DEBUG, "multicast address already filtered by port %d", p->port_id);
+			p->mcast_refcnt[i]++;
```
david-marchand, line 497: In which case do we have multiple modules requiring the same mcast address?

modules/infra/control/port.c
```diff
@@ -409,6 +409,9 @@ static int iface_port_init(struct iface *iface, const void *api_info) {
 	if (port->devargs == NULL)
 		goto fail;
 
+	memset(port->mcast_refcnt, 0, sizeof(port->mcast_refcnt));
+	memset(port->mcast, 0, sizeof(port->mcast));
```
rjarry, line 413: I am not sure this is necessary since you are calling memcpy() just after.

rjarry (changes requested):

modules/infra/control/port.c
christophefontaine: Multiple protocols may use the same multicast mac address, or the same multicast IPs from 2 VRFs.

modules/infra/control/port.c
christophefontaine: indeed !

rjarry (changes requested):

modules/infra/control/port.c
```diff
@@ -309,6 +309,28 @@ int iface_port_reconfig(
 			return errno_log(-ret, "rte_eth_macaddr_get");
 	}
 
+	if ((set_attrs & GR_PORT_ADD_MAC) && !eth_addr_is_zero(&api->mac)) {
```
rjarry, line 312: If the flag is set, should you actually check the value of the mac address? The other functions will fail if it is invalid.

modules/infra/control/port.c
```diff
@@ -309,6 +309,28 @@ int iface_port_reconfig(
...
+		memcpy(&mac, &api->mac, sizeof(mac));
+		if (rte_is_multicast_ether_addr(&mac))
+			ret = gr_port_mcast_addr_add(p, &mac);
+		else
+			ret = gr_port_unicast_addr_add(p, &mac);
+		if (ret)
+			return errno_log(-ret, "add mac address");
+	}
+
+	if ((set_attrs & GR_PORT_REM_MAC) && !eth_addr_is_zero(&api->mac)) {
```
rjarry, line 323: If the flag is set, should you actually check the value of the mac address? The other functions will fail if it is invalid.

modules/infra/control/port.c
```diff
@@ -406,8 +428,15 @@ static int iface_port_init(struct iface *iface, const void *api_info) {
 	if (port->devargs == NULL)
 		goto fail;
 
+	memset(port->multicast, 0, sizeof(port->multicast));
```
rjarry, line 431: Why is this needed?

modules/infra/control/port.c
```diff
@@ -461,6 +508,162 @@ static struct iface_type iface_type_port = {
...
+		LOG(INFO,
+		    "Too many multicast addresses requested for %s",
+		    port_get_iface(p->port_id)->name);
+		goto enable_allmulti;
+	}
+
+	count = mcast_count(p);
+	for (i = 0; i < count; i++) {
+		memcpy(&mac[i], &p->multicast[i].mac, sizeof(struct rte_ether_addr));
+	}
```
rjarry, line 551: No need for braces.

rjarry:
```c
sizeof(mac[i])
```

modules/infra/control/port.c
```diff
@@ -461,6 +508,162 @@ static struct iface_type iface_type_port = {
...
+		LOG(WARNING,
+		    "Enabling mc addr list: no space left on iface %s.",
+		    port_get_iface(p->port_id)->name);
+		p->flags |= GR_PORT_F_ALLMULTI_TOO_MANY_MCAST;
+	} else if (ret == -ENOTSUP) {
+		LOG(WARNING,
+		    "Enabling mc addr list not supported for iface %s",
+		    port_get_iface(p->port_id)->name);
+		p->flags |= GR_PORT_F_ALLMULTI_DEVICE_LIMIT;
+	}
```
rjarry, line 567: Minor style preferences:

* Do not use capital letters nor punctuation in log messages.
* Prefer shorter messages if possible.
* Maybe a switch case would be better here.

modules/infra/control/port.c
```diff
@@ -461,6 +508,162 @@ static struct iface_type iface_type_port = {
...
+				return 0;
+			}
+		}
+	}
+
+	if (i == GR_PORT_NUM_MCAST_ADDR) {
+		LOG(INFO,
+		    "multicast addr wasn't allowed for iface %s",
+		    port_get_iface(p->port_id)->name);
+		return -ENODATA;
```
rjarry, line 606:
```c
return errno_log(ENOENT, "mcast_addr_del");
```

modules/infra/control/port.c
```diff
@@ -461,6 +508,162 @@ static struct iface_type iface_type_port = {
...
+			return rte_eth_allmulticast_enable(p->port_id);
+		}
+	} else {
+		rte_eth_dev_set_mc_addr_list(p->port_id, NULL, 0);
+	}
+
+	iface = port_get_iface(p->port_id);
+	if (p->flags & ~GR_PORT_F_ALLMULTI_DEVICE_LIMIT && iface->flags & ~GR_IFACE_F_ALLMULTI) {
+		return rte_eth_allmulticast_disable(p->port_id);
+	}
```
rjarry, line 631: No braces please. And maybe this test can be shortened:

```c
if (p->flags & ~(GR_PORT_F_ALLMULTI_DEVICE_LIMIT | GR_IFACE_F_ALLMULTI))
```

modules/infra/control/port.c
```diff
@@ -461,6 +508,162 @@ static struct iface_type iface_type_port = {
...
+
+int gr_port_unicast_addr_add(struct iface_info_port *p, struct rte_ether_addr *ucast_addr) {
+	int ret;
+
+	if (p->flags & GR_PORT_F_PROMISC_TOO_MANY_UNICAST
+	    || p->flags & GR_PORT_F_PROMISC_DEVICE_LIMIT)
+		return 0;
+
+	ret = rte_eth_dev_mac_addr_add(p->port_id, ucast_addr, 0);
+	switch (ret) {
```
rjarry, line 644: `ret` is used only here. Prefer shorter:

```c
switch (rte_eth_dev_mac_addr_add(p->port_id, ucast_addr, 0)) {
```

modules/infra/control/port.c
```diff
@@ -461,6 +508,162 @@ static struct iface_type iface_type_port = {
...
+		return ret;
+		break;
+	}
+
+	return rte_eth_promiscuous_enable(p->port_id);
+}
+
+int gr_port_unicast_addr_del(struct iface_info_port *p, struct rte_ether_addr *ucast_addr) {
+	int ret = rte_eth_dev_mac_addr_remove(p->port_id, ucast_addr);
+	return ret;
```
rjarry, line 664:
```c
return rte_eth_dev_mac_addr_remove(p->port_id, ucast_addr);
```

rjarry (changes requested):

modules/infra/cli/port.c
```diff
@@ -25,6 +25,28 @@ static void port_show(const struct gr_api_client *, const struct gr_iface *iface
...
+		printf(" promisc (too many mac requested)");
+	if (port->flags & GR_PORT_F_PROMISC_DEVICE_LIMIT)
+		printf(" promisc (device)");
+	printf("\n");
+
+	for (int i = 0; i < GR_PORT_NUM_UCAST_ADDR; i++) {
+		if (!eth_addr_eq(&port->unicast[i], &port->mac)
+		    && !eth_addr_is_zero(&port->unicast[i]))
+			printf("additional mac: " ETH_ADDR_FMT "\n",
+			       ETH_BYTES_SPLIT(port->unicast[i].bytes));
```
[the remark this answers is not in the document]
christophefontaine: The mac list always includes the default mac.

modules/infra/control/port.c
christophefontaine: Can't be shortened as  port flags and iface flags are different.

PR #32, opened 2024-07-08
api: add node with name instead of edge

christophefontaine:
Reduce boiler plate code when we add new ether proto or ip proto.

- gr_eth_input_add_type
- ip_input_local_add_proto


rjarry (approved):
Hi Christophe, thanks for the patch! LGTM except one minor detail: please change the commit title `s/API/api/` to appease my git OCD. Thanks!

PR #35, opened 2024-07-09
mbuf: use private data instead of dynfield

rjarry:
The dynfield area is not big enough to fit data for IPv6. Allocate a private and dedicated mbuf area of one cache line (64 bytes) to store up at least two IPv6 addresses and a few pointers.


christophefontaine:
Nice: this allows the usage of Niantic (ixgbe) nics with grout.

PR #37, opened 2024-07-10
graph: sanitize dynamic node attach to parent

rjarry:
Rename node_name to next_node to make it more explicit.

Log the requested operation before actually calling gr_node_attach_parent so that is easier to debug in case of error.

Also update ip_output_add_tunnel to take a node name instead of an edge id.

### DIFF
--- a/modules/infra/datapath/eth_input.c
+++ b/modules/infra/datapath/eth_input.c
@@ -21,12 +21,9 @@ enum {
 
 static rte_edge_t l2l3_edges[1 << 16] = {UNKNOWN_ETHER_TYPE};
 
-void gr_eth_input_add_type(rte_be16_t eth_type, const char *node_name) {
-	rte_edge_t edge = gr_node_attach_parent("eth_input", node_name);
-	if (edge == RTE_EDGE_ID_INVALID)
-		ABORT("gr_node_attach_parent(eth_input, %s) failed", node_name);
-	l2l3_edges[eth_type] = edge;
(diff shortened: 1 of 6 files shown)


modules/infra/datapath/eth_input.c
```diff
@@ -21,12 +21,12 @@ enum {
 
 static rte_edge_t l2l3_edges[1 << 16] = {UNKNOWN_ETHER_TYPE};
 
-void gr_eth_input_add_type(rte_be16_t eth_type, const char *node_name) {
-	rte_edge_t edge = gr_node_attach_parent("eth_input", node_name);
+void gr_eth_input_add_type(rte_be16_t eth_type, const char *next_node) {
+	LOG(DEBUG, "eth_input: type=0x%04x -> %s", rte_be_to_cpu_16(eth_type), next_node);
+	rte_edge_t edge = gr_node_attach_parent("eth_input", next_node);
 	if (edge == RTE_EDGE_ID_INVALID)
```
christophefontaine, line 27: I wonder if we need to check the edge validity, as we abort in gr_node_attach_parent:

[modules/infra/control/graph.c +84
](https://github.com/rjarry/grout/blob/35b5fd76f0f1e1188e66f058bf132ed736a2fb59/modules/infra/control/graph.c#L84)
```
        if (edge == RTE_EDGE_ID_INVALID)
                ABORT("rte_node_edge_update: %s", rte_strerror(rte_errno));
```

rjarry: Indeed, it is redundant. I'll remove this check.

PR #42, opened 2024-07-11
ip: clean resources on iface deletion

christophefontaine:
On iface deletion, delete assoticated routes.
specify module destructor priorities to ensure
resources are not freed when accessing them.

### DIFF
--- a/modules/ip/control/address.c
+++ b/modules/ip/control/address.c
@@ -142,6 +142,19 @@ static struct api_out addr_list(const void *request, void **response) {
 	return api_out(0, len);
 }
 
+static void iface_event_handler(iface_event_t event, struct iface *iface) {
+	struct nexthop *nh;
+
+	if (event != IFACE_EVENT_PRE_REMOVE)
+		return;
+
+	nh = ip4_addr_get(iface->id);
+	if (nh)
+		ip4_route_cleanup(iface->vrf_id, nh);
+
+	addrs[iface->id] = NULL;
+}
+
 static void addr_init(struct event_base *) {
 	addrs = rte_calloc(__func__, MAX_IFACES, sizeof(struct nexthop *), RTE_CACHE_LINE_SIZE);
 	if (addrs == NULL)
@@ -172,6 +185,11 @@ static struct gr_module addr_module = {
 	.name = "ipv4 address",
 	.init = addr_init,
 	.fini = addr_fini,
+	.fini_prio = 2000,
+};
+
+static struct iface_event_handler iface_event_address_handler = {
+	.callback = iface_event_handler,
 };
 
 RTE_INIT(address_constructor) {
@@ -179,4 +197,5 @@ RTE_INIT(address_constructor) {
 	gr_register_api_handler(&addr_del_handler);
 	gr_register_api_handler(&addr_list_handler);
 	gr_register_module(&addr_module);
+	iface_event_register_handler(&iface_event_address_handler);
 }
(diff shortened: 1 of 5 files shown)


modules/ip/control/nexthop.c
```diff
@@ -74,6 +74,17 @@ int ip4_nexthop_add(uint16_t vrf_id, ip4_addr_t ip, uint32_t *idx, struct nextho
...
+
+		rte_spinlock_lock(&nh->lock);
+		// Flush all held packets.
+		struct rte_mbuf *m = nh->held_pkts_head;
+		while (m != NULL) {
+			struct rte_mbuf *next = queue_mbuf_data(m)->next;
+			rte_pktmbuf_free(m);
+			m = next;
+		}
+		rte_spinlock_unlock(&nh->lock);
```
rjarry, line 86: Can you split this into a separate patch? This seems more like a fix related to nexthops.

modules/ip/control/address.c
```diff
@@ -142,6 +142,22 @@ static struct api_out addr_list(const void *request, void **response) {
 	return api_out(0, len);
 }
 
+void route4_cleanup(struct iface *iface);
+void nh4_cleanup(struct iface *iface);
```
rjarry, line 146: These two declarations seem not used.

modules/ip/control/route.c
```diff
@@ -301,6 +301,76 @@ static void route4_fini(struct event_base *) {
...
+			    ((uint8_t *)&ip)[2],
+			    ((uint8_t *)&ip)[3],
+			    prefixlen,
+			    ((uint8_t *)&nh->ip)[0],
+			    ((uint8_t *)&nh->ip)[1],
+			    ((uint8_t *)&nh->ip)[2],
+			    ((uint8_t *)&nh->ip)[3]);
+
+			ip4_route_delete(vrf_id, ip, prefixlen);
+			ip4_route_delete(vrf_id, ip, 32);
```
rjarry, line 340: I don't know if it is safe to delete while iterating.

modules/ip/control/route.c
```diff
@@ -301,6 +301,76 @@ static void route4_fini(struct event_base *) {
...
+			    __func__,
+			    ((uint8_t *)&ip)[0],
+			    ((uint8_t *)&ip)[1],
+			    ((uint8_t *)&ip)[2],
+			    ((uint8_t *)&ip)[3],
+			    prefixlen,
+			    ((uint8_t *)&nh->ip)[0],
+			    ((uint8_t *)&nh->ip)[1],
+			    ((uint8_t *)&nh->ip)[2],
+			    ((uint8_t *)&nh->ip)[3]);
```
rjarry, line 337: Logging IP and ethernet addresses seem quite frequent. We need wrappers around `rte_log` and `snprintf` to reduce boilerplate. E.g.:

```diff
-LOG(DEBUG,
-    "%s ip4_route_delete %d.%d.%d.%d/%d via %d.%d.%d.%d",
-    __func__,
-    ((uint8_t *)&ip)[0],
-    ((uint8_t *)&ip)[1],
-    ((uint8_t *)&ip)[2],
-    ((uint8_t *)&ip)[3],
-    prefixlen,
-    ((uint8_t *)&nh->ip)[0],
-    ((uint8_t *)&nh->ip)[1],
-    ((uint8_t *)&nh->ip)[2],
-    ((uint8_t *)&nh->ip)[3]);
+ LOG(DEBUG, "%4/%hhu via %4", &ip, &nh->ip);
```

Or:

```diff
-scols_line_sprintf(line, 2, ETH_ADDR_FMT, ETH_BYTES_SPLIT(nh->mac.bytes));
+scols_line_sprintf(line, 2, "%2", &nh->mac));
```

What do you think?

rjarry (changes requested):

modules/ip/control/nexthop.c
christophefontaine: indeed.

modules/ip/control/route.c
christophefontaine: I guess this is safe as we iterate over ribs, but ip4_route_delete affects a fib.

rjarry (approved):

PR #49, opened 2024-07-19
add ping v4 & traceroute v4 (ICMP) utilities

christophefontaine:
`traceroute` relies on ICMP, and not UDP.

### DIFF
--- a/modules/infra/control/control_output.c
+++ b/modules/infra/control/control_output.c
@@ -1,0 +1,112 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2024 Christophe Fontaine
+
+#include <gr_control.h>
+#include <gr_control_output.h>
+#include <gr_graph.h>
+#include <gr_log.h>
+#include <gr_macro.h>
+#include <gr_mbuf.h>
+
+#include <event2/event.h>
+#include <rte_ether.h>
+#include <rte_graph_worker.h>
+
+#include <pthread.h>
+#include <stdatomic.h>
+#include <stdlib.h>
+
+static struct rte_ring *ctrlout_ring;
+
+int control_output_enqueue(struct rte_mbuf *m) {
+	return rte_ring_enqueue(ctrlout_ring, m);
+}
+
+static void control_output_poll(evutil_socket_t, short, void *) {
+	struct control_output_mbuf_data *data;
+	struct rte_mbuf *mbuf;
+	void *ring_item;
+
+	while (rte_ring_dequeue(ctrlout_ring, &ring_item) == 0) {
+		mbuf = ring_item;
+		data = control_output_mbuf_data(mbuf);
+		if (data->callback != NULL)
+			control_output_mbuf_data(mbuf)->callback(mbuf);
+		else
+			rte_pktmbuf_free(mbuf);
+	}
+}
+
+static atomic_bool thread_shutdown;
+static pthread_t thread_id;
+static pthread_cond_t cond;
+static pthread_mutex_t mutex;
+static struct event *ctrlout_ev;
+
+void control_output_done(void) {
+	pthread_cond_signal(&cond);
+}
+
+static void *cond_wait_to_event(void *) {
+	pthread_setname_np(pthread_self(), "gr:ctrl-output");
+
+	while (!atomic_load(&thread_shutdown)) {
+		pthread_mutex_lock(&mutex);
+		if (pthread_cond_wait(&cond, &mutex) == 0)
+			evuser_trigger(ctrlout_ev);
+		pthread_mutex_unlock(&mutex);
+	}
+
+	return NULL;
+}
+
+static pthread_attr_t attr;
+
+static void control_output_init(struct event_base *ev_base) {
+	atomic_init(&thread_shutdown, false);
+
+	if (pthread_attr_init(&attr))
+		ABORT("pthread_attr_init");
+
+	if (pthread_mutex_init(&mutex, NULL))
+		ABORT("pthread_mutex_init failed");
+
+	if (pthread_cond_init(&cond, NULL))
+		ABORT("pthread_cond_init failed");
+
+	ctrlout_ring = rte_ring_create(
+		"control_output",
+		RTE_GRAPH_BURST_SIZE * 4,
+		SOCKET_ID_ANY,
+		RING_F_MP_RTS_ENQ | RING_F_SC_DEQ
+	);
+	if (ctrlout_ring == NULL)
+		ABORT("rte_ring_create(ctrl_output): %s", rte_strerror(rte_errno));
+
+	ctrlout_ev = event_new(ev_base, -1, EV_PERSIST | EV_FINALIZE, control_output_poll, NULL);
+	if (ctrlout_ev == NULL)
+		ABORT("event_new() failed");
+
+	pthread_create(&thread_id, &attr, cond_wait_to_event, NULL);
+}
+
+static void control_output_fini(struct event_base *) {
+	atomic_store(&thread_shutdown, true);
+	control_output_done();
+	pthread_join(thread_id, NULL);
+	pthread_attr_destroy(&attr);
+	pthread_cond_destroy(&cond);
+	event_free(ctrlout_ev);
+	rte_ring_free(ctrlout_ring);
+	pthread_mutex_destroy(&mutex);
+}
+
+static struct gr_module control_output_module = {
+	.name = "control_output",
+	.init = control_output_init,
+	.fini = control_output_fini,
+};
+
+RTE_INIT(control_output_module_init) {
+	gr_register_module(&control_output_module);
+}
--- a/modules/infra/datapath/control_output.c
+++ b/modules/infra/datapath/control_output.c
@@ -1,0 +1,47 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2024 Christophe Fontaine
+
+#include "gr_control_output.h"
+
+#include <gr_control.h>
+#include <gr_graph.h>
+#include <gr_log.h>
+#include <gr_mbuf.h>
+
+#include <rte_ether.h>
+#include <rte_graph_worker.h>
+
+#define ERROR 0
+
+static uint16_t control_output_process(
+	struct rte_graph *graph,
+	struct rte_node *node,
+	void **objs,
+	uint16_t n_objs
+) {
+	unsigned sent = 0;
+
+	for (unsigned i = 0; i < n_objs; i++) {
+		if (control_output_enqueue(objs[i]) < 0)
+			rte_node_enqueue_x1(graph, node, ERROR, objs[i]);
+		else
+			sent++;
+	}
+	if (sent > 0)
+		control_output_done();
+
+	return n_objs;
+}
+
+static struct rte_node_register control_output_node = {
+	.name = "control_output",
+	.process = control_output_process,
+	.nb_edges = 1,
+	.next_nodes = {
+		[ERROR] = "control_output_error",
+	},
+};
+
+static struct gr_node_info info __attribute__((used)) = {
+	.node = &control_output_node,
+};
--- a/modules/infra/datapath/control_output.c
+++ b/modules/infra/datapath/control_output.c
@@ -42,6 +42,10 @@ static struct rte_node_register control_output_node = {
 	},
 };
 
-static struct gr_node_info info __attribute__((used)) = {
+static struct gr_node_info info = {
 	.node = &control_output_node,
 };
+
+GR_NODE_REGISTER(info);
+
+GR_DROP_REGISTER(control_output_error);
--- a/modules/infra/datapath/gr_control_output.h
+++ b/modules/infra/datapath/gr_control_output.h
@@ -1,0 +1,33 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2024 Christophe Fontaine
+
+#ifndef _GR_CONTROL_OUTPUT_PATH
+#define _GR_CONTROL_OUTPUT_PATH
+
+#include "gr_mbuf.h"
+
+#include <rte_graph.h>
+
+#include <time.h>
+
+// Callback definition when a packet is sent from the data plane to the control plane.
+// It is up to the function to free the received mbuf.
+//
+// @param struct rte_mbuf *
+//   Packet with the data offset set to the OSI layer of the originating node.
+typedef void (*control_output_cb_t)(struct rte_mbuf *);
+
+GR_MBUF_PRIV_DATA_TYPE(control_output_mbuf_data, {
+	control_output_cb_t callback;
+	clock_t timestamp;
+});
+
+// Enqueue a packet from data plane to a control plane ring.
+//
+// NB: control_output_done() must be called explcitly to wake up the control plane event loop.
+int control_output_enqueue(struct rte_mbuf *m);
+
+// Wake up the control plane event loop so that it processes the pending packets.
+void control_output_done(void);
+
+#endif
--- a/modules/ip/api/gr_ip4.h
+++ b/modules/ip/api/gr_ip4.h
@@ -176,4 +176,35 @@ struct gr_ip4_addr_list_resp {
 	struct gr_ip4_ifaddr addrs[/* n_addrs */];
 };
 
+// icmp ////////////////////////////////////////////////////////////////////////
+
+#define GR_IP4_ICMP_SEND REQUEST_TYPE(GR_IP4_MODULE, 0x0024)
+
+struct gr_ip4_icmp_send_req {
+	ip4_addr_t addr;
+	uint16_t vrf;
+	uint16_t seq_num;
+	uint8_t ttl;
+};
+
+struct gr_ip4_icmp_send_resp {
+	uint16_t id;
+};
+
+#define GR_IP4_ICMP_RECV REQUEST_TYPE(GR_IP4_MODULE, 0x0025)
+
+struct gr_ip4_icmp_recv_req {
+	uint16_t id;
+};
+
+struct gr_ip4_icmp_recv_resp {
+	uint8_t type;
+	uint8_t code;
+	uint8_t ttl;
+	uint16_t ident;
+	uint16_t seq_num;
+	ip4_addr_t src_addr;
+	clock_t response_time;
+};
+
 #endif
--- a/modules/ip/cli/icmp.c
+++ b/modules/ip/cli/icmp.c
@@ -1,0 +1,219 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2024 Christophe Fontaine
+
+#include <gr_api.h>
+#include <gr_cli.h>
+#include <gr_ip4.h>
+#include <gr_net_types.h>
+
+#include <ecoli.h>
+#include <rte_icmp.h>
+#include <rte_ip.h>
+
+#include <errno.h>
+#include <signal.h>
+#include <sys/queue.h>
+#include <unistd.h>
+
+static const char *icmp_dest_unreachable[] = {
+	[RTE_ICMP_CODE_UNREACH_NET] = "net unreachable",
+	[RTE_ICMP_CODE_UNREACH_HOST] = "host unreachable",
+	[RTE_ICMP_CODE_UNREACH_PROTO] = "protocol unreachable",
+	[RTE_ICMP_CODE_UNREACH_PORT] = "port unreachable",
+	[RTE_ICMP_CODE_UNREACH_FRAG] = "fragmentation needed and DF set",
+	[RTE_ICMP_CODE_UNREACH_SRC] = "source route failed",
+};
+
+static bool stop;
+
+static void sighandler(int) {
+	stop = true;
+}
+
+static cmd_status_t icmp_send(
+	const struct gr_api_client *c,
+	struct gr_ip4_icmp_send_req *req,
+	uint16_t msdelay,
+	uint16_t count,
+	bool mode_traceroute
+) {
+	struct gr_ip4_icmp_recv_resp *reply_resp;
+	struct gr_ip4_icmp_recv_req reply_req;
+	struct gr_ip4_icmp_send_resp *start_resp;
+	int timeout, ret, errors;
+	void *resp_ptr = NULL;
+
+	stop = false;
+	errors = 0;
+	errno = 0;
+
+	for (int i = mode_traceroute; i < count && stop == false; i++) {
+		req->ttl = mode_traceroute ? i : 64;
+		req->seq_num = i;
+
+		ret = gr_api_client_send_recv(c, GR_IP4_ICMP_SEND, sizeof(*req), req, &resp_ptr);
+		if (ret < 0)
+			return CMD_ERROR;
+
+		start_resp = resp_ptr;
+		reply_req.id = start_resp->id;
+		free(resp_ptr);
+		resp_ptr = NULL;
+
+		timeout = 50;
+		do {
+			usleep(10000);
+			ret = gr_api_client_send_recv(
+				c, GR_IP4_ICMP_RECV, sizeof(reply_req), &reply_req, &resp_ptr
+			);
+			if (resp_ptr != NULL)
+				break;
+		} while (ret == 0 && --timeout > 0);
+
+		if (ret < 0)
+			return CMD_ERROR;
+
+		reply_resp = resp_ptr;
+		if (reply_resp == NULL || timeout == 0) {
+			errors++;
+			errno = ETIMEDOUT;
+			if (mode_traceroute)
+				printf("%2d  timeout\n", i);
+			else
+				printf("timeout: icmp_seq=%d\n", i);
+		} else {
+			switch (reply_resp->type) {
+			case RTE_ICMP_TYPE_ECHO_REPLY:
+				if (mode_traceroute) {
+					printf("%2d  " IP4_ADDR_FMT " time=%.3f ms\n",
+					       i,
+					       IP4_ADDR_SPLIT(&reply_resp->src_addr),
+					       reply_resp->response_time / 1000.);
+					stop = true;
+					errors = 0;
+					errno = 0;
+				} else {
+					printf("reply from " IP4_ADDR_FMT ": icmp_seq=%d ttl=%d "
+					       "time=%.3f ms\n",
+					       IP4_ADDR_SPLIT(&reply_resp->src_addr),
+					       reply_resp->seq_num,
+					       reply_resp->ttl,
+					       reply_resp->response_time / 1000.);
+				}
+				break;
+			case RTE_ICMP_TYPE_DEST_UNREACHABLE:
+				errors++;
+				errno = EHOSTUNREACH;
+				printf("reply from " IP4_ADDR_FMT ": icmp_seq=%d ttl=%d: %s\n",
+				       IP4_ADDR_SPLIT(&reply_resp->src_addr),
+				       reply_resp->seq_num,
+				       reply_resp->ttl,
+				       icmp_dest_unreachable[reply_resp->code]);
+				break;
+			case RTE_ICMP_TYPE_TTL_EXCEEDED:
+				errors++;
+				errno = ETIMEDOUT;
+				printf("%2d  " IP4_ADDR_FMT " time=%.3f ms\n",
+				       i,
+				       IP4_ADDR_SPLIT(&reply_resp->src_addr),
+				       reply_resp->response_time / 1000.);
+				break;
+			}
+			free(resp_ptr);
+			resp_ptr = NULL;
+		}
+		usleep(msdelay * 1000);
+	}
+
+	if (!stop && errors > 0)
+		return CMD_ERROR;
+
+	return CMD_SUCCESS;
+}
+
+static cmd_status_t ping(const struct gr_api_client *c, const struct ec_pnode *p) {
+	struct gr_ip4_icmp_send_req req = {.seq_num = 0, .vrf = 0};
+	cmd_status_t ret = CMD_ERROR;
+	uint16_t count = UINT16_MAX;
+	uint16_t msdelay = 1000;
+
+	if (inet_pton(AF_INET, arg_str(p, "IP"), &req.addr) != 1) {
+		errno = EINVAL;
+		return CMD_ERROR;
+	}
+	if ((ret = arg_u16(p, "VRF", &req.vrf)) < 0 && ret != ENOENT)
+		return CMD_ERROR;
+	if ((ret = arg_u16(p, "COUNT", &count)) < 0 && ret != ENOENT)
+		return CMD_ERROR;
+	if ((ret = arg_u16(p, "DELAY", &msdelay)) < 0 && ret != ENOENT)
+		return CMD_ERROR;
+
+	sighandler_t prev_handler = signal(SIGINT, sighandler);
+	if (prev_handler == SIG_ERR)
+		return CMD_ERROR;
+
+	ret = icmp_send(c, &req, msdelay, count, false);
+
+	signal(SIGINT, prev_handler);
+
+	return ret;
+}
+
+static cmd_status_t traceroute(const struct gr_api_client *c, const struct ec_pnode *p) {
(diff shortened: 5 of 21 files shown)


christophefontaine:
Rebase, but switched back to draft, as I need to get the proper source IP address to emit the ICMP echo request, and not the default.

.github/workflows/ci.yml
```diff
@@ -52,7 +52,8 @@ jobs:
           sudo NEEDRESTART_MODE=l apt-get install -qy --no-install-recommends \
             git socat tcpdump traceroute \
             iproute2 iputils-ping libasan8 libedit2 \
-            libevent-2.1-7t64 libsmartcols1 libnuma1
+            libevent-2.1-7t64 libsmartcols1 libnuma1 \
+            traceroute
```
rjarry, line 56: `traceroute` is already installed line 53. You can drop this chunk.

modules/infra/control/mempool.c
```diff
@@ -6,6 +6,8 @@
 #include <gr_log.h>
 #include <gr_mbuf.h>
 
+#include <stdlib.h>
+
```
rjarry, line 10: Is this still required?

modules/infra/datapath/gr_control_output.h
```diff
@@ -0,0 +1,36 @@
...
+
+GR_MBUF_PRIV_DATA_TYPE(control_output_mbuf_data, {
+	control_output_cb_t callback;
+	clock_t timestamp;
+});
+
+struct gr_control_output_msg {
+	control_output_cb_t callback;
+	struct rte_mbuf *mbuf;
+} __rte_packed;
```
rjarry, line 33: Can you remove the __packed attribute here? The message will not travel through the wire.

modules/infra/control/control_output.c
```diff
@@ -0,0 +1,104 @@
...
+
+#include <event2/event.h>
+#include <rte_ether.h>
+#include <rte_graph_worker.h>
+
+#include <pthread.h>
+#include <stdatomic.h>
+#include <stdlib.h>
+
+extern struct rte_ring *control_output_ring;
```
rjarry, line 19: This ring should be a static variable in this module. And preferably with a shorter name. You should expose a public function to enqueue packets from dataplane without exposing the ring variable.

modules/infra/control/control_output.c
```diff
@@ -0,0 +1,104 @@
...
+static struct event *control_output_evt;
```
rjarry, line 20: Same remark about the variable name: `ctrlout_ev`

modules/infra/control/control_output.c
```diff
@@ -0,0 +1,104 @@
...
+pthread_mutex_t mp;
+
+static void poll_control_output_ring(evutil_socket_t, short, void *) {
+	struct gr_control_output_msg msg[RTE_GRAPH_BURST_SIZE];
+	int n = 0;
+
+	while ((n = rte_ring_dequeue_burst_elem(
+			control_output_ring, msg, sizeof(msg[0]), ARRAY_DIM(msg), NULL
+		))
+	       > 0) {
```
rjarry, line 34: Hopefully, shortening the ring variable name will make this less ugly ...

modules/infra/control/control_output.c
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+void signal_control_ouput_message() {
+	pthread_cond_signal(&cv);
+}
+
+static void *cond_wait_to_event(void *) {
+	struct timespec ts;
+	pthread_setname_np(pthread_self(), "gr:cond_wait_to_event");
```
rjarry, line 47: `cond_wait_to_event` seems overly focused on the implementation and not on the actual purpose of that thread. How about `"gr:ctrl-output"`?

modules/infra/control/control_output.c
```diff
@@ -0,0 +1,104 @@
...
+		ABORT("pthread_attr_init");
+	}
+
+	if (pthread_mutex_init(&mp, NULL)) {
+		ABORT("pthread_mutex_init failed");
+	}
+
+	if (pthread_cond_init(&cv, NULL)) {
+		ABORT("pthread_cond_init failed");
+	}
```
rjarry, line 75: Can we drop the braces here?

modules/infra/datapath/control_output.c
```diff
@@ -0,0 +1,82 @@
...
+) {
+	struct gr_control_output_msg msg;
+	struct rte_mbuf *mbuf;
+
+	for (unsigned i = 0; i < n_objs; i++) {
+		mbuf = objs[i];
+		msg.callback = control_output_mbuf_data(mbuf)->callback;
+		msg.mbuf = mbuf;
+
+		if (msg.callback == NULL
```
rjarry, line 37: It would be clearer to split in two separate tests with possibly two error nodes.

`NO_CALLBACK_ERROR`
`ENQUEUE_ERROR`

modules/infra/datapath/control_output.c
```diff
@@ -0,0 +1,82 @@
...
+		    || rte_ring_enqueue_elem(
+			       control_output_ring, &msg, sizeof(struct gr_control_output_msg)
+		       ) != 0) {
```
rjarry, line 40: hopefully, this will be less ugly with the public function exposed by the control plane.

modules/infra/datapath/control_output.c
```diff
@@ -0,0 +1,82 @@
...
+static void control_output_register(void) {
+	control_output_ring = rte_ring_create_elem(
+		"control_output",
+		sizeof(struct gr_control_output_msg),
+		RTE_GRAPH_BURST_SIZE * 4,
+		SOCKET_ID_ANY,
+		RING_F_MP_RTS_ENQ | RING_F_MC_RTS_DEQ
+	);
+	if (control_output_ring == NULL)
+		ABORT("rte_ring_create(arp_output_request): %s", rte_strerror(rte_errno));
```
rjarry, line 60: As already mentioned, this should be allocated by the control plane. And `rte_ring_create(arp_output_request)` is most likely a typo :)

modules/ip/cli/icmp.c
```diff
@@ -0,0 +1,220 @@
...
+	signal(SIGINT, prev_handler);
+	return ret_code;
+}
+
+static int ctx_init(struct ec_node *root) {
+	int ret;
+
+	ret = CLI_COMMAND(
+		CLI_CONTEXT(root, CTX_ARG("ping", "IPv4 icmp request")),
+		"IP [vrf VRF] [-c COUNT] [-w DELAY]",
```
rjarry, line 192: To be consistent with all other arguments, can we have this?

```c
"IP [vrf VRF] [count COUNT] [delay DELAY]",
```

modules/ip/cli/icmp.c
```diff
@@ -0,0 +1,220 @@
...
+		ping,
+		"Send ICMP request to destination IP.",
+		with_help("IPv4 address.", ec_node_re("IP", IPV4_RE)),
+		with_help("L3 routing domain ID.", ec_node_uint("VRF", 0, UINT16_MAX - 1, 10)),
+		with_help("-c number of packets to send.", ec_node_uint("COUNT", 1, 10000, 10)),
```
rjarry, line 197: `s/-c n/N/`

modules/ip/cli/icmp.c
```diff
@@ -0,0 +1,220 @@
...
+		with_help("-w delay in ms between icmp echo.", ec_node_uint("DELAY", 0, 10000, 10))
```
rjarry, line 198: `s/-w d/D/`

rjarry (changes requested):
Hi Chris! Thanks for rebasing!

I've made a first pass. I will address the rest later.

modules/ip/cli/icmp.c
```diff
@@ -0,0 +1,220 @@
...
+static const char *ICMP_DEST_UNREACHABLE[] = {
+	"net unreachable",
+	"host unreachable",
+	"protocol unreachable"
+	"port unreachable",
+	"fragmentation needed and DF set",
+	"source route failed",
+};
+
+static bool stop = false;
```
rjarry, line 49: no need for initialization here. Static data is always initialized to zero (false).

modules/ip/datapath/icmp_output_request.c
```diff
@@ -0,0 +1,131 @@
...
+			rte_pktmbuf_append(mbuf, sizeof(struct rte_icmp_hdr) + sizeof(clock_t));
+
+		payload = rte_pktmbuf_mtod_offset(mbuf, clock_t *, sizeof(struct rte_icmp_hdr));
+		*payload = clock();
+
+		// Build ICMP packet
+		icmp->icmp_type = RTE_IP_ICMP_ECHO_REQUEST;
+		icmp->icmp_code = 0;
+		icmp->icmp_seq_nb = rte_cpu_to_be_16(msg->seq_num);
+		icmp->icmp_ident = 0x4242;
```
rjarry, line 94: We should use that ICMP identifier to demux responses and dispatch them to the correct client.

modules/ip/datapath/ip_error.c
```diff
@@ -44,18 +45,15 @@ ip_error_process(struct rte_graph *graph, struct rte_node *node, void **objs, ui
...
-			goto next;
-		}
-		if ((nh = ip4_addr_get_preferred(iface->id, ip->src_addr)) == NULL) {
-			edge = NO_IP;
-			goto next;
+		input_iface = ip_output_mbuf_data(mbuf)->input_iface;
+		gw = ip4_route_lookup(input_iface->vrf_id, ip->src_addr);
+		if (gw == NULL || (nh = ip4_addr_get_preferred(input_iface->id, gw->ip)) == NULL) {
+			rte_node_enqueue_x1(graph, node, NO_IP, mbuf);
+			continue;
```
rjarry, line 52: I prefer keeping the checks for `input_iface` and `nh` separate.

rjarry: And there is a goto label. Do not use `continue` here.

modules/ip/datapath/ip_error.c
```diff
@@ -23,11 +23,12 @@ enum edges {
 static uint16_t
 ip_error_process(struct rte_graph *graph, struct rte_node *node, void **objs, uint16_t nb_objs) {
 	struct ip_local_mbuf_data *ip_data;
-	const struct iface *iface;
+	const struct iface *input_iface;
```
rjarry, line 26: `s/input_iface/iface/`

rjarry (changes requested):

christophefontaine:
On PTO til next week, will update the branch as soon as I get back.

modules/ip/cli/icmp.c
```diff
@@ -0,0 +1,220 @@
...
+	uint8_t type_of_service; /**< type of service */
+	uint16_t total_length; /**< length of packet */
+	uint16_t packet_id; /**< packet ID */
+	uint16_t fragment_offset; /**< fragmentation offset */
+	uint8_t time_to_live; /**< time to live */
+	uint8_t next_proto_id; /**< protocol ID */
+	uint16_t hdr_checksum; /**< header checksum */
+	uint32_t src_addr; /**< source address */
+	uint32_t dst_addr; /**< destination address */
+} __attribute__((packed));
```
rjarry, line 38: The cli is now complied with the dpdk headers available. You can reuse rte_ip.h and avoid copying the structure here.

smoke/ip_forward_error_test.sh
```diff
@@ -0,0 +1,30 @@
...
+	ip -n $p link set $p address ba:d0:ca:ca:00:0$n
+	ip -n $p link set $p up
+	ip -n $p addr add 172.16.$n.2/24 dev $p
+	ip -n $p route add default via 172.16.$n.1
+	ip -n $p addr show
+done
+
+ip netns exec $p0 ping -i0.01 -c3 172.16.1.2
+ip netns exec $p1 ping -i0.01 -c3 172.16.0.2
+ip netns exec $p0 traceroute 172.16.1.2
```
rjarry, line 30: I think this test is redundant with the existing ip forwarding tests which already runs traceroute.

However, it would be nice to have a test where you initiate a ping and a traceroute from `grcli`.

modules/ip/cli/icmp.c
```diff
@@ -0,0 +1,220 @@
...
+
+		if (gr_api_client_send_recv(c, GR_IP4_ICMP_REQUEST, sizeof(*req), req, NULL) < 0) {
+			ret_code = CMD_ERROR;
+			goto exit;
+		}
+
+		timeout = 50;
+		resp_ptr = NULL;
+		do {
+			usleep(10000);
```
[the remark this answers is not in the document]
christophefontaine: We'll never get the icmp response as we just sent the request, so it was a good tradeoff to wait initially, instead of waiting at the end of the loop.

modules/ip/cli/icmp.c
```diff
@@ -0,0 +1,220 @@
...
+			ret = gr_api_client_send_recv(c, GR_IP4_ICMP_GET_REPLY, 0, NULL, &resp_ptr);
```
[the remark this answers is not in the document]
christophefontaine: not yet, I'll update the PR with the fixes for the other comments and add the support for multiple clients in a subsequent patch.

modules/infra/control/control_output.c
david-marchand: Thread names are limited to 16 chars. So shorter is better.

modules/infra/control/control_output.c
```diff
@@ -0,0 +1,104 @@
...
+// Copyright (c) 2024 Christophe Fontaine
+
+#include <gr_control.h>
+#include <gr_control_output.h>
+#include <gr_graph.h>
+#include <gr_log.h>
+#include <gr_macro.h>
+#include <gr_mbuf.h>
+
+#include <event2/event.h>
```
david-marchand, line 11: Move this include with non dpdk incluses.

modules/infra/control/control_output.c
```diff
@@ -0,0 +1,104 @@
...
+#include <rte_graph_worker.h>
+
+#include <pthread.h>
+#include <stdatomic.h>
+#include <stdlib.h>
+
+extern struct rte_ring *control_output_ring;
+static struct event *control_output_evt;
+
+static int tshutdown;
```
david-marchand, line 22: bool ?

modules/infra/control/control_output.c
```diff
@@ -0,0 +1,104 @@
...
+pthread_t thread_id;
+pthread_cond_t cv;
+pthread_mutex_t mp;
+
+static void poll_control_output_ring(evutil_socket_t, short, void *) {
+	struct gr_control_output_msg msg[RTE_GRAPH_BURST_SIZE];
+	int n = 0;
```
david-marchand, line 33: No need for init. n is stored to right away.

modules/ipip/datapath_out.c
```diff
@@ -57,6 +57,7 @@ ipip_output_process(struct rte_graph *graph, struct rte_node *node, void **objs,
 		tunnel.len = rte_be_to_cpu_16(inner->total_length);
 		tunnel.vrf_id = iface->vrf_id;
 		tunnel.proto = IPPROTO_IPIP;
+		tunnel.ttl = IPV4_DEFAULT_TTL;
```
david-marchand, line 60: Strange that this issue is seen only now. Is this with a higher level of optimisations at compilation?

modules/ip/datapath/ip_error.c
```diff
@@ -45,11 +46,15 @@ ip_error_process(struct rte_graph *graph, struct rte_node *node, void **objs, ui
 
 		// Get the local router IP address from the input iface
 		iface = ip_output_mbuf_data(mbuf)->input_iface;
-		if (iface == NULL) {
+		gw = ip4_route_lookup(iface->vrf_id, ip->src_addr);
```
david-marchand, line 49: Just to make sure I understand the fix.
Previously a icmp error would be emitted with the ip address of the interface on which the packet was received.

However, this interface may have no address at all (but I can't picture this but only for loopback ifaces, that grout does not support yet?). Is there other cases?

modules/infra/control/control_output.c
```diff
@@ -0,0 +1,104 @@
...
+
+	while ((n = rte_ring_dequeue_burst_elem(
+			control_output_ring, msg, sizeof(msg[0]), ARRAY_DIM(msg), NULL
+		))
+	       > 0) {
+		for (int i = 0; i < n; i++) {
+			msg[i].callback(msg[i].mbuf);
```
david-marchand, line 40: Just an open question.
Do we want to batch those callback calls?

modules/infra/control/control_output.c
```diff
@@ -0,0 +1,104 @@
...
+	pthread_cond_signal(&cv);
+}
+
+static void *cond_wait_to_event(void *) {
+	struct timespec ts;
+	pthread_setname_np(pthread_self(), "gr:cond_wait_to_event");
+	clock_gettime(CLOCK_REALTIME, &ts);
+
+	while (atomic_load_explicit(&tshutdown, memory_order_acquire) == 0) {
+		ts.tv_sec += 1;
```
david-marchand, line 55: Probably not really important, but there will be some deviation over time.

modules/infra/control/control_output.c
```diff
@@ -0,0 +1,104 @@
...
+		pthread_mutex_lock(&mp);
+		if (pthread_cond_timedwait(&cv, &mp, &ts) == 0) {
+			evuser_trigger(control_output_evt);
+		}
+		pthread_mutex_unlock(&mp);
+	}
+	return NULL;
+}
+
+pthread_attr_t attr;
```
david-marchand, line 65: static ?

modules/infra/datapath/control_output.c
```diff
@@ -0,0 +1,82 @@
...
+#include <rte_graph_worker.h>
+
+#include <stdlib.h>
+
+enum {
+	CONTROL_OUTPUT_ERROR,
+	EDGE_COUNT,
+};
+
+struct rte_ring *control_output_ring;
```
david-marchand: static?

modules/infra/datapath/gr_control_output.h
```diff
@@ -0,0 +1,27 @@
...
+#ifndef _GR_CONTROL_OUTPUT_PATH
+#define _GR_CONTROL_OUTPUT_PATH
+
+#include "gr_mbuf.h"
+
+#include <rte_graph.h>
+
+#include <time.h>
+
+#define CONTROL_OUTPUT_UNKNOWN (uint8_t)0xff
```
david-marchand, line 13: CONTROL_OUTPUT_UNKNOWN is unused.

modules/infra/control/control_output.c
```diff
@@ -0,0 +1,104 @@
...
+
+#include <pthread.h>
+#include <stdatomic.h>
+#include <stdlib.h>
+
+extern struct rte_ring *control_output_ring;
+static struct event *control_output_evt;
+
+static int tshutdown;
+pthread_t thread_id;
```
david-marchand: static? (for the whole block of vars).

modules/infra/control/control_output.c
david-marchand: Ugly. This should be in some header.

modules/infra/control/control_output.c
```diff
@@ -16,20 +16,24 @@
 #include <stdatomic.h>
 #include <stdlib.h>
 
-extern struct rte_ring *control_output_ring;
-static struct event *control_output_evt;
+static struct rte_ring *ctrlout_ring;
```
david-marchand, line 19: This can be moved to previous patch.

modules/ip/cli/icmp.c
```diff
@@ -0,0 +1,205 @@
...
+					       resp->sequence_number,
+					       resp->ttl,
+					       resp->response_time / 1000.);
+				}
+				break;
+			case 3:
+				printf("Destination unreachable: %s\n",
+				       ICMP_DEST_UNREACHABLE[resp->code]);
+				break;
+			case 11:
```
david-marchand, line 101: Worth some define. I am not familiar with the meaning of 11 integer in this case.

modules/ip/cli/icmp.c
```diff
@@ -0,0 +1,205 @@
...
+	uint16_t msdelay = 1000;
+	uint16_t count = 4;
+
+	if (inet_pton(AF_INET, arg_str(p, "IP"), &req.addr) != 1) {
+		errno = EINVAL;
+		return CMD_ERROR;
+	}
+
+	stop = false;
+	sighandler_t prev_handler = signal(SIGINT, sighandler);
```
david-marchand, line 129: I think signal() is not that portable. And sigaction is usually preferred. Not sure it is important for grout though.

modules/ip/control/icmp.c
```diff
@@ -0,0 +1,146 @@
...
+			icmp = rte_pktmbuf_mtod(replies[i], struct rte_icmp_hdr *);
+			if (icmp->icmp_ident == htons(id))
+				return atomic_exchange(&replies[i], null_mbuf);
+		}
+	}
+	return NULL;
+}
+
+static void icmp_response(struct rte_mbuf *m) {
+	struct rte_mbuf *previous = atomic_exchange(&replies[reply_index++], m);
```
david-marchand, line 41: Not sure if incrementing reply_index this way is safe.
Can't we use a little ring instead?

modules/ip/control/icmp.c
```diff
@@ -0,0 +1,146 @@
...
+
+static void dissect_icmp_response(struct rte_mbuf *icmp_reply, struct gr_ip4_icmp_response *resp) {
+	struct rte_ipv4_hdr *inner_ip;
+	struct rte_icmp_hdr *icmp;
+	struct rte_ipv4_hdr *ip;
+	clock_t delay, *sent;
+
+	if (icmp_reply) {
+		ip = rte_pktmbuf_mtod_offset(
+			icmp_reply, struct rte_ipv4_hdr *, -sizeof(struct rte_ipv4_hdr)
```
david-marchand, line 53: Ugly.
There was a time when anything before the current data offset could be scrambled by some architecture specific implementation of the mbuf API.
Doing such hack probably works with DPDK, but still, ugly.

modules/ip/control/icmp.c
```diff
@@ -0,0 +1,146 @@
...
+		);
+		icmp = rte_pktmbuf_mtod(icmp_reply, struct rte_icmp_hdr *);
+		resp->answered = true;
+		resp->ident = icmp->icmp_ident;
+		resp->type = icmp->icmp_type;
+		resp->code = icmp->icmp_code;
+		resp->sequence_number = ntohs(icmp->icmp_seq_nb);
+		resp->ttl = ip->time_to_live;
```
david-marchand, line 61: Can this be stored as metadata instead of going back in the packet data?

smoke/ip_builtin_icmp.sh
```diff
@@ -0,0 +1,36 @@
...
+	echo ip netns del $p >> $tmp/cleanup
+	ip link set $p netns $p
+	ip -n $p link set $p address ba:d0:ca:ca:00:0$n
+	ip -n $p link set $p up
+	ip -n $p addr add 172.16.$n.2/24 dev $p
+	ip -n $p route add default via 172.16.$n.1
+	ip -n $p addr show
+done
+
+grcli ping 172.16.0.2 &
```
david-marchand, line 28: Ideally, we should join this subshell as some point in the cleanup of this script.

modules/infra/control/control_output.c
rjarry: the header sorting is done by clang-format:

https://github.com/DPDK/grout/blob/2d90e7ccdd257a5155a47a62ea9304910d6b7038/.clang-format#L89-L100

modules/infra/datapath/control_output.c
rjarry: This variable should be removed from here. See my previous review.

modules/infra/control/control_output.c
christophefontaine: yes, fixed in the new version.

modules/infra/control/control_output.c
christophefontaine: it depends on clang version though. (I applied clang format, it didn't fixed it).
Same as '-Wmissing-variable-declarations' which isn't supported in my system. (RHEL 9.4, gcc version 11.4.1 ...)

modules/ipip/datapath_out.c
christophefontaine: No, with the default option.
I don't know why it is triggered only now though. (And only on my system !?)

modules/infra/control/control_output.c
christophefontaine: this should be only running on the control plane thread, and I don't expect (yet) a huge amount of traffic between the 2.
I suggest to revisit that when we'll hit a performance issue.

modules/infra/control/control_output.c
christophefontaine: yes, fixed.

modules/infra/datapath/control_output.c
christophefontaine: fixed in the newer version, `control_output_push` is now hiding the ring.

modules/infra/control/control_output.c
christophefontaine: yes, fixed.

modules/infra/control/control_output.c
christophefontaine: indeed ! I messed with the 2 patches. Will fix that.

modules/ip/control/icmp.c
christophefontaine: will do.

modules/ip/control/icmp.c
christophefontaine: hell yeah, this will be much better.

smoke/ip_builtin_icmp.sh
christophefontaine: ack !

modules/infra/control/control_output.c
christophefontaine: yup.

modules/ip/cli/icmp.c
christophefontaine: signal is already used in other parts of grout.
Maybe we should fix all them in a separate patch ?

modules/infra/control/control_output.c
rjarry: Your formatting correctly follows what is configured for clang-format. I was replying to David's remark :)

modules/infra/control/control_output.c
david-marchand: Sure.

modules/ip/cli/icmp.c
david-marchand: Ok for now.

modules/ipip/datapath_out.c
david-marchand: Which gcc, which distro?

christophefontaine: rhel 9.4
gcc version 11.4.1 20231218 (Red Hat 11.4.1-3) (GCC)

modules/ip/control/icmp.c
christophefontaine: I used a preallocated array with stb_ds, hope this will do the job just as fine.

modules/ipip/datapath_out.c
david-marchand: Ok, I prefer such a reason :-).

david-marchand:
Looking at last PR, patch 4 should be squashed in patch 3, since it is the introduction of ttl field in ip_local_mbuf_data that triggers this compilation issue.

modules/infra/control/control_output.c
```diff
@@ -0,0 +1,117 @@
...
+}
+
+static void poll_control_output_ring(evutil_socket_t, short, void *) {
+	struct gr_control_output_msg msg[RTE_GRAPH_BURST_SIZE];
+	int n;
+
+	while ((n = rte_ring_dequeue_burst_elem(
+			ctrlout_ring, msg, sizeof(msg[0]), ARRAY_DIM(msg), NULL
+		))
+	       > 0) {
```
rjarry, line 38: Automatic formatting is a bit ugly here...

I wish we could do better but I don't see how.

modules/infra/datapath/gr_control_output.h
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @param struct rte_mbuf *
+ *   pointer to a mbuf, with the data offset set to the osi layer of
+ *   the node which punted the packet.
+ */
+typedef void (*control_output_cb_t)(struct rte_mbuf *);
+
+GR_MBUF_PRIV_DATA_TYPE(control_output_mbuf_data, {
+	control_output_cb_t callback;
+	clock_t timestamp;
```
rjarry, line 26: I don't see the timestamp used in this commit. Maybe it should move after?

modules/infra/datapath/control_output.c
```diff
@@ -0,0 +1,65 @@
...
+		signal_control_ouput_message();
+
+	return n_objs;
+}
+
+static struct rte_node_register control_output_node = {
+	.name = "control_output",
+	.process = control_output_process,
+	.nb_edges = EDGE_COUNT,
+	.next_nodes = {[NO_CALLBACK_ERROR] = "ctrl_out_no_callback", [ENQUEUE_ERROR] = "ctrl_out_enqueue"},
```
rjarry, line 55: not really important, but can you move to one edge per line, like the other nodes?

modules/infra/datapath/gr_control_output.h
```diff
@@ -23,6 +24,10 @@ typedef void (*control_output_cb_t)(struct rte_mbuf *);
 GR_MBUF_PRIV_DATA_TYPE(control_output_mbuf_data, {
 	control_output_cb_t callback;
 	clock_t timestamp;
+	// Node specific data
+	union {
+		struct rte_icmp_hdr;
+	};
```
rjarry, line 30: That is not a good idea to have icmp references in a generic control output node data. I would prefer a generic buffer that will be mapped to something else via the callback.

modules/ip/api/gr_ip4.h
```diff
@@ -176,4 +176,34 @@ struct gr_ip4_addr_list_resp {
...
+	uint16_t code;
+	uint16_t sequence_number;
+	uint16_t ttl;
+	clock_t response_time;
+	uint8_t data[64];
+};
+
+// ICMP types not defined in rte_icmp.h
+#define GR_IP_ICMP_DEST_UNREACHABLE 3
+#define GR_IP_ICMP_TTL_EXCEEDED 11
```
rjarry, line 207: For consistency with the rest of the API, could you rename the symbols as follows:

```c
#define GR_IP4_PING_START REQUEST_TYPE(GR_IP4_MODULE, 0x0024)

struct gr_ip4_ping_start_req {
...
};

struct gr_ip4_ping_start_resp {
	uint16_t id;
};

#define GR_IP4_PING_GET_REPLY REQUEST_TYPE(GR_IP4_MODULE, 0x0025)

// struct gr_ip4_ping_get_reply_req { };

struct gr_ip4_ping_get_reply_resp {
...
};
```

modules/ip/cli/icmp.c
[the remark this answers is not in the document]
rjarry: But shouldn't you usleep() for the duration of the msdelay?

rjarry (changes requested):

modules/ip/cli/icmp.c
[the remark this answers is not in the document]
christophefontaine: we have a fixed timeout of 500ms.
msdelay is the time between icmp probes.
Do you prefer to also have a configurable timeout ?

modules/infra/control/control_output.c
christophefontaine: I'll remove the dequeue burst, and will dequeue 1 el at a time.
This shouldn't be a critical part of the code considering the number of elements.

modules/infra/datapath/gr_control_output.h
christophefontaine: done, moved everything as part of private data of the mbuf.

PR #57, opened 2024-08-01
Fib nh ptr

rjarry:


modules/ip/control/address.c
```diff
@@ -87,22 +86,24 @@ static struct api_out addr_add(const void *request, void **response) {
 	if (ifaddrs->count == IP4_HOPLIST_MAX_SIZE)
 		return api_out(ENOSPC, 0);
 
-	if (ip4_nexthop_lookup(iface->vrf_id, req->addr.addr.ip, &nh_idx, &nh) == 0)
+	if (ip4_nexthop_lookup(iface->vrf_id, req->addr.addr.ip) != NULL)
 		return api_out(EADDRINUSE, 0);
 
-	if ((ret = ip4_nexthop_add(iface->vrf_id, req->addr.addr.ip, &nh_idx, &nh)) < 0)
-		return api_out(-ret, 0);
+	if ((nh = ip4_nexthop_get_new()) == NULL)
```
christophefontaine, line 92: Is there a case where we would need to get a new nh from the mempool without initialisizing it just after ?

I feel that the `mempool_get` could be hidden in `ip4_nexthop_init` call.

rjarry: Good point. I'll change that.

modules/infra/control/vlan.c
```diff
@@ -154,8 +164,10 @@ static int iface_vlan_init(struct iface *iface, const void *api_info) {
 		iface, IFACE_SET_ALL, iface->flags, iface->mtu, iface->vrf_id, api_info
 	);
 	if (ret < 0) {
+		ret = errno;
 		iface_vlan_fini(iface);
-		errno = -ret;
+		errno = ret;
+		ret = -ret;
```
christophefontaine, line 170: I'm not sure what's happening here ?

modules/infra/control/iface.c
```diff
@@ -34,10 +34,13 @@ void iface_type_register(struct iface_type *type) {
...
 }
 
+#define IFACE_ID_FIRST GR_IFACE_ID_UNDEF + 1
+
+// the first slot is wasted by GR_IFACE_ID_UNDEF
 static struct iface **ifaces;
 
 static int next_ifid(uint16_t *ifid) {
-	for (uint16_t i = 0; i < MAX_IFACES; i++) {
+	for (uint16_t i = IFACE_ID_FIRST; i < MAX_IFACES; i++) {
```
christophefontaine, line 43: Should we set GR_IFACE_ID_UNDEF to -1, and have MAX_IFACE to ((1<<16) -1) ?

modules/ip/control/nexthop.c
```diff
@@ -10,71 +10,58 @@
...
-		return errno_set(-nh_idx);
+	if ((ret = rte_mempool_get(nh_pool, &data)) < 0)
+		return errno_set_null(-ret);
 
-	*idx = nh_idx;
-	*nh = &nh_array[nh_idx];
+	nh = data;
+	nh->vrf_id = vrf_id;
+	nh->iface_id = iface_id;
+	nh->ip = ip;
```
christophefontaine, line 38: `nh->ref_count = 1` ?

modules/infra/control/iface.c
rjarry: My intent was to ensure that non-initialized values of iface_id (0) are invalid.

modules/ip/control/nexthop.c
rjarry: I thought about it. The issue is that ip4_route_insert() also increments the ref count. It will result in nexthop leaking. I don't have an elegant solution to this.

modules/infra/control/vlan.c
rjarry: The idea was to preserve errno. The iface_reconfig() function may return -1 which would be mistaken as EPERM. I'll rework this to make it prettier.

PR #64, opened 2024-09-06
worker: allow starting threads outside of current process affinity

rjarry:
When using tuned CPU partitioning, the default CPU affinity of processes are often restricted to a set of housekeeping cores. We want the control plane threads to run in these CPUs. However, we also want the data path threads NOT to run on these CPUs, but instead pin them on the previously isolated cores.

Remove the restriction of worker_rxq_assign() cpu_id to be part of the current process CPU affinity. Instead use a pthread attribute to specify the desired affinity based on cpu_id and pass that to pthread_create(). That way, if the cpu_id cannot be used (either because it is invalid or referencing to an offline CPU), pthread_create() will return an appropriate error code that can be returned to the caller. Since this call can now fail, delay the removal of the rxq from the source worker after the destination has been successfully created.

Remove the (now redundant) call to rte_thread_set_affinity() inside gr_datapath_loop().

Update unit tests accordingly.

### DIFF
--- a/main/dpdk.c
+++ b/main/dpdk.c
@@ -9,13 +9,13 @@
 #include <gr_log.h>
 #include <gr_stb_ds.h>
 
-#include <numa.h>
 #include <rte_eal.h>
 #include <rte_errno.h>
 #include <rte_log.h>
 #include <rte_mempool.h>
 #include <rte_version.h>
 
+#include <pthread.h>
 #include <sched.h>
 #include <stdio.h>
 #include <syslog.h>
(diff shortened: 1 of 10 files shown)


main/dpdk.c
```diff
@@ -66,25 +66,36 @@ static ssize_t log_write(void *, const char *buf, size_t size) {
...
+	gr_rte_log_type = rte_log_register_type_and_pick_level("grout", RTE_LOG_NOTICE);
+	if (gr_rte_log_type < 0)
+		return errno_log(-gr_rte_log_type, "rte_log_register_type_and_pick_level");
+
+	if ((log_stream = fopencookie(NULL, "w+", log_functions)) == NULL)
+		return errno_log(errno, "fopencookie");
+
 	rte_openlog_stream(log_stream);
+	if (args->log_level >= RTE_LOG_DEBUG)
+		rte_log_set_global_level(RTE_LOG_DEBUG);
```
david-marchand, line 87: Setting global log <level> it not equivalent to passing --log-level=*:<level>.
Doing this change will raise the global filter to <level>, but no DPDK component will have their log level touched.

IOW, if a DPDK component default log level is WARNING, then we will never get its debug level logs, while it was the case before this patch.

rjarry: I mad misunderstood what this `--log-level` option is meant for. I will change this to `rte_log_set_level_pattern("*", ...)`. Is there any reason to change the global level?

david-marchand: This global level filters all logs to a certain level, regardless of what a logtype is configured with. Its use is quite limited (I see it as useless) and it is best to use rte_log_set_("*") like before.

PR #67, opened 2024-09-13
Sleep

rjarry:


modules/infra/datapath/main_loop.c
```diff
@@ -192,7 +192,7 @@ void *gr_datapath_loop(void *priv) {
 			timestamp_tmp = rte_rdtsc();
 			cycles = timestamp_tmp - timestamp;
 			max_sleep_us = atomic_load_explicit(&w->max_sleep_us, memory_order_relaxed);
-			if (ctx.last_count == 0 && max_sleep_us > 0) {
+			if (unlikely(ctx.last_count == 0 && max_sleep_us > 0)) {
```
david-marchand, line 195: This evaluation is now only done every 1024 iterations. Adding such unlikely annotation is useless.

rjarry: Why is it related to the number of iterations? I want to ensure there is no possible branch prediction that could lead to an unwanted usleep()

david-marchand: It is not, and so it makes no sense to do this change. If you still want to do optimisation with no proof of impact, go ahead.

PR #74, opened 2024-10-02
Struct copy

rjarry:


david-marchand:
Acked-by: David Marchand <david.marchand@redhat.com>

PR #75, opened 2024-10-02
treewide: homogenize how unused parameters are identified

rjarry:
Remove all occurrences of:

	(void)unused_param;

Since these can be confusing and do not convey any meaning about the variable being unused. It is a legacy way to silence the unused-parameter compiler warning.

The whole grout code base is compiled with --std=gnu2x which allows unnamed parameters even in function definitions. Make use of that to indicate unused parameters. When the parameter type is explicit (e.g. struct rte_graph *), simply remove the parameter name which is usually redundant, e.g.:

	void foo(const struct rte_graph *, struct rte_node *node) { }

When the type is not explicit (e.g. int), keep the parameter wrapped in a comment, e.g.:

	int foo(void * /*cookie*/, const char *buf, size_t size) { }

### DIFF
--- a/cli/log.c
+++ b/cli/log.c
@@ -10,9 +10,9 @@
 #include <errno.h>
 #include <unistd.h>
 
-bool stdin_isatty;
-bool stdout_isatty;
-bool stderr_isatty;
+static bool stdin_isatty;
+static bool stdout_isatty;
+static bool stderr_isatty;
 
 void tty_init(void) {
 	stdin_isatty = isatty(0);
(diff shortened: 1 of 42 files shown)


rjarry:
@apconole, I find this much less verbose than __rte_unused or even __attribute__((unused)). What do you think?

cli/log.c
```diff
@@ -10,9 +10,9 @@
 #include <errno.h>
 #include <unistd.h>
 
-bool stdin_isatty;
-bool stdout_isatty;
-bool stderr_isatty;
+static bool stdin_isatty;
```
david-marchand, line 13: This seems unrelated.

rjarry: Completely unrelated. I have split into another commit.

david-marchand:
The series lgtm.

PR #77, opened 2024-10-06
calloc: fix order parameter

christophefontaine:
nit picking: First arg of calloc argument should specify number of elements, later size of each element.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -19,17 +19,25 @@ project(
   ],
 )
 
-compiler = meson.get_compiler('c')
-add_project_arguments('-D_GNU_SOURCE', language: 'c')
+# additional project C flags (keep alpha sorted)
 add_project_arguments('-DALLOW_EXPERIMENTAL_API', language: 'c')
-add_project_arguments('-Wmissing-prototypes', language: 'c')
-if compiler.has_argument('-Wmissing-variable-declarations')
-  add_project_arguments('-Wmissing-variable-declarations', language: 'c')
-endif
-add_project_arguments('-fstrict-aliasing', language: 'c')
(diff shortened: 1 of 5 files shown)


david-marchand:
I see 4 instances of this issue in origina/main.

$ git grep 'calloc(len'
modules/ip/control/address.c:   if ((resp = calloc(len, 1)) == NULL)
modules/ip/control/nexthop.c:   if ((resp = calloc(len, 1)) == NULL) {
modules/ip6/control/address.c:  if ((resp = calloc(len, 1)) == NULL)
modules/ip6/control/nexthop.c:  if ((resp = calloc(len, 1)) == NULL) {

rjarry (approved):

david-marchand:
Just a note that in recent gcc version, this issue could be detected with -Wcalloc-transposed-args.

david-marchand:
Changes lgtm.

meson.build
```diff
@@ -19,17 +19,25 @@ project(
...
-add_project_arguments('-fstrict-aliasing', language: 'c')
-add_project_arguments('-Wstrict-aliasing=2', language: 'c')
 add_project_arguments('-Wno-format-truncation', language: 'c')
-add_project_arguments('-DGROUT_VERSION="' + meson.project_version() + '"', language: 'c')
+add_project_arguments('-Wstrict-aliasing=2', language: 'c')
+add_project_arguments('-fstrict-aliasing', language: 'c')
+
+# optional project C flags (keep alpha sorted)
+optional_c_args = [
+  '-Wmissing-variable-declarations',
```
christophefontaine, line 34: Do you want to add `-Wcalloc-transposed-args` as well ?

rjarry: done :+1:

PR #79, opened 2024-10-06
infra: fix failure when numa is unsupported

julienmeunier:
During the EAL init, even if NUMA is unsupported on a given platform, DPDK will always return a socket id set to 0 instead of SOCKET_ID_ANY (-1). It breaks the current algorithm in grout, as no valid default worker for this given socket id can be found with libnuma.

On top of that, a pre-check should be done before calling any libnuma functions. According to man numa::

  Before any other calls in this library can be used numa_available()
  must be called. If it returns -1, all other functions in this
  library are undefined.

Let's fix that by checking if the current platform really supports NUMA.

### DIFF
--- a/modules/infra/control/port.c
+++ b/modules/infra/control/port.c
@@ -76,11 +76,14 @@ static struct rte_eth_conf default_port_config = {
 };
 
 static void port_queue_assign(struct iface_info_port *p) {
-	int socket_id = rte_eth_dev_socket_id(p->port_id);
+	int socket_id = SOCKET_ID_ANY;
 	struct worker *worker, *default_worker = NULL;
 	// XXX: can we assume there will never be more than 64 rxqs per port?
 	uint64_t rxq_ids = 0;
 	uint16_t txq = 0;
+
+	if (numa_available() != -1)
+		socket_id = rte_eth_dev_socket_id(p->port_id);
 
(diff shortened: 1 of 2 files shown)


david-marchand:
Devil's advocate, do we really want to support systems without libnuma?

rjarry:
For sure we'll have other issues down the line. But it is a good thing to avoid crashing from the start when running small tests.

julienmeunier:
> Devil's advocate, do we really want to support systems without libnuma?

Well, it's not about the libnuma supported or not, but more if the kernel has been compiled with ``CONFIG_NUMA=1``.

Main distributions should have that enabled so it's not a problem for Ubuntu / RHEL / Debian & co. I just did a test on WSL 2 which hasn't this ``CONFIG_NUMA`` enabled. Some specifics VNF or home made distribution (like Yocto) might not have enabled that.

Maybe this kind of "exotic" kernel is not the main target of grout. In that case, if NUMA is mandatory (meaning if ``numa_available() == -1``) grout should stop before the EAL init and explicitly request to have a kernel with a full NUMA support.

rjarry (approved):

rjarry:
Applied in https://github.com/DPDK/grout/commit/04461f06802bc558f078bf9c316a70e1091c3026

Issue #78, opened 2024-10-06
Assert raised when adding a vdev on a platform without any NUMA support


julienmeunier:
When creating a vdev interface (net_null) on a platform without any NUMA support (like on WSL2 / Ubuntu 24.04), an assert is raised.

Server side
```
$ ./build/grout -t -v -s grout.sock &
NOTICE: GROUT: main: starting grout version v0.2-6-g2d90e7cc
INFO: GROUT: dpdk_init: DPDK 24.07.0
INFO: GROUT: dpdk_init: EAL arguments: -l 0 -a 0000:00:00.0 --no-shconf --no-huge -m 2048
INFO: GROUT: dpdk_init: running control plane on CPUs 0-15
INFO: GROUT: listen_api_socket: listening on API socket grout.sock
```

Client 
```
$ ./build/grcli -s grout.sock
Welcome to the graph router CLI version v0.2-6-g2d90e7cc.
Use ? for help and <tab> for command completion.
grout# add interface port p0 devargs net_null
WARN: GROUT: worker_ensure_default: no ideal CPU found on socket 0 for a new worker, falling back to CPU 0
NOTICE: GROUT: gr_datapath_loop: [CPU 0] starting tid=46190
INFO: GROUT: gr_datapath_loop: [CPU 0] lcore_id = 1
INFO: GROUT: worker_create: worker 0 started
grout: ../modules/infra/control/port.c:119: port_queue_assign: Assertion `default_worker != NULL' failed.
error: command failed: Bad message
grout#
```

With some debug traces added, it seems that we have a wrong socket_id returned by either DPDK or libnuma

```diff
diff --git a/modules/infra/control/port.c b/modules/infra/control/port.c
index fd6ef11..d359ec5 100644
--- a/modules/infra/control/port.c
+++ b/modules/infra/control/port.c
@@ -112,6 +112,7 @@ static void port_queue_assign(struct iface_info_port *p) {
                                }
                        }
                }
+               LOG(INFO, "%d - socket_id=%d / worker_socket_id=%d", __LINE__, socket_id, numa_node_of_cpu(worker->cpu_id));
                if (socket_id == SOCKET_ID_ANY || socket_id == numa_node_of_cpu(worker->cpu_id)) {
                        default_worker = worker;
                }
diff --git a/modules/infra/control/worker.c b/modules/infra/control/worker.c
index 3fccc05..6849332 100644
--- a/modules/infra/control/worker.c
+++ b/modules/infra/control/worker.c
[...]

PR #80, opened 2024-10-06
devtools: fix check-patches with dash

julienmeunier:
If dash is used as the default shell interpreter at /bin/sh, which is more POSIX-compliant than bash, the hook related to the issue check might fail if one field of the JSON output (like "body") contains multiples lines.

Example with the issue 78::

  jq: parse error: Invalid string: control characters from U+0000
      through U+001F must be escaped at line 130, column 4
  error [PATCH 1/1]  [...] 'https://github.com/DPDK/grout/issues/78'
        does not reference a valid open issue

From POSIX spec for echo::

  It is not possible to use echo portably across all POSIX systems
  unless both -n (as the first argument) and escape sequences are
  omitted.
  [...]
  New applications are encouraged to use printf instead of echo.

Use ``printf`` instead of ``echo`` to be more portable.

### DIFF
--- a/devtools/check-patches
+++ b/devtools/check-patches
@@ -47,7 +47,7 @@ check_issue() {
 		-H "Accept: application/vnd.github+json" \
 		-H "X-GitHub-Api-Version: 2022-11-28" \
 		"$api_url/issues/${1##*/}") || return 1
-	test "$(echo "$json" | jq -r .state)" = open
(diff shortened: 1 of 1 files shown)


rjarry:
Applied in https://github.com/DPDK/grout/commit/158fc82756409c110840184c1e377426ce1691bc

PR #83, opened 2024-10-17
iface: avoid use after free on duplicate name

rjarry:
When an interface name is already taken, we return an error but the iface pointer is set the already existing interface. Before returning from iface_create, that pointer is freed without removing the interface from the system.

When listing interface or accessing anything after that causes an invalid memory access.

Avoid going through the free code path and return immediately when either the interface type or name are invalid or if the name is duplicate.

Fixes: a0f9d6699605 ("iface: forbid duplicate names on creation")
Reported-by: Christophe Fontaine <cfontain@redhat.com>

### DIFF
--- a/modules/infra/control/iface.c
+++ b/modules/infra/control/iface.c
@@ -79,6 +79,7 @@ struct iface *iface_create(
 		goto fail;
 	while ((iface = iface_next(GR_IFACE_TYPE_UNDEF, iface)) != NULL) {
 		if (strcmp(name, iface->name) == 0) {
+			iface = NULL;
 			errno = EEXIST;
 			goto fail;
 		}


modules/infra/control/iface.c
```diff
@@ -71,14 +71,15 @@ struct iface *iface_create(
...
 
 	if (type == NULL)
 		goto fail;
 	if (utf8_check(name, GR_IFACE_NAME_SIZE) < 0)
 		goto fail;
-	while ((iface = iface_next(GR_IFACE_TYPE_UNDEF, iface)) != NULL) {
-		if (strcmp(name, iface->name) == 0) {
+	while ((i = iface_next(GR_IFACE_TYPE_UNDEF, i)) != NULL) {
+		if (strcmp(name, i->name) == 0) {
 			errno = EEXIST;
```
christophefontaine, line 83: maybe just set "iface = NULL;" here, instead of defining a new var ?

rjarry: Good point :)

Done.

PR #84, opened 2024-10-20
dpdk: update to v24.11-rc1

rjarry:
Update the DPDK version grout relies on. All our local patches have been applied upstream; remove them.


david-marchand:
Acked-by: David Marchand <david.marchand@redhat.com>

PR #85, opened 2024-10-22
Cleanup

rjarry:


david-marchand (commented):
Why call this header with _datapath?
This header is local to datapath/ already.

rjarry:
Good point. I wanted to avoid any confusion if we had "gr_eth" someday. I'll remove that _datapath suffix.

Issue #89, opened 2024-10-30
Duplicated entries in the next hop table


christophefontaine:
When we add a route before the arp resolution is done, we end up with a duplicated entry in the next hop table, and grout fails to forward packets.

```
grout# show ip nexthop
VRF  IP       MAC                IFACE    QUEUE  AGE  STATE
0    1.1.1.1  3c:fd:fe:a0:a4:c4  dpdk-p0  0      0    reachable static local link
0    2.2.2.1  3c:fd:fe:a0:a4:c5  dpdk-p1  0      0    reachable static local link
0    1.1.1.2  ??:??:??:??:??:??  ?        0      ?    gateway
0    2.2.2.2  ??:??:??:??:??:??  ?        0      ?    gateway
0    1.1.1.2  b4:83:51:05:43:ca  dpdk-p1  0      20   reachable
0    2.2.2.2  b4:83:51:05:43:cb  dpdk-p1  0      20   reachable
```

Reproducer:
1. add an interface
2. configure IP
3. add route
4. process arp request message for the gateway (for example, from trex service mode, run "arp command")

If run the step 4 before step 3, the nh entry is properly populated, and the ARP node updates the nh properly.

```
grout# show ip nexthop
VRF  IP       MAC                IFACE    QUEUE  AGE  STATE
0    1.1.1.1  3c:fd:fe:a0:a4:c4  dpdk-p0  0      0    reachable static local link
0    2.2.2.1  3c:fd:fe:a0:a4:c5  dpdk-p1  0      0    reachable static local link
0    1.1.1.2  b4:83:51:05:43:ca  dpdk-p1  0      26   reachable gateway
0    2.2.2.2  b4:83:51:05:43:cb  dpdk-p1  0      26   reachable gateway
```

PR #90, opened 2024-10-30
arp: fix nh lookup

christophefontaine:
During the processing of an arp message, update the existing nexthop instead of creating a duplicated entry in the nh table.

Closes: https://github.com/DPDK/grout/issues/89

### DIFF
--- a/modules/ip/datapath/arp_input.c
+++ b/modules/ip/datapath/arp_input.c
@@ -109,7 +109,7 @@ arp_input_process(struct rte_graph *graph, struct rte_node *node, void **objs, u
 		sip = arp->arp_data.arp_sip;
 		iface = eth_input_mbuf_data(mbuf)->iface;
 		local = ip4_addr_get_preferred(iface->id, sip);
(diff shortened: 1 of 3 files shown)


rjarry:
Hey Chris, can you explain your reasoning here? I don't understand why it is necessary to update both?

christophefontaine:
I update both entries as in the next hop table, we may have 2 entries for a single IP:
```
grout# show ip nexthop
VRF  IP       MAC                IFACE    QUEUE  AGE  STATE
0    1.1.1.2  b4:83:51:05:43:ca  dpdk-p0  0      420  reachable gateway
0    1.1.1.2  b4:83:51:05:43:ca  dpdk-p0  0      420  reachable
```

This is because we may have have an additional 1.1.1.2/32 in the route table:
```
grout# show ip route
DESTINATION  NEXT_HOP
1.1.1.2/32   1.1.1.2
1.1.1.0/24   1.1.1.1
16.0.0.0/8   1.1.1.2
```

rjarry (approved):

PR #92, opened 2024-11-14
meson: scdoc and cmocka are optionals

vjardin:
Avoid the dependencies when manual and tests are disabled.


david-marchand (commented):
Indeed, it is better to leave it to the packager to choose those options.
LGTM.

david-marchand:
I had a quick look on the other two patches, lgtm too.

vjardin:
thanks @david-marchand and @rjarry

PR #96, opened 2024-11-28
arp: fix source ip in arp reply

christophefontaine:
Given the following configuration:
iface0: 192.168.0.1/32
iface1: 192.168.0.2/24

When an arp_request arrives on iface1 (who has 192.168.0.1?), grout answers with the ip belonging to the interface iface1, so the tip field in the arp message is filled with "192.168.0.2" instead of the requested IP.

Swap the fields SIP and TIP to avoid this: this is safe as the only possible upstream node is "arp_input".


david-marchand:
Well, either grout should ignore this arp request, or reply aboutthe ip it was asked about :-).
I think the latter approach is correct.

PR #100, opened 2024-12-01
loopback & management plane support

christophefontaine:
A loopback interface is often required as it will hold internal VIPs.
We back this interface with a tun interface to create a corresponding netdev.
This netdev can be used to interact with the external world "in-band".

To avoid sys calls in the datapath to send/receive packets from this interface, delegate read and write sys calls to the control plane, relying on libevent.

Exactly 1 loopback interface is created automatically per VRF, so there is no need to implement an API to create/delete these interfaces.
The creation relies on iface events create/delete/reconfig.


christophefontaine (commented):
naming "redirect" won't be good enough to be explicit about it's action, will find a better name to that.

modules/ip/datapath/l4_input_local.c
```diff
@@ -0,0 +1,108 @@
...
+	.nb_edges = EDGE_COUNT,
+	.next_nodes = {
+		[REDIRECT] = "loop_output",
+		[NO_IFACE] = "no_loop_iface",
+	},
+};
+
+static void udp_input_local_register(void) {
+	ip_input_local_add_proto(IPPROTO_UDP, "udp_input_local");
+	ip6_input_local_add_proto(IPPROTO_UDP, "udp_input_local");
```
christophefontaine, line 77: @rjarry I'm not sure if it is clean to call ip6_input_local_add_proto here ?

rjarry: Good question. We have multiple options here. Either you duplicate the nodes in the ip6 module (udp6_input_local, etc.) or we create a new l4 module that holds udp_input_local which will handle both ipv4 and ipv6.

I don't have a strong preference. Probably duplicating the nodes is OK and will be less work. But maybe we will have things to do in the new l4 module.

christophefontaine: Ack, I'll create a new l4 module, as it doesn't make sense to me to duplicate this code.

modules/infra/api/gr_infra.h
```diff
@@ -18,6 +18,7 @@
 #define GR_IFACE_TYPE_UNDEF 0x0000
 #define GR_IFACE_TYPE_PORT 0x0001
 #define GR_IFACE_TYPE_VLAN 0x0002
+#define GR_IFACE_TYPE_LOOPBACK 0x0004
```
rjarry, line 21: Maybe we could use a value that does not risk being accidentally taken by another logical interface type? How about `0x1337` or something similar?

modules/infra/datapath/loop_input.c
```diff
@@ -0,0 +1,68 @@
...
+			edge = IP6_INPUT;
+
+		rte_node_enqueue_x1(graph, node, edge, ip);
+		rte_pktmbuf_free(mbuf);
+	}
+	return nb_objs;
+}
+
+static struct rte_node_register loop_input_node = {
+	.name = "loop_input",
```
rjarry, line 49: I would prefer `"loopback_input"` to match the interface type which is called `"loopback"`.

modules/infra/datapath/loop_input.c
```diff
@@ -0,0 +1,68 @@
...
+	.process = loop_input_process,
+	.nb_edges = EDGE_COUNT,
+	.next_nodes = {
+		//FIXME: use something similar to icmp_local_send instead of forward
```
rjarry, line 53: Is this addressed in another commit?

modules/infra/datapath/loop_output.c
```diff
@@ -0,0 +1,53 @@
...
+	.process = loop_output_process,
+	.nb_edges = EDGE_COUNT,
+	.next_nodes = {
+		[CONTROL_OUTPUT] = "control_output",
+	},
+};
+
+static void loop_output_register(void) {
+	ip_output_add_tunnel(GR_IFACE_TYPE_LOOPBACK, "loop_output");
+	ip6_output_add_tunnel(GR_IFACE_TYPE_LOOPBACK, "loop_output");
```
rjarry, line 45: I thought you would be using control_output for this. Maybe this is done in another commit. But I don't think there should be any IP level output done on loopback interfaces.

Same remark about the node name -> `s/loop_output/loopback_output/`

modules/l4/l4_input_local.c
```diff
@@ -0,0 +1,108 @@
...
+
+static void tcp_input_local_register(void) {
+	ip_input_local_add_proto(IPPROTO_TCP, "tcp_input_local");
+	ip6_input_local_add_proto(IPPROTO_TCP, "tcp_input_local");
+}
+
+static void sctp_input_local_register(void) {
+	ip_input_local_add_proto(IPPROTO_SCTP, "sctp_input_local");
+	ip6_input_local_add_proto(IPPROTO_SCTP, "sctp_input_local");
+}
```
rjarry, line 88: I know it does not cost much, but do we have a use case for sctp?

modules/infra/api/iface.c
```diff
@@ -53,8 +53,17 @@ static struct api_out iface_add(const void *request, void **response) {
...
+	struct iface *iface;
 	int ret;
 
+	// Loopback interfaces are specials, and are deleted
+	// when the last interface of a VRF is destroyed.
+	if ((iface = iface_from_id(req->iface_id)) == NULL)
+		return api_out(ENODEV, 0);
+
+	if (iface->type_id == GR_IFACE_TYPE_LOOPBACK)
+		return api_out(EPERM, 0);
```
rjarry, line 65: EINVAL would be more suitable.

modules/ip/datapath/gr_ip4_datapath.h
```diff
@@ -16,9 +16,7 @@
 
 #include <stdint.h>
 
-GR_MBUF_PRIV_DATA_TYPE(ip_output_mbuf_data, {
-	struct nexthop *nh;
-});
+GR_MBUF_PRIV_DATA_TYPE(ip_output_mbuf_data, { struct nexthop *nh; });
```
rjarry: Should this go in the previous commit?

modules/infra/cli/loopback.c
```diff
@@ -0,0 +1,25 @@
...
+#include <gr_api.h>
+#include <gr_cli.h>
+#include <gr_cli_iface.h>
+#include <gr_infra.h>
+
+static void loop_show(const struct gr_api_client *, const struct gr_iface *) { }
+
+static void
+loop_list_info(const struct gr_api_client *, const struct gr_iface *, char *buf, size_t len) {
+	snprintf(buf, len, "%s", "");
```
rjarry, line 13: Probably not needed.

rjarry (changes requested):

modules/infra/datapath/gr_mbuf.h
```diff
@@ -23,10 +23,13 @@ struct gr_trace_item {
 
 STAILQ_HEAD(gr_trace_head, gr_trace_item);
 
-#define GR_MBUF_PRIV_MAX_SIZE RTE_CACHE_LINE_MIN_SIZE
+#define GR_MBUF_PRIV_MAX_SIZE RTE_CACHE_LINE_MIN_SIZE * 2
```
christophefontaine, line 26: @rjarry 
As of now, the only common data is `struct iface *`, but this is enough to overflow with ipv6 related data structures.
The other option would be reuse a dynfield and create a specific accessor for it. Even if I'm not fan of that option, as it is specific, it avoids the increase of the PRIV_SIZE.
WDYT ?

modules/ip/datapath/gr_ip4_datapath.h
christophefontaine: It seems so: I may not have run `make format` on the previous commit !

modules/infra/datapath/gr_mbuf.h
rjarry: Honestly, I don't think it is a big deal to have 128 bytes of priv data. Unless we have proof that it has negative effects on performance, I would say we keep it like you changed it.

modules/infra/api/gr_infra.h
christophefontaine: Any arbitrary number may be used here.
GR_IFACE_TYPE_IPIP uses 0x003, that's why I took 0x004.

modules/infra/datapath/loop_input.c
christophefontaine: ack.

modules/infra/datapath/loop_input.c
christophefontaine: Not yet.

modules/infra/datapath/loop_output.c
christophefontaine: `loop_output` is using `control_ouput`. Yet, you still have to format the mbuf to prepare for control output though.

Loopback interfaces are backed by a tun, which is at the L3. So, the packets must not go into eth_output.
TUN faces are similar to IPIP ifaces in that sense.

modules/l4/l4_input_local.c
christophefontaine: No, but as you said, it doesn't cost anything.

modules/infra/api/iface.c
christophefontaine: ack.

modules/infra/datapath/loop_output.c
rjarry: Maybe the naming isn't right then. I was thinking of the following:

```mermaid
graph LR;
eth_input-->ip_input;
ip_input-->ip_forward;
ip_forward-->ip_output;
ip_output-->eth_output;
ip_output-->ipip_output;
ipip_output-->ip_output;
ip_input-->ip_input_local;
ip_input_local-->ipip_input;
ipip_input-->ip_input;
ip_input_local-->icmp_input;
ip_input_local-->tcp_input_local;
ip_input_local-->udp_input_local;
tcp_input_local-->loopback_input;
udp_input_local-->loopback_input;
loopback_input-->control_output;
control_input-->loopback_output;
loopback_output-->tcp_output_local;
loopback_output-->udp_output_local;
tcp_output_local-->ip_output;
udp_output_local-->ip_output;
```

christophefontaine: you inverted loopback_input and loopback_output. That's why you're confused !

modules/infra/cli/loopback.c
christophefontaine: buf is already filled with the parent function, so it is required to clear it.

rjarry: Oh I get it, `buf` should be reset before calling `type->list_info()` here:

https://github.com/DPDK/grout/blob/bd52b2c275e23812e18d39601e179256551ab19f/modules/infra/cli/iface.c#L298

modules/infra/datapath/gr_mbuf.h
christophefontaine: ack, thanks !

modules/infra/datapath/loop_input.c
christophefontaine: I added the ip(6)_lookup call directly in loop_input.

christophefontaine:
- Most comments have been addressed.
- rebased on main
- svg update
- Rewrite of loopback_input done, putting back as draft until a full IP and IP6 test is done.

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,196 @@
...
+	int fd;
+	struct event *ev;
+};
+
+void loopback_tx(struct rte_mbuf *m) {
+	struct mbuf_data *d = mbuf_data(m);
+	struct iface_info_loop *loop;
+
+	loop = (struct iface_info_loop *)d->iface->info;
+	if (write(loop->fd, rte_pktmbuf_mtod(m, char *), m->data_len) < 0)
```
rjarry, line 40: @david-marchand is this correct to use `m->data_len` here? Think it would be better to use `rte_pktmbuf_pkt_len(m)`.

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,196 @@
...
+	struct iface *iface = ev_iface;
+	struct rte_mbuf *mbuf;
+	size_t len;
+	char *data;
+
+	loop = (struct iface_info_loop *)iface->info;
+
+	mbuf = rte_pktmbuf_alloc(loopback_pool);
+	if (!mbuf)
+		goto end;
```
rjarry, line 58: Maybe log an error here?

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,196 @@
...
+
+	mbuf->data_off = 256;
```
rjarry, line 60: I think you should use the pktmbuf API for this. Probably @david-marchand can help.

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,196 @@
...
+	data = rte_pktmbuf_mtod(mbuf, char *);
+	len = read(loop->fd, data, 1600);
```
rjarry, line 62: Could we have a different MTU configured on the loopback from linux? Is there any way to detect this?

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,196 @@
...
+
+	if (mbuf->buf_len <= 0) {
+		rte_pktmbuf_free(mbuf);
+		goto end;
```
rjarry, line 66: Maybe you meant to check `if (len <= 0)`. Also, add an error here.

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,196 @@
...
+	}
+	mbuf->data_len = len;
+	mbuf->pkt_len = mbuf->data_len;
+
+	if ((*data & 0xf0) == 0x40) {
+		mbuf->packet_type = RTE_PTYPE_L3_IPV4;
+	} else if ((*data & 0xf0) == 0x60) {
+		mbuf->packet_type = RTE_PTYPE_L3_IPV6;
+		rte_pktmbuf_free(mbuf);
```
rjarry, line 75: Is this normal to free the packet here?

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,196 @@
...
+
+	iface->flags = GR_IFACE_F_UP;
+	iface->state = GR_IFACE_S_RUNNING;
+	loop->ev = event_new(global_ev_base, loop->fd, EV_READ, iface_loopback_poll, iface);
+	event_add(loop->ev, NULL);
+	close(ioctl_sock);
+	return 0;
+
+cleanup:
+	printf("error: %s\n", strerror(errno));
```
rjarry, line 150: Can you remove this printf? If there is an error to log, it should be done with the LOG macro.

modules/infra/datapath/loop_input.c
```diff
@@ -0,0 +1,123 @@
...
+
+		if (gr_mbuf_is_traced(objs[i])
+		    || mbuf_data(mbuf)->iface->flags & GR_IFACE_F_PACKET_TRACE) {
+			// control input may already have added a trace.
+			// Copy it and clear it so we do not display it twice.
+			struct __mbuf_data *priv_old = rte_mbuf_to_priv(objs[i]);
+			struct __mbuf_data *priv = rte_mbuf_to_priv(mbuf);
+			priv->traces = priv_old->traces;
+			STAILQ_INIT(&priv_old->traces);
+			gr_mbuf_trace_add(mbuf, node, 0);
```
rjarry, line 50: Huh, this seems hacky. I don't understand why you need to reset it.

modules/infra/datapath/loop_input.c
```diff
@@ -0,0 +1,123 @@
...
+		}
+		rte_pktmbuf_free(objs[i]);
```
rjarry, line 52: Why do you need to free the packet before sending it?

rjarry (changes requested):

modules/infra/control/loopback.c
david-marchand: Well, it is not clear what is expected here.
rte_pktmbuf_pkt_len() would be incorrect, since write() expects a single linear buffer.

If m is a multi segment packet, then we have a problem.

modules/infra/control/loopback.c
david-marchand: People directly accessing directly mbuf fields should be banned from using DPDK :-p.

Packets have a default headroom of 128 bytes.

If the intention is to reserve some extra headroom in prevision of encapsulations, it is better to call rte_pktmbuf_append(), then rte_pktmbuf_adj().

modules/infra/control/loopback.c
christophefontaine: yes, will fix it.

modules/infra/control/loopback.c
christophefontaine: fixed.

modules/infra/datapath/loop_input.c
```diff
@@ -0,0 +1,117 @@
...
+static uint16_t loopback_input_process(
+	struct rte_graph *graph,
+	struct rte_node *node,
+	void **objs,
+	uint16_t nb_objs
+) {
+	struct rte_mbuf *mbuf;
+	rte_edge_t edge;
+
+	for (unsigned i = 0; i < nb_objs; i++) {
```
vjardin, line 39: if ng_objs is uint16_t, i should be uint16_t but not unsigned. It just helps to support the consistency.

modules/infra/datapath/loop_output.c
```diff
@@ -0,0 +1,57 @@
...
+static uint16_t loopback_output_process(
+	struct rte_graph *graph,
+	struct rte_node *node,
+	void **objs,
+	uint16_t nb_objs
+) {
+	struct control_output_mbuf_data *co;
+	struct rte_mbuf *mbuf;
+
+	for (unsigned i = 0; i < nb_objs; i++) {
```
vjardin, line 29: same, avoid unsigned.

modules/l4/l4_redirect_loopback.c
```diff
@@ -0,0 +1,113 @@
...
+	struct rte_node *node,
+	void **objs,
+	uint16_t nb_objs
+) {
+	struct rte_ipv4_hdr *ip;
+	struct rte_mbuf *mbuf;
+	struct mbuf_data *d;
+	int edge = NO_IFACE;
+
+	for (uint16_t i = 0; i < nb_objs; i++) {
```
vjardin, line 33: here, you did it

vjardin (changes requested):

modules/infra/control/loopback.c
christophefontaine: Thanks ! Will add that to the next version.

modules/infra/control/loopback.c
christophefontaine: fixed.

modules/infra/datapath/loop_input.c
christophefontaine: fixed with the updated control to stack mechanism.

modules/infra/datapath/loop_input.c
christophefontaine: good catch. Will send a new revision today.

modules/infra/control/loopback.c
christophefontaine: yes, done.

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,212 @@
...
+};
+
+void loopback_tx(struct rte_mbuf *m) {
+	struct mbuf_data *d = mbuf_data(m);
+	struct iface_info_loop *lo;
+	void *data;
+
+	lo = (struct iface_info_loop *)d->iface->info;
+	if (rte_pktmbuf_linearize(m) == 0) {
+		data = rte_pktmbuf_mtod(m, char *);
```
david-marchand, line 44: Nit: data is void *, so call mtod accordingly.

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,212 @@
...
+cleanup:
+	printf("error: %s\n", strerror(errno));
+	if (lo->fd > 0)
+		close(lo->fd);
+	if (ioctl_sock > 0)
+		close(ioctl_sock);
+	return -1;
+}
+
+static int iface_loop_fini(struct iface *iface) {
```
rjarry, line 175: s/iface_loop_fini/iface_loopback_fini/

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,212 @@
...
+	struct iface_info_loop *lo = (struct iface_info_loop *)iface->info;
+	event_del(lo->ev);
+	event_free(lo->ev);
+	close(lo->fd);
+	return 0;
+}
+
+static void loop_init(struct event_base *ev_base) {
```
rjarry, line 183: s/loop_init/loopback_init/

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,212 @@
...
+	loopback_pool = gr_pktmbuf_pool_get(SOCKET_ID_ANY, RTE_GRAPH_BURST_SIZE);
+	global_ev_base = ev_base;
+}
+
+static void loop_fini(struct event_base *) {
```
rjarry, line 188: s/loop_fini/loopback_fini/

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,212 @@
...
+	gr_pktmbuf_pool_release(loopback_pool, RTE_GRAPH_BURST_SIZE);
+}
+
+static void iface_loop_to_api(void * /* info */, const struct iface * /* iface */) { }
```
rjarry, line 192: s/iface_loop_to_api/iface_loopback_to_api/

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,212 @@
...
+		goto cleanup;
+
+	if (iface->mtu == 0)
+		iface->mtu = 1500;
+
+	if ((data = rte_pktmbuf_append(mbuf, iface->mtu)) == NULL)
+		goto cleanup;
+
+	if ((len = read(lo->fd, data, iface->mtu)) <= 0)
+		goto cleanup;
```
rjarry, line 84: Check in `main/api.c` for other potential errors returned by read() in a libevent callback.

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,212 @@
...
+	// Emulate ethernet input, required by ip(6)_input
+	e = eth_input_mbuf_data(mbuf);
+	e->iface = iface;
+	e->eth_dst = ETH_DST_LOCAL;
+
+	post_to_stack(loopback_get_control_id(), mbuf);
+	return;
+
+cleanup:
+	if (mbuf)
```
david-marchand, line 105: rte_pktmbuf_free already checks if mbuf is NULL.

modules/infra/control/loopback.c
rjarry: It is still there :)

modules/infra/datapath/control_input.c
```diff
@@ -70,13 +77,22 @@ static uint16_t control_input_process(
...
-			rte_node_enqueue_x1(graph, node, control_input_edges[msg[i].type], mbuf);
+		if (control_input_edges[msg[i].type].data_is_mbuf) {
+			mbuf = msg[i].data;
+		} else {
+			mbuf = rte_pktmbuf_alloc(mp);
+			if (mbuf)
+				control_input_mbuf_data(mbuf)->data = msg[i].data;
+			else {
+				LOG(ERR, "rte_pktmbuf_alloc %s", rte_strerror(rte_errno));
+				continue;
```
david-marchand, line 88: More readable to invert this test.

if (!mbuf) {
    LOG(ERR, "rte_pktmbuf_alloc %s", rte_strerror(rte_errno));
    continue;
}

control_input_mbuf_data(mbuf)->data = msg[i].data;

modules/infra/control/loopback.c
christophefontaine: Modified data to char* as need to access to fill in struct tun_pi.

modules/infra/datapath/control_input.c
christophefontaine: done

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,212 @@
...
+	} else {
+		data = rte_malloc(NULL, rte_pktmbuf_pkt_len(m), 0);
+		if (data == NULL) {
+			LOG(ERR, "rte_malloc failed %s", rte_strerror(rte_errno));
+			return;
+		}
+		data = (void *)rte_pktmbuf_read(m, 0, rte_pktmbuf_pkt_len(m), data);
+	}
+
+	if (write(lo->fd, data, rte_pktmbuf_pkt_len(m)) < 0)
```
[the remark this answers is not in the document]
christophefontaine: As the tun is set 'non-blocking", this won't block, will it ?

modules/infra/datapath/loop_input.c
```diff
@@ -0,0 +1,127 @@
...
+	.process = loopback_input_process,
+	.nb_edges = EDGE_COUNT,
+	.next_nodes = {
+		[IP_LOCAL] = "ip_input_local",
+		[IP_OUTPUT] = "ip_output",
+		[IP6_LOCAL] = "ip6_input_local",
+		[IP6_OUTPUT] = "ip6_output",
+		[IP_NO_ROUTE] = "ip_error_dest_unreach",
+		[IP6_NO_ROUTE] = "ip6_error_dest_unreach",
+		[BAD_PROTO] = "tun_error_bad_proto",
```
rjarry, line 113: s/tun/loopback/

rjarry (changes requested):
This is almost perfect! I have only a few minor remarks.

modules/infra/control/loopback.c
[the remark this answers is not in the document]
rjarry: Yes, I meant it could return `EAGAIN`. But you addressed my remark in the patch with an explicit comment :)

All good.

modules/infra/control/loopback.c
christophefontaine: lib event wouldn't call this callback if there is no data to read, would it ?
I'm reworking a bit this part of code to use more common paradigms of libevent though.

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,232 @@
...
+			LOG(ERR, "rte_malloc failed %s", rte_strerror(rte_errno));
+			return;
+		}
+		data = (char *)rte_pktmbuf_read(m, 0, rte_pktmbuf_pkt_len(m), data);
+	}
+	pi.flags = 0;
+	if ((data[0] & 0xf0) == 0x40)
+		pi.proto = RTE_BE16(RTE_ETHER_TYPE_IPV4);
+	else
+		pi.proto = RTE_BE16(RTE_ETHER_TYPE_IPV6);
```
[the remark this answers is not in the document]
christophefontaine: ack

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,232 @@
...
+	lo = (struct iface_info_loopback *)iface->info;
+
+	mbuf = rte_pktmbuf_alloc(loopback_pool);
+	if (!mbuf) {
+		LOG(ERR, "rte_pktmbuf_alloc %s", rte_strerror(rte_errno));
+		goto cleanup;
+	}
+
+	if ((data = rte_pktmbuf_append(mbuf, iface->mtu)) == NULL)
+		goto cleanup;
```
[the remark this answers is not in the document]
christophefontaine: fixed.

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,253 @@
...
+	rte_pktmbuf_trim(mbuf, iface->mtu - len);
+
+	// packet sent from linux tun iface, no need to compute checksum;
+	mbuf->ol_flags = RTE_MBUF_F_RX_IP_CKSUM_GOOD;
+	// Emulate ethernet input, required by ip(6)_input
+	e = eth_input_mbuf_data(mbuf);
+	e->iface = iface;
+	e->eth_dst = ETH_DST_LOCAL;
+
+	post_to_stack(loopback_get_control_id(), mbuf);
```
david-marchand, line 138: I see no implementation of loopback_get_control_id() in this commit.
Compilation between patches is likely broken.

rjarry (changes requested):
Just a few minor changes and I think it is good to go :+1:

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,275 @@
...
+	lo = (struct iface_info_loopback *)d->iface->info;
+	if (rte_pktmbuf_linearize(m) == 0) {
+		data = rte_pktmbuf_mtod(m, char *);
+	} else {
+		data = rte_malloc(NULL, rte_pktmbuf_pkt_len(m), 0);
+		if (data == NULL) {
+			LOG(ERR, "rte_malloc failed %s", rte_strerror(rte_errno));
+			goto end;
+		}
+		data = (char *)rte_pktmbuf_read(m, 0, rte_pktmbuf_pkt_len(m), data);
```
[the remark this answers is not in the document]
christophefontaine: fixed, added a comment as well.

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,275 @@
...
+	}
+	pi.flags = 0;
+	if ((data[0] & 0xf0) == 0x40)
+		pi.proto = RTE_BE16(RTE_ETHER_TYPE_IPV4);
+	else if ((data[0] & 0xf0) == 0x60)
+		pi.proto = RTE_BE16(RTE_ETHER_TYPE_IPV6);
+	else {
+		LOG(ERR, "Bad proto: 0x%x - drop packet", pi.proto);
```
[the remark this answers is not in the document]
christophefontaine: indeed. fixed.

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,275 @@
...
+
+int iface_loopback_delete(uint16_t vrf_id) {
+	const struct iface *i = NULL;
+	while ((i = iface_next(GR_IFACE_TYPE_LOOPBACK, i)) != NULL)
+		if (i->vrf_id == vrf_id)
+			break;
+	if (i == NULL)
+		return -ENODEV;
+
+	return iface_destroy(i->id);
```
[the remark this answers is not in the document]
christophefontaine: done

modules/infra/control/loopback.c
```diff
@@ -0,0 +1,275 @@
...
+}
+
+static int iface_loopback_fini(struct iface *iface) {
+	struct iface_info_loopback *lo = (struct iface_info_loopback *)iface->info;
+	event_free_finalize(0, lo->ev, finalize_fd);
+	return 0;
+}
+
+static void loopback_module_init(struct event_base *base) {
+	loopback_pool = gr_pktmbuf_pool_get(SOCKET_ID_ANY, RTE_GRAPH_BURST_SIZE);
```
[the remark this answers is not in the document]
christophefontaine: crash early :)

modules/infra/api/gr_infra.h
```diff
@@ -15,9 +15,10 @@
 #include <sys/types.h>
 
 // Value for gr_iface.type
-#define GR_IFACE_TYPE_UNDEF 0x0000
-#define GR_IFACE_TYPE_PORT 0x0001
-#define GR_IFACE_TYPE_VLAN 0x0002
+#define GR_IFACE_TYPE_UNDEF 0x00
+#define GR_IFACE_TYPE_PORT 0x01
+#define GR_IFACE_TYPE_VLAN 0x02
+#define GR_IFACE_TYPE_MAX 256
```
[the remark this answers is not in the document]
christophefontaine: yes, on iface registration, I added a check instead of modifying gr_iface.type.

rjarry (approved):

PR #101, opened 2024-12-02
Graph

rjarry:


modules/infra/api/graph.c
```diff
@@ -37,15 +41,34 @@ static struct api_out graph_dump(const void * /*request*/, void **response) {
 	if ((ret = rte_graph_export(graph_name, stream)) < 0)
 		goto end;
 	fflush(stream);
-	resp_len = sizeof(*resp) + buf_len + 1;
```
david-marchand, line 40: Rather than allocate another buffer and carefully pay attention to move back data, I propose a slightly different approach and only copy what is needed in the final reply (slightly oversized though containing only the right amount of data) buffer.

```
@@ -37,15 +41,37 @@ found:
        if ((ret = rte_graph_export(graph_name, stream)) < 0)
                goto end;
        fflush(stream);
-       resp_len = sizeof(*resp) + buf_len + 1;
+
+       resp_len = sizeof(*resp) + buf_len;
        if ((resp = calloc(1, resp_len)) == NULL) {
                ret = -ENOMEM;
                resp_len = 0;
                goto end;
        }
 
-       resp->len = buf_len + 1;
-       memccpy(resp->dot, buf, 0, buf_len + 1);
+       if (!(req->flags & GR_INFRA_GRAPH_DUMP_F_ERRORS)) {
+               unsigned copied = 0;
+               char *line = buf;
+               char *eol;
+
+               while ((eol = strchr(line, '\n')) != NULL) {
+                       eol[0] = '\0';
+                       // Remove sink nodes from the output. They all have a "darkorange" color.
+                       // Also remove non-sink nodes that contain "error" in their name.
+                       if (!strstr(line, "darkorange") && !strstr(line, "error")) {
+                               eol[0] = '\n';
+                               memccpy(&resp->dot[copied], line, 0, eol - line + 1);
+                               copied += eol - line + 1;
+                       }
+
+                       line = eol + 1;
+               }
+               buf_len = copied;
+       } else {
+               memccpy(resp->dot, buf, 0, buf_len);
+       }
+
+       resp->len = buf_len;
        *response = resp;
 end:
        fclose(stream);

```

david-marchand: There is also the option of simply tracking the value returned by memccpy instead of incrementing a copied length.

modules/infra/api/graph.c
```diff
@@ -37,15 +41,34 @@ static struct api_out graph_dump(const void * /*request*/, void **response) {
...
+	}
+
+	resp_len = sizeof(*resp) + buf_len;
 	if ((resp = calloc(1, resp_len)) == NULL) {
 		ret = -ENOMEM;
 		resp_len = 0;
 		goto end;
 	}
 
-	resp->len = buf_len + 1;
```
david-marchand, line 47: This change is confusing, as it is actually unrelated to the change.
This should go to a separate commit (along with the update in cli/graph.c lines 22/23 hunk).

rjarry: done

modules/infra/api/graph.c
rjarry: done

modules/infra/api/graph.c
```diff
@@ -44,8 +48,27 @@ static struct api_out graph_dump(const void * /*request*/, void **response) {
...
+		char *line = buf;
+		char *eol;
+
+		while ((eol = strchr(line, '\n')) != NULL) {
+			*eol = '\0';
+			// Remove sink nodes from the output. They all have a "darkorange" color.
+			// Also remove non-sink nodes that contain "error" in their name.
+			if (!strstr(line, "darkorange") && !strstr(line, "error")) {
+				*eol = '\n'; // restore newline char
+				copy = memccpy(copy, line, 0, eol - line + 1);
```
david-marchand, line 62: Should be memccpy(.., '\n', ..)

rjarry: you think I would have tested again ? :)

david-marchand:
Series lgtm.

PR #103, opened 2024-12-02
treewide: add custom printf specifiers for addresses

rjarry:
Formatting networking addresses can be tedious and lead to ugly code, especially when dealing with IPv6.

Unfortunately, we cannot use custom printf specifiers since these lead to `-Wformat` warnings. Neither GCC nor clang know of any libc custom printf handlers.

Redefine the builtin `"%p"` pointer printf specifier with a custom function. Leverage the fact that printf supports width specifiers. Use well known widths to associate pointers with specific types. Add short defines to hold these specifiers for clarity:

|  define    |   specifier     |   expected argument type   |
|------------| ----------------| -------------------------   |
|  `ETH_F`     |   `"%2p"`         |   `struct rte_ether_addr *` | 
|  `IP4_F`     |   `"%4p"`         |   `ip4_addr_t *`   |
|  `IP6_F`     |   `"%6p"`         |   `struct rte_ipv6_addr *` |

Replace all printf/log statements to use these specifiers. Remove all manual address splitting and `inet_ntop()` calls.

Also add a generic address format specifier that is meant to be used with a variable width argument:

| define      | specifier       | expected arguments types   |
| ------------ |---------------- |-------------------------  |
| `ADDR_F` |      `"%*p"`    |        `int, void *`    |

Which can be used as follows:

```c
void print_addr(int family, void *addr) {
        printf(ADDR_F "\n", family == AF_INET ? 4 : 6, addr);
}
```

If any of the pointer arguments are NULL, they will be formatted as `"(nil)"` like with the default `"%p"` specifier, regardless of the width value.

The ability to define custom handlers for printf specifiers is only supported in GLIBC. We do not target any other platform at the moment.

Add unit tests to ensure it works as advertised.

### DIFF
--- a/api/printf.c
+++ b/api/printf.c
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2024 Robin Jarry
+
+#include <gr_net_types.h>
+
+#include <arpa/inet.h>
+#include <printf.h>
+#include <stdlib.h>
+#include <sys/socket.h>
+
+static int format_pointer(FILE *f, const struct printf_info *info, const void *const *args) {
+	char buf[INET6_ADDRSTRLEN];
+	const void *arg = *(const void **)*args;
+
+	if (arg == NULL)
+		return fprintf(f, "(nil)");
+
+	switch (info->width) {
+	case 2: // struct rte_ether_addr *
+		const struct rte_ether_addr *mac = arg;
+		return fprintf(
+			f,
+			"%02hhx:%02hhx:%02hhx:%02hhx:%02hhx:%02hhx",
+			mac->addr_bytes[0],
+			mac->addr_bytes[1],
+			mac->addr_bytes[2],
+			mac->addr_bytes[3],
+			mac->addr_bytes[4],
+			mac->addr_bytes[5]
+		);
+	case 4: // ip4_addr_t *
+		inet_ntop(AF_INET, arg, buf, sizeof(buf));
+		return fprintf(f, "%s", buf);
+	case 6: // struct rte_ipv6_addr *
+		inet_ntop(AF_INET6, arg, buf, sizeof(buf));
+		return fprintf(f, "%s", buf);
+	}
+
+	return fprintf(f, "0x%lx", (uintptr_t)arg);
+}
+
+static int check_pointer_arg(const struct printf_info *, size_t n, int *argtypes, int *sizes) {
+	if (n == 1) {
+		sizes[0] = sizeof(void *);
+		argtypes[0] = PA_POINTER;
+		return 1;
+	}
+	return -1;
+}
+
+static void __attribute__((constructor, used)) init(void) {
+	if (register_printf_specifier('p', format_pointer, check_pointer_arg) < 0)
+		abort();
+}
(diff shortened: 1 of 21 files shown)


vjardin:
What's about being inspired by https://github.com/FRRouting/frr/tree/master/tools/gcc-plugins ?

rjarry:
I will add a unit test to verify that this works as expected.

modules/infra/control/port.c
```diff
@@ -448,10 +448,10 @@ static int port_mac_add(struct iface *iface, const struct rte_ether_addr *mac) {
 	for (i = 0; i < filter->count; i++) {
 		if (rte_is_same_ether_addr(&filter->mac[i], mac)) {
 			LOG(DEBUG,
-			    "%s: %s mac " ETH_ADDR_FMT " already filtered (refs=%u)",
+			    "%s: %s mac %2p already filtered (refs=%u)",
```
vjardin, line 451: instead of using %np all around into the code, which will lead to maintenance complexity, let's keep a FMT with the proper size. For instance:

ETH_ADDR_FMT "%2p"

and same for the other cases

So only those FMT will need to be updated if we come with something different some days.

modules/ip6/cli/route.c
```diff
@@ -74,11 +73,9 @@ static cmd_status_t route6_list(const struct gr_api_client *c, const struct ec_p
 	for (size_t i = 0; i < resp->n_routes; i++) {
 		struct libscols_line *line = scols_table_new_line(table, NULL);
 		const struct gr_ip6_route *route = &resp->routes[i];
-		ip6_net_format(&route->dest, dest, sizeof(dest));
-		inet_ntop(AF_INET6, &route->nh, nh, sizeof(nh));
 		scols_line_sprintf(line, 0, "%u", route->vrf_id);
-		scols_line_set_data(line, 1, dest);
-		scols_line_set_data(line, 2, nh);
+		scols_line_sprintf(line, 1, "%6p/%hhu", &route->dest, route->dest.prefixlen);
+		scols_line_sprintf(line, 2, "%6p", &route->nh);
```
vjardin, line 78: typical complexity here, FMT is needed for "%6p"

modules/infra/control/port.c
rjarry: I have added back defines for these printf specifiers. I only shortened them to avoid logs spanning over multiple lines as much as possible. The idea was to make the code more readable.

modules/ip6/cli/route.c
rjarry: done

rjarry:
I have updated the PR initial comment with the new description.

vjardin:
LGTM

api/printf.c
```diff
@@ -0,0 +1,54 @@
...
+		);
+	case 4: // ip4_addr_t *
+		inet_ntop(AF_INET, arg, buf, sizeof(buf));
+		return fprintf(f, "%s", buf);
+	case 6: // struct rte_ipv6_addr *
+		inet_ntop(AF_INET6, arg, buf, sizeof(buf));
+		return fprintf(f, "%s", buf);
+	}
+
+	return fprintf(f, "0x%lx", (uintptr_t)arg);
```
david-marchand, line 39: Quick nit, looking at the merged code.
This should use PRIxPTR (coming from inttypes.h).

PR #104, opened 2024-12-03
Generic nexthops

rjarry:


vjardin:
I try to understand the benefits: it is a nice code clean up, but then, down the road, how could it help to drop the `rte_spinlock_lock(&nh->lock);` from the datapath (arp input) ?

rjarry:
Yes that is the endgame goal. For now, I am planing on having a common code path between ipv4 and ipv6 for the nexthop handling, ageing, etc.

And finally, I will move all datapath write operations on nexthops and routing tables in the slow path.

PR #105, opened 2024-12-03
Iface internal notifications

christophefontaine:
Add notifications for up/down status, and trigger post-reconfig event for virtual interfaces as well.
This post-reconfig event is required when a 3rd party system will want to get a notification when the VRF associated to an interface is modified.

### DIFF
--- a/modules/infra/control/port.c
+++ b/modules/infra/control/port.c
@@ -302,7 +302,7 @@ int iface_port_reconfig(
 	if (stopped && (ret = rte_eth_dev_start(p->port_id)) < 0)
 		return errno_log(-ret, "rte_eth_dev_start");
 
-	iface_event_notify(IFACE_EVENT_PORT_POST_RECONFIG, iface);
+	iface_event_notify(IFACE_EVENT_POST_RECONFIG, iface);
 
 	return port_plug(p->port_id);
 }
--- a/modules/infra/control/port.c
+++ b/modules/infra/control/port.c
@@ -267,9 +267,11 @@ int iface_port_reconfig(
 		if (flags & GR_IFACE_F_UP) {
 			ret = rte_eth_dev_set_link_up(p->port_id);
 			iface->flags |= GR_IFACE_F_UP;
+			iface_event_notify(IFACE_EVENT_STATUS_UP, iface);
 		} else {
 			ret = rte_eth_dev_set_link_down(p->port_id);
 			iface->flags &= ~GR_IFACE_F_UP;
+			iface_event_notify(IFACE_EVENT_STATUS_DOWN, iface);
 		}
 		if (ret < 0)
 			errno_log(-ret, "rte_eth_dev_set_link_{up,down}");
@@ -634,17 +636,20 @@ static void link_event_cb(evutil_socket_t, short /*what*/, void * /*priv*/) {
 			if (rte_eth_link_get_nowait(qmap->port_id, &link) < 0) {
 				LOG(INFO, "%s: link status down", iface->name);
 				iface->state &= ~GR_IFACE_S_RUNNING;
+				iface_event_notify(IFACE_EVENT_STATUS_DOWN, iface);
 				continue;
 			}
 			if (link.link_status == RTE_ETH_LINK_UP) {
 				if (!(iface->state & GR_IFACE_S_RUNNING)) {
 					LOG(INFO, "%s: link status up", iface->name);
 					iface->state |= GR_IFACE_S_RUNNING;
+					iface_event_notify(IFACE_EVENT_STATUS_UP, iface);
 				}
 			} else {
 				if (iface->state & GR_IFACE_S_RUNNING) {
 					LOG(INFO, "%s: link status down", iface->name);
 					iface->state &= ~GR_IFACE_S_RUNNING;
+					iface_event_notify(IFACE_EVENT_STATUS_DOWN, iface);
 				}
 				continue;
 			}
--- a/modules/infra/control/vlan.c
+++ b/modules/infra/control/vlan.c
@@ -262,7 +262,35 @@ static struct gr_module vlan_module = {
 	.fini_prio = 1000,
 };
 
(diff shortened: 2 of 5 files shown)


modules/infra/control/port.c
```diff
@@ -267,9 +267,11 @@ int iface_port_reconfig(
 		if (flags & GR_IFACE_F_UP) {
 			ret = rte_eth_dev_set_link_up(p->port_id);
 			iface->flags |= GR_IFACE_F_UP;
+			iface_event_notify(IFACE_EVENT_STATUS_UP, iface);
```
rjarry, line 270: This is just the administrative state. Don't you think we should only rely on the runtime state in link_event_cb?

rjarry (commented):
I have one minor remark. But overall, looks good to me.

modules/infra/control/port.c
christophefontaine: In that case, should we have 2 flags, for the physical link status and the administrative status ?
As of now, we don't have the admin state, so we can't put a vlan interface down for example.

rjarry: We do have two separate flag masks. iface->flags (for administrative status) and iface->state (for runtime status)

rjarry: VLAN interfaces runtime state is not properly managed though. It should follow their parent port runtime state.

modules/infra/control/vlan.c
```diff
@@ -129,6 +129,8 @@ static int iface_vlan_reconfig(
 	if (set_attrs & GR_IFACE_SET_VRF)
 		iface->vrf_id = vrf_id;
 
+	iface_event_notify(IFACE_EVENT_POST_RECONFIG, iface);
```
vjardin, line 132: could we get from this API the iface state before and the newly iface state ?

rjarry: What do you mean by "from the API"?

rjarry:
@christophefontaine could you add another fix in this series to avoid the workaround you added here: https://github.com/DPDK/grout/pull/100/files#r1866149506

```diff
diff --git a/modules/infra/cli/iface.c b/modules/infra/cli/iface.c
index d0ca91669391..77973709f1d7 100644
--- a/modules/infra/cli/iface.c
+++ b/modules/infra/cli/iface.c
@@ -295,6 +295,7 @@ static cmd_status_t iface_list(const struct gr_api_client *c, const struct ec_pn
                        // type
                        scols_line_set_data(line, 4, type->name);
                        // info
+                       buf[0] = 0;
                        type->list_info(c, iface, buf, sizeof(buf));
                        scols_line_set_data(line, 5, buf);
                }
```

modules/infra/control/vlan.c
vjardin: We need a third argument with the iface before. I know that it seems painful but it'll help the management planes and control planes to diff it.

christophefontaine: What about having a IFACE_EVENT_PRE_RECONFIG in addition to POST ?
This wouldn't require any modification in the API, and we may be able to get this diff.

vjardin: OK, then with a transaction ID (whatever uuid or counter) that shall be the same for the IFACE_EVENT_PRE_RECONFIG and then IFACE_EVENT_POST_RECONFIG so the management plan can correlate both.

christophefontaine: The PRE_RECONFIG event and the transition id may not be required at all.
One of the assumption is that the mgmt app will be on the same node (grout uses a unix socket for its configuration).
When the management app connects, it reads the *current* state of the interfaces, so it only needs an event and the associated iface.

So, it can make a diff between the known state and the new state (hence the POST_RECONFIF notification).
WDYT ?

I have an ugly code snippet with FRR, for a dplane plugin: (Far from being ready, I just use it to validate that the notification system is working and useful. )
```
	if(gr_api_client_recv_notification(grout_ctx.notifs, &n) == 0) {
		switch (n->type) {
		case IFACE_EVENT_POST_ADD:
		case IFACE_EVENT_STATUS_UP:
		case IFACE_EVENT_STATUS_DOWN:
		case IFACE_EVENT_POST_RECONFIG:
			if (n->payload_len == sizeof(*p)) {
				p = (struct gr_infra_iface_get_resp *)&n[1];
				zlog_debug("Iface %s: %s", evt_to_str(n->type) , p->iface.name);
				iface = if_get_by_name(p->iface.name, p->iface.vrf_id, NULL);
				if(iface == NULL)
					break;
				sync_iface_status(iface, &p->iface);
			};
			break;
		case IFACE_EVENT_PRE_REMOVE:
			if (n->payload_len == sizeof(*p)) {
				p = (struct gr_infra_iface_get_resp *)&n[1];
				zlog_debug("Iface %s: %s", evt_to_str(n->type) , p->iface.name);
				iface = if_get_by_name(p->iface.name, p->iface.vrf_id, NULL);
				if(iface == NULL)
					break;
				if_delete(&iface);
			};
			break;
		default:
			zlog_debug("Unknown notification 0x%x received", n->type);
			break;
		}
```

modules/infra/control/vlan.c
```diff
@@ -262,7 +262,36 @@ static struct gr_module vlan_module = {
...
+	while ((vlan = iface_next(GR_IFACE_TYPE_VLAN, vlan)) != NULL) {
+		info = (struct iface_info_vlan *)vlan->info;
+		if (info->parent_id == iface->id) {
+			if (event == IFACE_EVENT_STATUS_UP) {
+				vlan->flags |= GR_IFACE_F_UP;
+			} else {
+				vlan->flags &= ~GR_IFACE_F_UP;
+			}
+			iface_event_notify(event, vlan);
+			break;
```
rjarry, line 284: You should continue here, in case there are multiple sub interfaces for a single port.

christophefontaine: good catch !
Will fix it now.

rjarry (approved):

PR #107, opened 2024-12-05
mk: add short target to build with debug options

rjarry:
Add a debug target that will force BUILDTYPE=debug and SANITIZE=address.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -16,6 +16,11 @@ endif
 all: $(BUILDDIR)/build.ninja
 	$Q ninja -C $(BUILDDIR) $(ninja_opts)
 
+.PHONY: debug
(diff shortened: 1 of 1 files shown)


david-marchand:
One problem with this patch is that the debug target does not "configure" the sources.

Example:
```
$ git clean -xffd
$ make
...
Found ninja-1.11.1 at /usr/bin/ninja
ninja: Entering directory `build'
[704/704] Linking target grout

$ make debug
ninja: Entering directory `build'
ninja: no work to do.
```

rjarry:
For this we would have to use meson configure and it can be a bit fragile. Sometimes, not all sources get recompiled with new options.

```sh
rm -rf build && make debug
```

Seems safer but I don't want to always delete everything.

PR #108, opened 2024-12-06
api: fix deadlock on shutdown when clients are connected

rjarry:
The docstring of event_base_foreach_event() states explicitly that modifying events in the callback function is unsafe and will cause crashes.

event_free_finalize tries to acquire a lock that is already held when the callback is called. Leading to a deadlock:

```
 (gdb) bt
 ...
 #3  ___pthread_mutex_lock (mutex=0x504000001550) at pthread_mutex_lock.c:93
 #4  0x00007f3f211b8485 in event_finalize_impl_ (flags=65536, ev=0x50c000005200, cb=0x4031ba <finalize_fd>)
 #5  0x00000000004042ed in close_connections (ev=0x50c000005200) at main/api.c:174
 ...
 #10 0x00007f3f211b263a in event_base_foreach_event (base=0x517000006d00, fn=0x40429f <close_connections>, arg=0x0)
 #11 0x0000000000404a3a in api_socket_stop () at main/api.c:253
 #12 0x00000000004072aa in main (argc=4, argv=0x7ffcc575d898) at main/main.c:210
```

Only use event_base_foreach_event() for iterating over the events that we actually want to free (namely, ones that have read_cb and write_cb as callbacks).

Only *after* returning from event_base_foreach_event(), call event_free_finalize on all these events.

Fixes: 865332077351 ("main: close active connections on shutdown")

### DIFF
--- a/main/api.c
+++ b/main/api.c
@@ -8,6 +8,7 @@
 #include <gr_api.h>
 #include <gr_log.h>
 #include <gr_macro.h>
+#include <gr_vec.h>
 
 #include <event2/event.h>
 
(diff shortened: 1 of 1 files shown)


christophefontaine:
Thank you: Tested, lgtm.

PR #109, opened 2024-12-10
Arp ndp control plane

rjarry:


david-marchand:
On the principle, it looks good.
Just needs some fixing on the CI side.

modules/ip/control/nexthop.c
```diff
@@ -32,6 +33,80 @@ struct nexthop *ip4_nexthop_lookup(uint16_t vrf_id, ip4_addr_t ip) {
...
+		struct ip_output_mbuf_data *d = ip_output_mbuf_data(m);
+		d->nh = nh;
+		if (post_to_stack(ip_output_node, m) < 0) {
+			LOG(ERR, "post_to_stack: %s", strerror(errno));
+			goto free;
+		}
+		return;
+	}
+
+	if (nh->held_pkts_num < NH_MAX_HELD_PKTS) {
```
david-marchand, line 91: Just a note that holding on this packet in control path means that the data path packet pool may end up in a starvation situation. Imagine many packets being queued, with unresolved nh, and no packet left in the packet pool to receive a arp reply.
Grout would probably break from this situation later (as packets are held with a timeout), but still it would be more robust to hold on a packet copy and free the original mbuf.

modules/ip/datapath/arp_input.c
```diff
@@ -17,125 +17,75 @@
...
 			edge = OP_REPLY;
 			break;
 		default:
 			edge = OP_UNSUPP;
 			goto next;
 		}
 
-		sip = arp->arp_data.arp_sip;
-		iface = eth_input_mbuf_data(mbuf)->iface;
-		local = ip4_addr_get_preferred(iface->id, sip);
```
david-marchand, line 111: It seems the behavior around lookups is changed in this commit at the same time the moving from dp to cp is done. Please split unrelated changes in separate patches.

modules/ip6/datapath/ndp_ns_input.c
```diff
@@ -83,20 +86,62 @@ static uint16_t ndp_ns_input_process(
...
+			// Otherwise, the node MUST set the Solicited flag to one and unicast the
+			// advertisement to the Source Address of the solicitation.
+			solicited = true;
+		}
+
 		// Reuse the mbuf to forge a neighbour advertisement reply.
-		// XXX,TODO: avoid issues with encap, remember previous data offset?
+		offset = mbuf->data_off;
 		rte_pktmbuf_reset(mbuf);
+		mbuf->data_off = offset;
```
david-marchand, line 137: Ugh!

modules/infra/control/nexthop.c
```diff
@@ -192,7 +191,6 @@ void nexthop_decref(struct nexthop *nh) {
 		}
 		rte_mempool_put(nhp->mp, nh);
 		memset(nh, 0, sizeof(*nh));
-		rte_spinlock_unlock(&nh->lock);
```
david-marchand, line 195: Good thing we are removing this spinlock as it was buggy here:
- previous line memset released the lock,
- a concurrent thread spinning would acquire the lock,
- this code then would incorrectly release the lock,

modules/ip/datapath/arp_input.c
rjarry: done

modules/ip6/datapath/ndp_ns_input.c
rjarry: I removed this. Completely unrelated. I wonder if calling `rte_pktmbuf_reset()` is required after all.

david-marchand: This seems more like a global scrub on the packet without knowing what was the reason.
I agree, rte_pktmbuf_reset() does not seem required.

rjarry: And that would partially address the concern I had about the data offset.

david-marchand:
Compilation is broken when stopping at: 5dfa958 - (HEAD) arp: reply to all local iface addresses (81 minutes ago) <Robin Jarry>.


``
$ git rbi origin/main -x 'make'
...

../modules/ip/datapath/arp_input.c: In function ‘arp_input_process’:
../modules/ip/datapath/arp_input.c:134:28: error: implicit declaration of function ‘arp_mbuf_data’; did you mean ‘mbuf_data’? [-Werror=implicit-function-declaration]
  134 |                 arp_data = arp_mbuf_data(mbuf);
      |                            ^~~~~~~~~~~~~
      |                            mbuf_data
../modules/ip/datapath/arp_input.c:134:26: error: assignment to ‘struct arp_mbuf_data *’ from ‘int’ makes pointer from integer without a cast [-Werror=int-conversion]
  134 |                 arp_data = arp_mbuf_data(mbuf);
      |                          ^
../modules/ip/datapath/arp_input.c:135:25: error: invalid use of undefined type ‘struct arp_mbuf_data’
  135 |                 arp_data->local = local;
      |                         ^~

```

**Comment by david-marchand** [comment]
With this last issue fixed, the series lgtm.

**Comment by rjarry** [comment]
Should be OK now. I have tested:

```sh
git rebase --exec 'make all lint unit-tests && sudo smoke/run.sh build' upstream/main
```

christophefontaine:
Seems ipv6 is broken: there is no neighbor solicitation response (packets are "hold" in ip6_output), and no next hop is created from the control plane.

PR #112, opened 2024-12-19
l4 port redirect & fix loopback for ip6 packets

christophefontaine:
- Fixes the redirection of IPv6 packets
- Prepare the tunnel over UDP, so we can register for a specific UDP port
- simplifies the number of nodes by removing TCP /  SCTP dedicated nodes as we don't use them.


modules/infra/control/loopback.c
```diff
@@ -134,6 +134,7 @@ static void iface_loopback_poll(evutil_socket_t, short reason, void *ev_iface) {
 
 	// packet sent from linux tun iface, no need to compute checksum;
 	mbuf->ol_flags = RTE_MBUF_F_RX_IP_CKSUM_GOOD;
+	mbuf->packet_type = data[0] == 6 ? RTE_PTYPE_L3_IPV6 : RTE_PTYPE_L3_IPV4;
```
david-marchand, line 137: Ugh, ok.
I guess the previous patch relies on this.

This must be documented, or a different mechanism is needed.

modules/l4/l4_redirect_loopback.c
```diff
@@ -25,89 +26,66 @@ static uint16_t redirect_loopback_process(
...
-		rte_pktmbuf_prepend(mbuf, sizeof(*ip));
+		if (!d->iface) {
+			edge = NO_IFACE;
+			goto next;
+		}
+
+		if (mbuf->packet_type & RTE_PTYPE_L3_IPV4) {
+			struct ip_local_mbuf_data *d = ip_local_mbuf_data(mbuf);
+			struct rte_ipv4_hdr *ip;
+			ip = (struct rte_ipv4_hdr *)rte_pktmbuf_prepend(mbuf, sizeof(*ip));
```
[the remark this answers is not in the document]
christophefontaine: indeed, will fix it

modules/l4/l4_redirect_loopback.c
```diff
@@ -25,89 +26,66 @@ static uint16_t redirect_loopback_process(
...
+			ip->src_addr = d->src;
+			ip->dst_addr = d->dst;
+			ip->total_length = rte_cpu_to_be_16(d->len) + sizeof(*ip);
+			ip->next_proto_id = d->proto;
```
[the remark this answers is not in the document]
christophefontaine: Will also fix for ipv6 hdr.

modules/infra/control/loopback.c
christophefontaine: I wished to avoid setting unnecessary data if the packet type was already defined by the drivers.
The other option is to add a `family` field in the buffer metadata for both `ip_local_mbuf_data` and `ip6_local_mbuf_data` to define which kind of header we should fill in.
I found it uglier as the 2 priv data types are defined in different modules.

```
GR_MBUF_PRIV_DATA_TYPE(ip_local_mbuf_data, {
+     uint8_t family;
        ip4_addr_t src;
        ip4_addr_t dst;
        uint16_t len;
        uint16_t vrf_id;
        uint8_t proto;
        uint8_t ttl;
});


GR_MBUF_PRIV_DATA_TYPE(ip6_local_mbuf_data, {
+     uint8_t family;
        struct rte_ipv6_addr src;
        struct rte_ipv6_addr dst;
        uint16_t len;
        uint8_t hop_limit;
        uint8_t proto;
});
```

@rjarry Any opinion about that ? 
Should we define instead a common header for both local_mbuf_data instead ?
```
GR_MBUF_PRIV_DATA_TYPE(ip46_local_mbuf_data, {
        uint8_t family;
        union {
            struct rte_ipv6_addr v6;
            ip4_addr_t v4;  
        } src;
        union {
            struct rte_ipv6_addr v6;
            ip4_addr_t v4;  
        } dst;
        uint16_t len;
        union {
            uint8_t hop_limit;
            uint8_t ttl;
        };
        uint8_t proto;
});
```

modules/l4/l4_input_local.c
```diff
@@ -0,0 +1,96 @@
...
+		edge = udp_edges[hdr->dst_port];
+next:
+		rte_node_enqueue_x1(graph, node, edge, mbuf);
+	}
+	return nb_objs;
+}
+
+static void l4_input_local_register(void) {
+	ip_input_local_add_proto(IPPROTO_UDP, "l4_input_local");
+	ip6_input_local_add_proto(IPPROTO_TCP, "l4_input_local");
```
christophefontaine, line 77: ouch .... should register for both ip4/udp ip6/udp and ip4/tcp ip4/tcp

modules/infra/control/loopback.c
rjarry: oh dear. I think I prefer the `packet_type` hacking, even though it is ugly.

modules/l4/l4_redirect_loopback.c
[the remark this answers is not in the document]
rjarry: there are dedicated functions to fill in ip(v6) headers based on local data.

https://github.com/DPDK/grout/blob/713562721e8554bc26841663e207ddfb1c412e63/modules/ip/datapath/gr_ip4_datapath.h#L40-L52

https://github.com/DPDK/grout/blob/713562721e8554bc26841663e207ddfb1c412e63/modules/ip6/datapath/gr_ip6_datapath.h#L36-L49

modules/l4/l4_redirect_loopback.c
```diff
@@ -25,89 +26,66 @@ static uint16_t redirect_loopback_process(
...
-static struct rte_node_register udp_redirect_loopback_node = {
-	.name = "udp_redirect_loopback",
-	.process = redirect_loopback_process,
-	.nb_edges = EDGE_COUNT,
-	.next_nodes = {
-		[REDIRECT] = "loopback_output",
-		[NO_IFACE] = "no_loop_iface",
-	},
-};
-
```
rjarry, line 68: I liked the idea of having two separate nodes for tcp and udp. We may not want to register the same local ports depending on the protocol.

Maybe these ports should be renamed to tcp_input_local and udp_input_local which would (or would not) redirect to loopback_output after reconstructing the ip(v6) headers.

modules/l4/l4_redirect_loopback.c
```diff
@@ -25,89 +26,66 @@ static uint16_t redirect_loopback_process(
...
-static struct rte_node_register sctp_redirect_loopback_node = {
-	.name = "sctp_redirect_loopback",
```
rjarry, line 70: However, sctp is probably out of scope.

rjarry (changes requested):

modules/l4/l4_redirect_loopback.c
christophefontaine: This is already the case, the register function already takes a proto as a parameter, so we can select TCP/UDP/...
We only support udp port as of now, as I'm not yet aware of a TCP based tunnel.
All non-handled traffic will be redirected to the loop interface.
Even this behavior may have to be refined, in a future patch.

modules/infra/control/loopback.c
christophefontaine: I documented the usage of packet_type.

rjarry (approved):

PR #113, opened 2024-12-19
Ip6 scoped linklocal

christophefontaine:
Scoped link local addresses implementation

Still has some issues:
```
grout# add ip6 route ::0/0 via fe80::4e77:6dff:fe7d:1cc7
error: command failed: No route to host
```

This link local address doesn't exist in the FIB, but fe80:100::4e77:6dff:fe7d:1cc7 and fe80:300::4e77:6dff:fe7d:1cc7 are present internally.

```
grout# add ip6 route ::0/0 via fe80:100::4e77:6dff:fe7d:1cc7
grout# show ip6 route
VRF  DESTINATION                    NEXT_HOP
0    fe80::209:c000:17e:f6b/128     fe80::209:c000:17e:f6b
0    fe80::4e77:6dff:fe7d:1cc7/128  fe80::4e77:6dff:fe7d:1cc7
0    fe80::1:1/128                  fe80::1:1
0    fe80::4e77:6dff:fe7d:1cc7/128  fe80::4e77:6dff:fe7d:1cc7
0    fe80::90e2:ba00:314:2f8c/128   fe80::90e2:ba00:314:2f8c
0    ::/0                           fe80::4e77:6dff:fe7d:1cc7
```

### DIFF
--- a/modules/infra/control/nexthop.c
+++ b/modules/infra/control/nexthop.c
@@ -150,6 +150,7 @@ void nh_pool_iter(struct nh_pool *nhp, nh_iter_cb_t nh_cb, void *priv) {
 
 struct lookup_filter {
 	uint16_t vrf_id;
+	uint16_t iface_id;
 	uint8_t family;
 	const void *addr;
 	struct nexthop *nh;
@@ -167,14 +168,20 @@ static void nh_lookup_cb(struct nexthop *nh, void *priv) {
 			filter->nh = nh;
 		break;
 	case AF_INET6:
-		if (rte_ipv6_addr_eq(&nh->ipv6, filter->addr))
-			filter->nh = nh;
+		if (rte_ipv6_addr_eq(&nh->ipv6, filter->addr)) {
+			bool is_linklocal = rte_ipv6_addr_is_linklocal(&nh->ipv6);
+			if (!is_linklocal || nh->iface_id == filter->iface_id)
+				filter->nh = nh;
+		}
 		break;
 	}
 }
 
-struct nexthop *nexthop_lookup(struct nh_pool *nhp, uint16_t vrf_id, const void *addr) {
-	struct lookup_filter filter = {.family = nhp->family, .vrf_id = vrf_id, .addr = addr};
+struct nexthop *
+nexthop_lookup(struct nh_pool *nhp, uint16_t vrf_id, uint16_t iface_id, const void *addr) {
+	struct lookup_filter filter = {
+		.family = nhp->family, .vrf_id = vrf_id, .iface_id = iface_id, .addr = addr
+	};
 	nh_pool_iter(nhp, nh_lookup_cb, &filter);
 	return filter.nh ?: errno_set_null(ENOENT);
 }
--- a/modules/ip/control/nexthop.c
+++ b/modules/ip/control/nexthop.c
@@ -31,7 +31,8 @@ struct nexthop *ip4_nexthop_new(uint16_t vrf_id, uint16_t iface_id, ip4_addr_t i
 }
 
 struct nexthop *ip4_nexthop_lookup(uint16_t vrf_id, ip4_addr_t ip) {
-	return nexthop_lookup(nh_pool, vrf_id, &ip);
+	// XXX: should we scope ip4 nh lookup based on rfc3927 ?
+	return nexthop_lookup(nh_pool, vrf_id, GR_IFACE_ID_UNDEF, &ip);
 }
 
 static control_input_t ip_output_node;
--- a/modules/ip6/control/address.c
+++ b/modules/ip6/control/address.c
@@ -90,8 +90,8 @@ static int ip6_mcast_addr_add(struct iface *iface, const struct rte_ipv6_addr *i
 	if (i == ARRAY_DIM(maddrs->nh))
 		return errno_set(ENOSPC);
 
-	if ((nh = ip6_nexthop_lookup(iface->vrf_id, ip)) == NULL) {
-		if ((nh = ip6_nexthop_new(iface->vrf_id, GR_IFACE_ID_UNDEF, ip)) == NULL)
+	if ((nh = ip6_nexthop_lookup(iface->vrf_id, iface->id, ip)) == NULL) {
+		if ((nh = ip6_nexthop_new(iface->vrf_id, iface->id, ip)) == NULL)
 			return errno_set(-errno);
 		rte_ether_mcast_from_ipv6(&nh->lladdr, ip);
 	}
@@ -155,7 +155,7 @@ iface6_addr_add(const struct iface *iface, const struct rte_ipv6_addr *ip, uint8
 	if (addrs->count == ARRAY_DIM(addrs->nh))
 		return errno_set(ENOSPC);
 
-	if (ip6_nexthop_lookup(iface->vrf_id, ip) != NULL)
+	if (ip6_nexthop_lookup(iface->vrf_id, iface->id, ip) != NULL)
 		return errno_set(EADDRINUSE);
 
 	if ((nh = ip6_nexthop_new(iface->vrf_id, iface->id, ip)) == NULL)
@@ -170,7 +170,7 @@ iface6_addr_add(const struct iface *iface, const struct rte_ipv6_addr *ip, uint8
 			return errno_set(-ret);
 		}
 
-	if ((ret = ip6_route_insert(iface->vrf_id, &nh->ipv6, nh->prefixlen, nh)) < 0)
+	if ((ret = ip6_route_insert(iface->vrf_id, iface->id, ip, nh->prefixlen, nh)) < 0)
 		return errno_set(-ret);
 
 	addrs->nh[addr_index] = nh;
--- a/modules/ip6/control/address.c
+++ b/modules/ip6/control/address.c
@@ -298,12 +298,7 @@ static void ip6_iface_event_handler(iface_event_t event, struct iface *iface) {
 	case IFACE_EVENT_POST_ADD:
 		if (iface_get_eth_addr(iface->id, &mac) == 0) {
 			rte_ipv6_llocal_from_ethernet(&link_local, &mac);
-			// Avoid address conflicts with VLAN interfaces (same mac address)
-			// Maybe we should do better than this and fallback on a pseudo random
-			// algorithm (such as SLAAC, RFC 7217).
-			link_local.a[11] = (iface->id >> 8) & 0xff;
-			link_local.a[12] = iface->id & 0xff;
-			if (iface6_addr_add(iface, &link_local, RTE_IPV6_MAX_DEPTH) < 0)
+			if (iface6_addr_add(iface, &link_local, 64) < 0)
 				errno_log(errno, "iface_addr_add");
 
 			rte_ipv6_solnode_from_addr(&solicited_node, &link_local);
--- a/modules/ip6/control/gr_ip6_control.h
+++ b/modules/ip6/control/gr_ip6_control.h
@@ -20,18 +20,28 @@
 #define IP6_MAX_NEXT_HOPS (1 << 16)
 #define IP6_MAX_ROUTES (1 << 16)
 
-struct nexthop *ip6_nexthop_lookup(uint16_t vrf_id, const struct rte_ipv6_addr *);
+struct nexthop *
+ip6_nexthop_lookup(uint16_t vrf_id, uint16_t iface_id, const struct rte_ipv6_addr *);
 struct nexthop *ip6_nexthop_new(uint16_t vrf_id, uint16_t iface_id, const struct rte_ipv6_addr *);
 
 void ip6_nexthop_unreachable_cb(struct rte_mbuf *m);
 void ndp_probe_input_cb(struct rte_mbuf *m);
 
-int ip6_route_insert(uint16_t vrf_id, const struct rte_ipv6_addr *, uint8_t prefixlen, struct nexthop *);
-int ip6_route_delete(uint16_t vrf_id, const struct rte_ipv6_addr *, uint8_t prefixlen);
+int ip6_route_insert(uint16_t vrf_id, uint16_t iface_id, const struct rte_ipv6_addr *, uint8_t prefixlen, struct nexthop *);
+int ip6_route_delete(
+	uint16_t vrf_id,
+	uint16_t iface_id,
+	const struct rte_ipv6_addr *,
+	uint8_t prefixlen
+);
 void ip6_route_cleanup(struct nexthop *);
-struct nexthop *ip6_route_lookup(uint16_t vrf_id, const struct rte_ipv6_addr *);
-struct nexthop *
-ip6_route_lookup_exact(uint16_t vrf_id, const struct rte_ipv6_addr *, uint8_t prefixlen);
+struct nexthop *ip6_route_lookup(uint16_t vrf_id, uint16_t iface_id, const struct rte_ipv6_addr *);
+struct nexthop *ip6_route_lookup_exact(
+	uint16_t vrf_id,
+	uint16_t iface_id,
+	const struct rte_ipv6_addr *,
+	uint8_t prefixlen
+);
 
 // get the default address for a given interface
 struct nexthop *ip6_addr_get_preferred(uint16_t iface_id, const struct rte_ipv6_addr *);
@@ -40,4 +50,30 @@ struct hoplist *ip6_addr_get_all(uint16_t iface_id);
 // determine if the given interface is member of the provided multicast address group
 struct nexthop *ip6_mcast_get_member(uint16_t iface_id, const struct rte_ipv6_addr *mcast);
 
+static inline const struct rte_ipv6_addr *ip6_addr_linklocal_scope(
+	const struct rte_ipv6_addr *ip,
+	struct rte_ipv6_addr *scoped_ip,
+	uint16_t iface_id
+) {
+	if (rte_ipv6_addr_is_linklocal(ip)) {
+		*scoped_ip = *ip;
+		scoped_ip->a[2] = (iface_id >> 8) & 0xff;
+		scoped_ip->a[3] = iface_id & 0xff;
+		return scoped_ip;
+	} else {
+		return ip;
+	}
+}
+
+static inline const struct rte_ipv6_addr *
+ip6_addr_linklocal_unscope(const struct rte_ipv6_addr *ip, struct rte_ipv6_addr *unscoped_ip) {
+	if (rte_ipv6_addr_is_linklocal(ip)) {
+		*unscoped_ip = *ip;
+		unscoped_ip->a[2] = 0;
+		unscoped_ip->a[3] = 0;
+		return unscoped_ip;
+	} else {
+		return ip;
+	}
+}
 #endif
--- a/modules/ip6/control/nexthop.c
+++ b/modules/ip6/control/nexthop.c
@@ -33,30 +33,33 @@ ip6_nexthop_new(uint16_t vrf_id, uint16_t iface_id, const struct rte_ipv6_addr *
 	return nexthop_new(nh_pool, vrf_id, iface_id, ip);
 }
 
-struct nexthop *ip6_nexthop_lookup(uint16_t vrf_id, const struct rte_ipv6_addr *ip) {
-	return nexthop_lookup(nh_pool, vrf_id, ip);
+struct nexthop *
+ip6_nexthop_lookup(uint16_t vrf_id, uint16_t iface_id, const struct rte_ipv6_addr *ip) {
+	return nexthop_lookup(nh_pool, vrf_id, iface_id, ip);
 }
 
 static control_input_t ip6_output_node;
 
 void ip6_nexthop_unreachable_cb(struct rte_mbuf *m) {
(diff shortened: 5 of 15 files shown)


christophefontaine:
I'll update the API to allow the client to specify  the interface.

modules/ip6/control/address.c
```diff
@@ -76,24 +76,25 @@ struct nexthop *ip6_mcast_get_member(uint16_t iface_id, const struct rte_ipv6_ad
...
 	if (i == ARRAY_DIM(maddrs->nh))
 		return errno_set(ENOSPC);
 
-	if ((nh = ip6_nexthop_lookup(iface->vrf_id, ip)) == NULL) {
-		if ((nh = ip6_nexthop_new(iface->vrf_id, GR_IFACE_ID_UNDEF, ip)) == NULL)
+	if ((nh = ip6_nexthop_lookup(iface->vrf_id, &ip)) == NULL) {
+		if ((nh = ip6_nexthop_new(iface->vrf_id, GR_IFACE_ID_UNDEF, &ip)) == NULL)
 			return errno_set(-errno);
-		rte_ether_mcast_from_ipv6(&nh->lladdr, ip);
+		rte_ether_mcast_from_ipv6(&nh->lladdr, &ip);
```
rjarry, line 97: I think I missed something. Why are you introducing a new local ip variable to use it instead of the const parameter?

modules/ip6/control/address.c
```diff
@@ -135,9 +136,10 @@ static int ip6_mcast_addr_del(struct iface *iface, const struct rte_ipv6_addr *i
 }
 
 static int
-iface6_addr_add(const struct iface *iface, const struct rte_ipv6_addr *ip, uint8_t prefixlen) {
+iface6_addr_add(const struct iface *iface, const struct rte_ipv6_addr *ipv6, uint8_t prefixlen) {
 	struct hoplist *addrs;
 	unsigned addr_index;
+	struct rte_ipv6_addr *ip = (struct rte_ipv6_addr *)ipv6;
```
rjarry, line 142: if we need to change the signature of the function we can discuss it, no worries. But please no hacks like this.

modules/ip6/control/address.c
```diff
@@ -180,15 +183,17 @@ iface6_addr_add(const struct iface *iface, const struct rte_ipv6_addr *ip, uint8
...
 	struct iface *iface;
 	int ret;
 
 	iface = iface_from_id(req->addr.iface_id);
 	if (iface == NULL)
 		return api_out(errno, 0);
+	ip = req->addr.addr.ip;
 
-	if ((ret = iface6_addr_add(iface, &req->addr.addr.ip, req->addr.addr.prefixlen)) < 0)
+	ip6_addr_linklocal_scope(&ip, req->addr.iface_id);
```
rjarry, line 195: you are adding the iface id scope twice: here and in iface6_addr_add.

modules/ip6/control/address.c
```diff
@@ -212,8 +217,11 @@ static struct api_out addr6_del(const void *request, void ** /*response*/) {
 	if ((addrs = ip6_addr_get_all(req->addr.iface_id)) == NULL)
 		return api_out(errno, 0);
 
+	scoped = req->addr.addr.ip;
+	ip6_addr_linklocal_scope(&scoped, req->addr.iface_id);
```
rjarry, line 221: I think it is safe to scope the address present in the request. no need for a local variable.

modules/ip6/control/gr_ip6_control.h
```diff
@@ -40,4 +40,18 @@ struct hoplist *ip6_addr_get_all(uint16_t iface_id);
 // determine if the given interface is member of the provided multicast address group
 struct nexthop *ip6_mcast_get_member(uint16_t iface_id, const struct rte_ipv6_addr *mcast);
 
+static inline int ip6_addr_linklocal_scope(struct rte_ipv6_addr *ip, uint16_t iface_id) {
+	if (!rte_ipv6_addr_is_linklocal(ip))
+		return errno_set(EFAULT);
```
rjarry, line 45: EINVAL would be more appropriate.

modules/ip6/control/gr_ip6_control.h
```diff
@@ -40,4 +40,18 @@ struct hoplist *ip6_addr_get_all(uint16_t iface_id);
 // determine if the given interface is member of the provided multicast address group
 struct nexthop *ip6_mcast_get_member(uint16_t iface_id, const struct rte_ipv6_addr *mcast);
 
+static inline int ip6_addr_linklocal_scope(struct rte_ipv6_addr *ip, uint16_t iface_id) {
```
rjarry, line 43: I am not fond of the function name but I don't have great alternatives to suggest either.

modules/ip6/control/gr_ip6_control.h
```diff
@@ -40,4 +40,18 @@ struct hoplist *ip6_addr_get_all(uint16_t iface_id);
...
+	if (!rte_ipv6_addr_is_linklocal(ip))
+		return errno_set(EFAULT);
+	((uint16_t *)ip->a)[1] = iface_id;
```
rjarry, line 46: It would be best to change this to network order to ensure consistent behavior across architectures.

```c
rte_be16_t scope_id = htons(iface_id);
// rte_ipv6_addr is aligned to 1, avoid alignment issues with casting to uint16_t
ip->a[2] = 0xff & (scope_id >> 8);
ip->a[3] = 0xff & scope_id;
```

modules/ip6/control/gr_ip6_control.h
```diff
@@ -40,4 +40,18 @@ struct hoplist *ip6_addr_get_all(uint16_t iface_id);
...
+	return 0;
+}
+
+static inline int ip6_addr_linklocal_unscope(struct rte_ipv6_addr *ip) {
+	if (!rte_ipv6_addr_is_linklocal(ip))
+		return errno_set(EFAULT);
+	((uint16_t *)ip->a)[1] = 0;
```
rjarry, line 53:
```c
// rte_ipv6_addr is aligned to 1, avoid alignment issues with casting to uint16_t
ip->a[2] = 0;
ip->a[3] = 0;
```

rjarry (changes requested):
Hi chris, thanks for working on this. I think we should read carefully the bsd stack and get inspiration from where they scope and unscope the addresses.

Here are a couple of pointers which may be good to check:

https://github.com/freebsd/freebsd-src/blob/9efd215411bb5ead2bc0ab208b4c19e46da0d2c9/sys/netinet6/ip6_input.c#L752-L770

https://github.com/freebsd/freebsd-src/blob/9efd215411bb5ead2bc0ab208b4c19e46da0d2c9/sys/netinet6/ip6_output.c#L738-L752

https://github.com/freebsd/freebsd-src/blob/9efd215411bb5ead2bc0ab208b4c19e46da0d2c9/sys/netinet6/ip6_output.c#L829-L861

modules/ip6/datapath/ip6_output.c
```diff
@@ -77,6 +77,7 @@ ip6_output_process(struct rte_graph *graph, struct rte_node *node, void **objs,
 			goto next;
 		}
 
+		ip6_addr_linklocal_unscope(&ip->src_addr);
```
rjarry, line 83: Maybe move this to nodes that send from local ip (ndp_na_output*, etc.)

modules/ip6/datapath/ndp_ns_input.c
```diff
@@ -80,7 +81,10 @@ static uint16_t ndp_ns_input_process(
 		// - Target Address is not a multicast address.
 		ASSERT_NDP(!rte_ipv6_addr_is_mcast(&ns->target));
 
-		local = ip6_nexthop_lookup(iface->vrf_id, &ns->target);
+		scoped_target = ns->target;
+		ip6_addr_linklocal_scope(&scoped_target, iface->id);
+
+		local = ip6_nexthop_lookup(iface->vrf_id, &scoped_target);
```
rjarry, line 87: Instead of scoping the nexthop address, I would find it more practical to only scope the route prefixes and leave the nexthop address unscoped. That requires to add an `iface_id` argument to `ip{,6}_nexthop_lookup`.

rjarry (changes requested):
Can you also add a commit that changes the /128 (incorrect) prefix to /64 for link local addresses?

And scope the link-local route prefixes only.

Bonus, update `ip6_addr_linklocal_{un,}scope` to take the original address and a local ip address and returns a (const) pointer to the original or modified address (depending whether the original address had to be copied and modified or not).

modules/ip6/datapath/ndp_ns_input.c
christophefontaine: Done in the new patch set.

modules/ip/control/nexthop.c
```diff
@@ -31,7 +31,7 @@ struct nexthop *ip4_nexthop_new(uint16_t vrf_id, uint16_t iface_id, ip4_addr_t i
 }
 
 struct nexthop *ip4_nexthop_lookup(uint16_t vrf_id, ip4_addr_t ip) {
-	return nexthop_lookup(nh_pool, vrf_id, &ip);
+	return nexthop_lookup(nh_pool, vrf_id, 0, &ip);
```
david-marchand, line 34: Why is ipv4 lookup different from the ipv6 one?

christophefontaine:
@rjarry with the route & next hop api update, the code is much cleaner, and all the technicalities related to the scoped addresses are hidden in `control/route.c` and `control/nexthop.c`.

I keep this PR as draft now to get a feedback on the API side: should we add a parameter to specify the interface ?
It would be only used for static routes with a link local next hop.
Or should we go 'as-is' (with David's fixes) and add this in a subsequent PR ?

modules/infra/control/nexthop.c
```diff
@@ -167,14 +168,20 @@ static void nh_lookup_cb(struct nexthop *nh, void *priv) {
 			filter->nh = nh;
 		break;
 	case AF_INET6:
-		if (rte_ipv6_addr_eq(&nh->ipv6, filter->addr))
-			filter->nh = nh;
+		if (rte_ipv6_addr_eq(&nh->ipv6, filter->addr)) {
+			bool is_linklocal = rte_ipv6_addr_is_linklocal(&nh->ipv6);
+			if (!is_linklocal || (is_linklocal && nh->iface_id == filter->iface_id))
```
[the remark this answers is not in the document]
christophefontaine: indeed ! Will fix it.

modules/ip/control/nexthop.c
christophefontaine: I (maybe wrongly) assumed that we don't need to scope the nexthop lookup for ipv4.
Do we also want to scope next hop and route lookup for ipv4 ?

david-marchand: I think there is also a notion of link local addresses with the 169.254.0.0/16 range.
https://datatracker.ietf.org/doc/html/rfc3927

But if we don't need/care about it, then we can keep this a IPv6 only thing in grout.

modules/ip6/control/address.c
```diff
@@ -299,12 +299,7 @@ static void ip6_iface_event_handler(iface_event_t event, struct iface *iface) {
 	case IFACE_EVENT_POST_ADD:
 		if (iface_get_eth_addr(iface->id, &mac) == 0) {
 			rte_ipv6_llocal_from_ethernet(&link_local, &mac);
-			// Avoid address conflicts with VLAN interfaces (same mac address)
-			// Maybe we should do better than this and fallback on a pseudo random
-			// algorithm (such as SLAAC, RFC 7217).
-			link_local.a[11] = (iface->id >> 8) & 0xff;
-			link_local.a[12] = iface->id & 0xff;
-			if (iface6_addr_add(iface, &link_local, RTE_IPV6_MAX_DEPTH) < 0)
+			if (iface6_addr_add(iface, &link_local, 64) < 0)
```
rjarry, line 302: Do we have other places with RTE_IPV6_MAX_DEPTH?

rjarry (changes requested):

modules/ip6/control/nexthop.c
```diff
@@ -72,7 +75,10 @@ void ip6_nexthop_unreachable_cb(struct rte_mbuf *m) {
 
 		// Create an associated /128 route so that next packets take it
 		// in priority with a single route lookup.
-		if (ip6_route_insert(nh->vrf_id, &ip->dst_addr, RTE_IPV6_MAX_DEPTH, remote) < 0) {
+		if (ip6_route_insert(
+			    nh->vrf_id, nh->iface_id, &ip->dst_addr, RTE_IPV6_MAX_DEPTH, remote
+		    )
+		    < 0) {
```
rjarry, line 81: clang-format does an ugly job here...

modules/ip6/control/gr_ip6_control.h
```diff
@@ -20,18 +20,28 @@
...
-struct nexthop *ip6_route_lookup(uint16_t vrf_id, const struct rte_ipv6_addr *);
-struct nexthop *
-ip6_route_lookup_exact(uint16_t vrf_id, const struct rte_ipv6_addr *, uint8_t prefixlen);
+struct nexthop *ip6_route_lookup(uint16_t vrf_id, uint16_t iface_id, const struct rte_ipv6_addr *);
+struct nexthop *ip6_route_lookup_exact(
+	uint16_t vrf_id,
+	uint16_t iface_id,
+	const struct rte_ipv6_addr *,
+	uint8_t prefixlen
+);
```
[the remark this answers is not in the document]
christophefontaine: there is no more scoping/unscoping in the caller contexts though ?

modules/ip6/control/nexthop.c
christophefontaine: yup. Maybe we could update clang-format rules ?

modules/ip6/control/gr_ip6_control.h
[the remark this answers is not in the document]
rjarry: The contexts where scoping/unscoping should happen would be very limited I think. I may be wrong though.

modules/ip6/control/address.c
christophefontaine: Yes, but this is done on purpose, to add a specific next hop or route.
The only occurrences of `RTE_IPV6_MAX_DEPTH` are in `modules/ip6/control/nexthop.c` and `modules/ip6/control/route.c`.

modules/ip6/control/nexthop.c
rjarry: If you manage to make it behave in that case, please do so :)

I tried to play with wrapping penalties but some things are obscure.

modules/ip6/control/address.c
```diff
@@ -271,6 +271,7 @@ static struct api_out addr6_list(const void *request, void **response) {
 			const struct nexthop *nh = addrs->nh[i];
 			addr = &resp->addrs[resp->n_addrs++];
 			addr->addr.ip = nh->ipv6;
+			ip6_addr_linklocal_unscope(&addr->addr.ip, &addr->addr.ip);
```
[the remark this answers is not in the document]
christophefontaine: This was to avoid an additional copy: when there is no need to unscope, the original data is neither touched or copied.
Here, even for non link local addresses, the ipv6 is copied systematically.

modules/ip6/control/gr_ip6_control.h
```diff
@@ -40,4 +50,30 @@ struct hoplist *ip6_addr_get_all(uint16_t iface_id);
...
+	}
+}
+
+static inline const struct rte_ipv6_addr *
+ip6_addr_linklocal_unscope(const struct rte_ipv6_addr *ip, struct rte_ipv6_addr *scoped_ip) {
+	if (rte_ipv6_addr_is_linklocal(ip)) {
+		*scoped_ip = *ip;
+		scoped_ip->a[2] = 0;
+		scoped_ip->a[3] = 0;
+		return scoped_ip;
```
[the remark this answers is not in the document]
christophefontaine: ack.

modules/ip6/control/gr_ip6_control.h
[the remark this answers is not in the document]
rjarry: Yeah I can see the benefit. But scoping should only happen on ip6_input and unscoping in ip6_output, and a couple of API functions, or am I forgetting stuff?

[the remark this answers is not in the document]
christophefontaine: every time you call `ip6_route_lookup`
- loop_input.c
- nexthop.c
- icmp6_output.c
- ip6_input.c

Unscoping isn't done in ip6_output anymore, as the unscopped ipv6 address is stored in the nexthop.
So, route_lookup is very similar to nexthop_lookup now.

modules/ip6/control/nexthop.c
christophefontaine: We'll have a look at that later :)

modules/ip/control/nexthop.c
christophefontaine: I added a comment, so we can work on this later.

modules/ip6/control/address.c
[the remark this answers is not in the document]
christophefontaine: removed the call in the new version as this call to unscope is useless.

modules/ip6/control/address.c
```diff
@@ -90,8 +90,8 @@ static int ip6_mcast_addr_add(struct iface *iface, const struct rte_ipv6_addr *i
 	if (i == ARRAY_DIM(maddrs->nh))
 		return errno_set(ENOSPC);
 
-	if ((nh = ip6_nexthop_lookup(iface->vrf_id, ip)) == NULL) {
-		if ((nh = ip6_nexthop_new(iface->vrf_id, GR_IFACE_ID_UNDEF, ip)) == NULL)
+	if ((nh = ip6_nexthop_lookup(iface->vrf_id, iface->id, ip)) == NULL) {
+		if ((nh = ip6_nexthop_new(iface->vrf_id, iface->id, ip)) == NULL)
```
rjarry, line 94: I am having second thoughts about scoping multicast addresses. Is this required?

modules/ip6/control/nexthop.c
```diff
@@ -33,26 +33,29 @@ ip6_nexthop_new(uint16_t vrf_id, uint16_t iface_id, const struct rte_ipv6_addr *
...
 
-	if (nh->flags & GR_NH_F_LINK && !rte_ipv6_addr_eq(&dst, &nh->ipv6)) {
+	if (nh->flags & GR_NH_F_LINK && !rte_ipv6_addr_eq(&ip->dst_addr, &nh->ipv6)) {
 		// The resolved nexthop is associated with a "connected" route.
 		// We currently do not have an explicit route entry for this
 		// destination IP.
-		struct nexthop *remote = ip6_nexthop_lookup(nh->vrf_id, &ip->dst_addr);
+		struct nexthop *remote = ip6_nexthop_lookup(
+			nh->vrf_id, mbuf_data(m)->iface->id, &ip->dst_addr
+		);
```
rjarry, line 58: I don't recall why we needed a local `dst` variable here. Maybe we should either always use `&ip->dst_addr` or always `dst`. But the mix of both is confusing. Especially since we are not modifying `dst`.

modules/ip6/control/nexthop.c
```diff
@@ -116,7 +122,7 @@ void ndp_probe_input_cb(struct rte_mbuf *m) {
 	const struct iface *iface = mbuf_data(m)->iface;
 	const struct icmp6_neigh_solicit *ns;
 	const struct icmp6_neigh_advert *na;
-	struct rte_ipv6_addr target;
+	const struct rte_ipv6_addr *target;
```
rjarry, line 125: Same remark as above. Do we really need a local variable? If so, why change its type to a pointer?

modules/ip6/datapath/ndp_ns_input.c
```diff
@@ -32,9 +32,9 @@ static uint16_t ndp_ns_input_process(
 ) {
 	const struct nexthop *local, *remote;
 	struct icmp6_neigh_solicit *ns;
+	struct rte_ipv6_addr src, dst;
 	struct icmp6_neigh_advert *na;
 	struct ip6_local_mbuf_data *d;
-	struct rte_ipv6_addr src, dst;
```
rjarry, line 37: probably no need for this diff

rjarry (changes requested):

modules/ip6/control/nexthop.c
christophefontaine: we do need a local var as target may be set either from na->target or from ns->target.
And it is a const pointer as we don't modify the value, and we need to get a pointer anyway for the calls to nh lookup/insert.

modules/ip6/control/nexthop.c
christophefontaine: yes, I'll fix that and use `dst` everywhere.

modules/ip6/datapath/ndp_ns_input.c
christophefontaine: indeed !

modules/ip6/control/nexthop.c
christophefontaine: I think `dst` was here to have a nice output with clang format.

rjarry (approved):

rjarry:
Thanks Chris!

PR #118, opened 2025-01-08
main: improve api socket bind

ogournet:
If the grout process is launched a second time, then socket api bind() failed and the process unlink socket file before exiting, making the first running process unreachable.

If the grout process crash, it won't remove api socket file, and the next invocation will fail at socket api bind().

Change api_socket_start() to check, if bind() fails, whether there's a listening process behind the scene:

* If there is, exit without removing socket api file
* If there is not, then remove file and resume bind.

### DIFF
--- a/main/api.c
+++ b/main/api.c
@@ -206,6 +206,7 @@ int api_socket_start(struct event_base *base) {
 		struct sockaddr_un un;
 		struct sockaddr a;
 	} addr;
+	int ret;
 	int fd;
 
 	fd = socket(AF_UNIX, SOCK_STREAM | SOCK_NONBLOCK | SOCK_CLOEXEC, 0);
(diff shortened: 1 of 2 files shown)


main/main.c
```diff
@@ -183,7 +183,10 @@ int main(int argc, char **argv) {
 	modules_init(ev_base);
 
 	if (api_socket_start(ev_base) < 0) {
-		err = errno;
+		if (errno == EADDRINUSE)
+			ret = 2;
+		else
+			err = errno;
```
rjarry, line 189: instead of a magic number, it can you just check for EADDRINUSE below?

main/main.c
```diff
@@ -211,7 +214,8 @@ int main(int argc, char **argv) {
 		modules_fini(ev_base);
 		event_base_free(ev_base);
 	}
-	unlink(args.api_sock_path);
+	if (ret != 2)
+		unlink(args.api_sock_path);
```
rjarry, line 218:
```c
if (err != EADDRINUSE)
   unlink(args.api_sock_path);
```

main/api.c
```diff
@@ -215,7 +216,24 @@ int api_socket_start(struct event_base *base) {
...
 	memccpy(addr.un.sun_path, path, 0, sizeof(addr.un.sun_path) - 1);
 
-	if (bind(fd, &addr.a, sizeof(addr.un)) < 0) {
+	ret = bind(fd, &addr.a, sizeof(addr.un));
+	if (ret < 0 && errno == EADDRINUSE) {
+		// unix socket file exists, check if there is a process
+		// listening on the other side.
+		ret = connect(fd, &addr.a, sizeof(addr.un));
+		if (ret == 0) {
+			LOG(INFO, "grout already running on API socket %s, exiting", path);
```
rjarry, line 225: Maybe LOG(ERR would be more suitable.

rjarry (changes requested):
Hi Olivier, this looks good. I just have some minor remarks.

main/main.c
ogournet: to let ret = EXIT_FAILURE:

		if (errno != EADDRINUSE)
			err = errno;

main/main.c
ogournet: Right, i will update this way.
Still I don't like it much because errno = EADDRINUSE could be set elsewhere in main()

rjarry: Fair enough. In that case, could you add a dedicated local bool variable to determine whether the socket should be unlinked on exit?

```c
	bool cleanup_sock = true;

	// ...........

	if (api_socket_start(ev_base) < 0) {
		if (errno == EADDRINUSE)
			cleanup_sock = false;
		err = errno;
		goto shutdown;
	}

	// ...........

shutdown:
	// ...........

	if (cleanup_sock)
		unlink(args.api_sock_path);
```

rjarry (approved):
thanks :+1:

Issue #114, opened 2024-12-23
Dangerous uses of rte_pktmbuf_adj


david-marchand:
I noted that in some places of grout, we have constructs like:

```
struct proto_hdr *p;

p = rte_pktmbuf_mtod(m, struct proto_hdr *);
rte_pktmbuf_adj(m, sizeof(*p));
...
if (p->something) {
  struct next_proto_hdr *n;

  n = rte_pktmbuf_mtod(m, struct next_proto_hdr *);
...
```

Calling rte_pktmbuf_adj() means that the data before current offset is in an undefined state.

As long as nothing else touches the mbuf, the current code should be ok.
However,  if a future change introduces something like a prepend and starts writing into this prepended area, then the content of p is overwritten.

In general, calling adj should be understood as "everything before current data offset is lost".

PR #125, opened 2025-01-13
Fix calls to rte_pktmbuf_adj().

david-marchand:
Here is a series trying to fix issues described in #114.

One small concern is wrt the packet tracing feature, where should the data offset be pointing at?


rjarry (approved):
I don't think the data offset matters for packet tracing. We copy header info into a trace item.

PR #126, opened 2025-01-14
Router advertisement & Router sollicitation

christophefontaine:
Add router advertisement support, with basic configuration (lifetime, priority, interval, but no DNS, etc.).
Also answer to router sollicitation messages.


modules/ip6/control/router_advertisement.c
```diff
@@ -0,0 +1,264 @@
...
+
+static void iface_event_handler(iface_event_t event, struct iface *iface) {
+	if (event == IFACE_EVENT_POST_ADD) {
+		ra_conf[iface->id].interval = RA_DEFAULT_INTERVAL;
+		ra_conf[iface->id].lifetime = RA_DEFAULT_LIFETIME;
+		ra_conf[iface->id].preference = RA_DEFAULT_PREFERENCE;
+
+		ra_conf[iface->id].timer = event_new(
+			ev_base, -1, EV_PERSIST | EV_FINALIZE, send_ra_cb, iface
+		);
```
rjarry, line 247: I never recall this libevent api... Is this adding a 1s timer or do you need to "enable" it afterwards?

modules/ip6/control/router_advertisement.c
```diff
@@ -0,0 +1,264 @@
...
+	ip6_output_mbuf_data(m)->nh = ip6_nexthop_lookup(vrf_id, iface_id, &dst);
+	ip = (struct rte_ipv6_hdr *)rte_pktmbuf_append(m, sizeof(*ip));
+	icmp6 = (struct icmp6 *)rte_pktmbuf_append(m, sizeof(*icmp6));
+	icmp6->type = ICMP6_TYPE_ROUTER_ADVERT;
+	icmp6->code = 0;
+	ra = (struct icmp6_router_advert *)rte_pktmbuf_append(m, sizeof(*ra));
+	ra->cur_hoplim = IP6_DEFAULT_HOP_LIMIT; // Default TTL for this network
+	ra->managed_addr = 0; // DHCPv6 is available
+	ra->other_config = 0; // DNS available, ...
+	ra->lifetime = rte_cpu_to_be_16(5); // Not a default router
```
rjarry, line 156: is that in seconds? maybe worth a #define

modules/ip6/control/router_advertisement.c
```diff
@@ -0,0 +1,264 @@
...
+	ra->reachable_time = rte_cpu_to_be_32(0);
+	ra->retrans_timer = rte_cpu_to_be_32(0);
```
rjarry, line 158: Not essential, but RTE_BE32(0) would be better for constants.

modules/ip6/control/router_advertisement.c
```diff
@@ -0,0 +1,264 @@
...
+		resp->ras[n_ras].lifetime = ra_conf[iface_id].lifetime;
+		resp->ras[n_ras].preference = ra_conf[iface_id].preference;
+		n_ras++;
+	}
+
+	*response = resp;
+	return api_out(0, len);
+}
+void ndp_router_sollicit_input_cb(struct rte_mbuf *m) {
+	uint16_t iface_id = mbuf_data(m)->iface->id;
```
rjarry, line 127: Maybe we're missing some RFC checks for router solicitations?

rjarry (changes requested):

modules/ip6/cli/router_advertisement.c
```diff
@@ -0,0 +1,140 @@
...
+		return CMD_ERROR;
+	return CMD_SUCCESS;
+}
+
+static int ctx_init(struct ec_node *root) {
+	int ret;
+
+	ret = CLI_COMMAND(
+		IP6_SHOW_CTX(root),
+		"router-advertisment [IFACE]",
```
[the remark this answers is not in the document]
christophefontaine: even shorter `ra` ? Do you think it may be possible to have "short commands" in the CLI ?
So we could use both `ra` and `router-advertisement` ? (or to even accept partial commands ?)

modules/ip6/control/router_advertisement.c
christophefontaine: I'll double check that.

modules/ip6/control/router_advertisement.c
christophefontaine: Indeed, while the lifetime is in seconds, the other parameters below are in milliseconds...

christophefontaine: The defines are created in the next commit, when the configuration is settable by the user.
```
	ra_conf[req->iface_id].interval = RA_DEFAULT_INTERVAL;
	ra_conf[req->iface_id].lifetime = RA_DEFAULT_LIFETIME;
	ra_conf[req->iface_id].preference = RA_DEFAULT_PREFERENCE;
```
I'll use them in this commit.

modules/ip6/cli/router_advertisement.c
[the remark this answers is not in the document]
rjarry: Yes you can have token alternatives separated by pipes (without spaces)

```c
"ra|router-advert|... [IFACE]"
```

No partial marching though. You need to explicitly list all alternatives to allow libecoli to build a token graph.

modules/ip6/control/router_advert.c
```diff
@@ -0,0 +1,161 @@
...
+		opt->type = ICMP6_OPT_SRC_LLADDR;
+		opt->len = ICMP6_OPT_LEN(sizeof(*opt) + sizeof(*lladdr));
+		lladdr = (struct icmp6_opt_lladdr *)rte_pktmbuf_append(m, sizeof(*lladdr));
+		lladdr->mac = mac;
+		payload_len += sizeof(*opt) + sizeof(*lladdr);
+	}
+
+	ip6_set_fields(ip, payload_len, IPPROTO_ICMPV6, &src, &dst);
+	icmp6->cksum = 0;
+	icmp6->cksum = rte_ipv6_udptcp_cksum(ip, icmp6);
```
rjarry: I forgot to comment on this. Could you follow the same mechanism than arp requests and ndp neighbour solicitations? The control plane sends a nexthop and the mbuf is allocated and built in a datapath node.

modules/ip6/control/router_advert.c
```diff
@@ -0,0 +1,161 @@
...
+			}
+			mbuf_data(m)->iface = iface;
+			build_ra_packet(m, &ip);
+			post_to_stack(ra_output, m);
+		}
+	}
+}
+
+static void ra_init(struct event_base *ev_base) {
+	ra_output = gr_control_input_register_handler("ip6_output", true);
```
rjarry, line 198: following on my previous comment, this should be changed as:

```c
gr_control_input_register_handler("ndp_ra_output", false);
```

rjarry (changes requested):

modules/ip6/control/router_advert.c
christophefontaine: Won't this increase complexity, as we'd need to access to the interface configuration from the datapath to build the router advertisement packet ?
This would makes sense if all exception path are defined as nodes, and they'd be parsed in a dedicated thread in a pipeline mode.

rjarry: The only config you need is the lifetime for the interface id associated with the nexthop. I guess that's OK for now and it will make the graph consistent with arp/ndp output nodes?

README.md
```diff
@@ -237,17 +237,28 @@ image.
...
 
 ```sh
 apt install git gcc make meson ninja-build pkgconf scdoc python3-pyelftools \
         libcmocka-dev libedit-dev libevent-dev libnuma-dev \
-        libsmartcols-dev libarchive-dev libibverbs-dev \
-        clang-format jq curl traceroute graphviz
+        libsmartcols-dev libarchive-dev libibverbs-dev
+```
+
+### Install optional dependencies
```
david-marchand, line 251: This deserves a separate commit for readability.

smoke/ip6_forward_test.sh
```diff
@@ -32,5 +32,5 @@ ip netns exec $p1 ping6 -i0.01 -c3 fd00:ba4:2::2
 ip netns exec $p2 ping6 -i0.01 -c3 fd00:ba4:1::2
 ip netns exec $p1 ping6 -i0.01 -c3 fd00:ba4:1::1
 ip netns exec $p2 ping6 -i0.01 -c3 fd00:ba4:2::1
-ip netns exec $p1 traceroute6 -N1 fd00:ba4:2::2
```
david-marchand, line 35: Idem, it can go to a separate commit especially as you did not describe the reason for this change in the current commitlog.

david-marchand:
Some commitlog has a typo "perdiodic" (commenting here, as I don't see how to comment in gh webui for the specific commit..)

rjarry (approved):
looks good to me after you rebase on the latest master and address @david-marchand's comments :+1:

PR #133, opened 2025-01-17
Split RIB and FIBs between control and data planes

rjarry:


modules/ip/datapath/fib4.c
```diff
@@ -0,0 +1,163 @@
...
+	struct rte_fib *fib;
+	int ret;
+
+	if (nh->ref_count == 0)
+		ABORT("nexthop has ref_count==0: vrf=%u iface=%u " IP4_F,
+		      nh->vrf_id,
+		      nh->iface_id,
+		      &nh->ipv4);
+
+	if (rte_lcore_has_role(rte_lcore_id(), ROLE_NON_EAL))
```
david-marchand, line 109: Not directly related to this patch, but could we add such check in nh[46]_new and friends?

rjarry: Yes, that wouldn't hurt.

modules/ip/control/route.c
```diff
@@ -121,7 +121,7 @@ struct nexthop *ip4_route_lookup(uint16_t vrf_id, ip4_addr_t ip) {
 	return nh_id_to_ptr(nh_id);
 }
 
-struct nexthop *ip4_route_lookup_exact(uint16_t vrf_id, ip4_addr_t ip, uint8_t prefixlen) {
+static struct nexthop *rib4_lookup_exact(uint16_t vrf_id, ip4_addr_t ip, uint8_t prefixlen) {
```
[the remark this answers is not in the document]
rjarry: I'll move the renaming to the other commit.

PR #139, opened 2025-01-24
Glitches

rjarry:


modules/ip6/datapath/ndp_ns_output.c
```diff
@@ -100,6 +100,7 @@ static uint16_t ndp_ns_output_process(
 			memcpy(t, icmp6, trace_len);
 		}
 		nh->last_request = rte_get_tsc_cycles();
+		ip6_output_mbuf_data(mbuf)->iface = iface_from_id(nh->iface_id);
```
christophefontaine, line 108: We should evaluate the creation of scoped multicast nexthop.
We could remove this line, and all checks `rte_ipv6_addr_is_mcast` in ip6_output_process.

rjarry: agreed, but in the meantime can we just fix the issue of NS packets going out in ip6_error?

rjarry: it only occurs since 07ff3ca67b291

christophefontaine (approved):

modules/ip6/datapath/ndp_ns_output.c
christophefontaine: Yes, I created a new issue to track this.

rjarry:
I'll update the commit message to be more explicit about the issue.

christophefontaine (approved):

PR #141, opened 2025-01-25
Always let control plane respond to arp requests & neighbor solicitations

rjarry:


modules/ip6/datapath/ndp_na_output.c
```diff
@@ -0,0 +1,105 @@
...
+	uint16_t payload_len;
+	struct icmp6 *icmp6;
+
+	for (uint16_t i = 0; i < nb_objs; i++) {
+		mbuf = objs[i];
+
+		local = ndp_na_output_mbuf_data(mbuf)->local;
+		remote = ndp_na_output_mbuf_data(mbuf)->remote;
+
+		rte_pktmbuf_reset(mbuf);
```
david-marchand, line 44: I think I already commented on this but I don't remember the reason.
Why call rte_pktmbuf_reset()?

Here, rather than call rte_pktmbuf_reset(), you can simply drop any data by trimming (rte_pktmbuf_trim(m, rte_pktmbuf_pkt_len(m))).
This will preserve current data offset (and this reply packet will likely get symmetrically recapsulated on its output path).

rjarry:
David Marchand, Jan 26, 2025 at 16:54:
> I think I already commented on this but I don't remember the reason.
> Why call rte_pktmbuf_reset()?

In this case, for a bad reason: I wasn't sure which of
rte_pktmbuf_pkt_len() or rte_pktmbuf_data_len() I should use.

I will fix it.

david-marchand:
I did not review in details, but on the principle, this approach looks saner.
You can add my ack on this series.

PR #142, opened 2025-01-26
docs: write man pages in markdown

aharivel:
Move from scdoc to markdown.

Reorder alphabetically grout options.

Add necessary tool to build man pages from markdown.


david-marchand:
This is an open question.

There is a lot of copy/paste happening between cli/main.c and docs/grcli.<whatever_format> (for example).
Would it be possible to have all the cli options description in a single file (iow, in the code)?

rjarry:
The goal in rewriting to markdown is to allow access via github with direct links to options and commands:

https://github.com/DPDK/grout/blob/704e49fa53ad9ae9707a428a836d535c90410ac6/docs/grout.8.md#-p---poll-mode

If we generate the man pages from source code we loose this. We "could" do it the other way around and generate C code from the man page though.

rjarry:
@aharivel could you mention this as motivation for the change from scdoc to markdown?

> The goal in rewriting to markdown is to allow access via github with direct links to options and commands.

rjarry (approved):

PR #143, opened 2025-01-27
smoke: prevent any address to name resolution

rjarry:
We have tests that use publicly routable addresses (16.0.0.0/8). By default, ping and traceroute will try to resolve any IP address to a hostname via the configured name server method for getnameinfo(3).

We certainly don't want that to happen as these programs will at best give up after the default timeout, or at worse print a garbage host name.

Force numeric output only with the -n flag which is supported by all ping, traceroute and ndisc6.

This has the side benefit of making the smoke tests a lot faster:

	before:  ip_forward_test.sh ... OK (01m08s)
	after:   ip_forward_test.sh ... OK (00m06s)

Fixes: 75286b77c3df ("ip: ensure no nexthop is created with invalid iface id")

### DIFF
--- a/smoke/ip6_forward_test.sh
+++ b/smoke/ip6_forward_test.sh
@@ -30,15 +30,15 @@ done
 
 sleep 3  # wait for DAD
 
-ip netns exec $p1 ping6 -i0.01 -c3 fe80::d2f0:cff:feba:a411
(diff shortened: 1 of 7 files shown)


christophefontaine (approved):
RFCs specify multiple ranges of IPs dedicated to documentation and tests. 
Shouldn't we (in the future) use these ranges instead ?

rjarry:
Yes we surely should do that as well.

PR #144, opened 2025-01-28
Smoke ping6

ogournet:


rjarry (changes requested):
Hi Olivier,

thanks for adding a test. The code looks good to me. Can you just add a short explanation of what was happening before your fix?

PR #151, opened 2025-02-07
Srv6

ogournet:
no interface are created, instead a hook is added in ip/ip6_input to visit our node.
need inputs on core grout modification before going further.
    
grcli trace from smoke-test srv6:
    
    --------- 23:06:34.382273411 cpu 1 ---------
    port_rx: port=0 queue=0
    eth_input: d2:ad:ca:fe:b4:10 > d2:f0:0c:ba:a5:10 type=IP(0x0800) iface=p0
    ip_input: 192.168.61.2 > 192.168.60.1 ttl=64 proto=ICMP(1)
    srv6_steer: match=192.168.0.0/16
    ip6_output: fd00:102::1 > fd00:202::2 ttl=64 proto=IPIP(4)
    eth_output: d2:f0:0c:ba:a5:11 > d2:ad:ca:fe:b4:11 type=IPv6(0x86dd) iface=p1
    port_tx: port=1 queue=0
    
    --------- 23:06:34.382273411 cpu 1 ---------
    port_rx: port=1 queue=0
    eth_input: d2:ad:ca:fe:b4:11 > d2:f0:0c:ba:a5:11 type=IPv6(0x86dd) iface=p1
    ip6_input: fd00:102::2 > fd00:202::100 ttl=64 proto=RouteOpts(43)
    srv6_local: action=end.dt4
    ip_input: 192.168.60.1 > 192.168.61.2 ttl=64 proto=ICMP(1)
    ip_forward:
    ip_output: 192.168.60.1 > 192.168.61.2 ttl=63 proto=ICMP(1)
    eth_output: d2:f0:0c:ba:a5:10 > d2:ad:ca:fe:b4:10 type=IP(0x0800) iface=p0
    port_tx: port=0 queue=0


modules/srv6/cli.c
```diff
@@ -0,0 +1,286 @@
...
+		str = ec_strvec_val(v, 0);
+		if (inet_pton(AF_INET6, str, &req->s.nh[i]) != 1) {
+			free(req);
+			return CMD_ERROR;
+		}
+	}
+
+	if ((str = arg_str(p, "DEST6")) != NULL) {
+		if (ip6_net_parse(str, &req->s.dest6, true) < 0)
+			return CMD_ERROR;
```
vjardin, line 43: don't you need a `free(req)` for both error cases DEST6 and DEST4 here after ?

I feel that all the cases:
```
free(req);
return CMD_ERROR;
```

should be replaced by a `goto cmd_error;`

then the end of this function becomes:
```
free(req);
free(resp_ptr);
return CMD_SUCCESS;

cmd_error:
  free(req);
  return CMD_ERROR;
```

modules/srv6/control.c
```diff
@@ -0,0 +1,437 @@
...
+
+static struct api_out srv6_steer_add(const void *request, void ** /*response*/) {
+	const struct gr_srv6_steer_add_req *req = request;
+	struct srv6_steer_data *sd;
+	struct nexthop *nh;
+	int ret;
+
+	// create nexthop
+	// ipv4/v6 address on nexthop are not really useful, only for trace
+	if (req->s.is_dest6) {
```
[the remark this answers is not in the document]
rjarry: That would be nice but it would tie the ip and ip6 modules. Currently the CLI commands are defined in each module and are completely independent.

modules/srv6/gr_srv6_api.h
```diff
@@ -0,0 +1,120 @@
...
+
+#include <gr_api.h>
+#include <gr_bitops.h>
+#include <gr_infra.h>
+#include <gr_macro.h>
+#include <gr_net_types.h>
+
+
+#define GR_IFACE_TYPE_SRV6_STEER 0x04
+#define GR_IFACE_TYPE_SRV6_LOCAL 0x05
```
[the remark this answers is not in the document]
ogournet: I forgot to remove them. There was a late switch from interface + hook in ip6_output to no interface + small patch in nexthop + hook in ip6_input.

The main purpose of this draft PR is to evaluate whether it is acceptable or not, or what improvement can be made on it. We spoke about adding more meaning to nh->family field, but it doesn't look nice to me.

ogournet:
It seems that CI is not running with the latest libecoli, hence the failure.

cli/ecoli.c
```diff
@@ -162,3 +162,20 @@ int arg_eth_addr(const struct ec_pnode *p, const char *id, struct rte_ether_addr
 
 	return 0;
 }
+
+int arg_ip6(const struct ec_pnode *p, const char *id, struct rte_ipv6_addr *addr) {
```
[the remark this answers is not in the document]
ogournet: ip6_parse() parses only ipv6 addr, whereas ip6_net_parse() also parse netmask.
also, ip6_parse takes a ec_pnode and is cli-specialized, when ip6_net_parse() takes a string as input.

I did have lots of ipv6 address (without netmask) before, so I added this helper. There are less now,
so I can remove this helper if you want.

modules/srv6/cli.c
```diff
@@ -0,0 +1,286 @@
...
+	const struct ec_strvec *v;
+	void *resp_ptr = NULL;
+
+	n = ec_pnode_find(p, "behavior");
+	if (n == NULL || (n = ec_pnode_next(n)) == NULL)
+		return CMD_ERROR;
+	v = ec_pnode_get_strvec(n);
+	snprintf(req.l.behavior, sizeof (req.l.behavior), "%s", ec_strvec_val(v, 0));
+
+	if (arg_ip6(p, "SID", &req.l.lsid) < 0)
```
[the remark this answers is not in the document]
ogournet: to be complete, we could add:
arg_ip4(), arg_ip4_net(), arg_ip6(), arg_ip6_net()
in cli/gr_cli.h

and arg_ipX_net() will make use of ipX_net_parse()

modules/srv6/control.c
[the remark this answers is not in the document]
rjarry: It silently ignores one of the two redundant branches in the command tree :)

rjarry:
> It seems that CI is not running with the latest libecoli, hence the failure.

It should be running with the v0.4.0 tag.

modules/infra/control/gr_nh_control.h
```diff
@@ -83,6 +85,9 @@ struct __rte_cache_aligned nexthop {
 	struct rte_mbuf *held_pkts_head;
 	struct rte_mbuf *held_pkts_tail;
 
+	// custom node to visit on packet input
+	rte_edge_t input_node;
+
```
rjarry, line 90: Instead of this mechanism. I would prefer either a custom family AF_SRV6 or a flag attached to nexthops (similar to GR_NH_F_LOCAL) that indicates special handling in ip6_input.

ogournet:
> > It seems that CI is not running with the latest libecoli, hence the failure.
> 
> It should be running with the v0.4.0 tag.

oh right, thx. found the problem.

modules/infra/control/gr_nh_control.h
rjarry: Hmm, I need to think about it some more.

rjarry: I am going through RFCs to get my head around SRv6. But in the meantime, it seems that it this feature is strongly tied to IPv6 itself. It may not make sense to have it as a separate module. I am not 100% decided on that matter.

That said, the "segment routing" concept seems to apply to both IPv4 and IPv6 (and MPLS as well).

Instead of a custom "family" on nexthop objects which (as you pointed out) goes against the current constructs of nexthop pools per family (AF_INET and AF_INET6 currently), we could introduce a GR_NH_F_SEGMENT flag which is generic so it can be defined in the infra module where gr_nexthop is defined.

After that you could have hook points in ip_input and ip6_input where the GR_NH_F_SEGMENT is checked on the resolved nexthop after fib{,6}_lookup. Either it would steer the packets in ip{,6}_segment_steer nodes defined in the ip and ip6 modules. Or these nodes could be defined in a separate module.

What do you think?

rjarry: After some brainstorming, here's what we came up with @christophefontaine 

image

Basically, we need to define dynamic hook points in `ip{,6}_input` based on `nexthop->family` only when `nh->flags & GR_NH_F_LOCAL`. And with a similar mechanism in `ip{,6}_output`.

I will propose some changes in a separate pull request.

ogournet: That looks a little complex, but you know better than me what you're going to accept in modules/ip6.

On srv6 endpoint, the use of F_LOCAL flag means that localSID in going into an interface. There is none in the current draft, but this could be done.
One useful feature, but not mandatory, is to be able to ping localSID.

rjarry: #158 should cover the bottom diagram (SRv6 endpoint). I am preparing another patch to allow redirecting to SRv6 output based on a nexthop type but this may take some more time.

rjarry: Along with #170 this should cover what you need.

That means you'll need to change your implementation to follow the new paradigms.

modules/infra/control/nexthop.c
```diff
@@ -142,8 +143,9 @@ void nexthop_decref(struct nexthop *nh) {
 	}
 }
 
-void nexthop_incref(struct nexthop *nh) {
+struct nexthop *nexthop_incref(struct nexthop *nh) {
 	nh->ref_count++;
+	return nh;
```
rjarry, line 148: This does not seem used anywhere. Could you remove it?

modules/srv6/gr_srv6.h
```diff
@@ -0,0 +1,63 @@
...
+struct ip6_sr_rthdr {
+	uint8_t nxt; /* next header */
+	uint8_t len; /* length in units of 8 octets */
+	uint8_t type; /* routing type */
+	uint8_t segleft; /* segments left */
+	uint8_t last_entry;
+	uint8_t flags;
+	rte_be16_t tag;
+	struct rte_ipv6_addr segments[];
+} __rte_packed;
```
rjarry, line 28: Could you reuse the routing header extension that is already defined in dpdk?

https://github.com/DPDK/dpdk/blob/fd233ad17e5ffa42d50f1625165a7fce3f1cbc5f/lib/net/rte_ip6.h#L511-L526

modules/srv6/gr_srv6.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include <rte_ip6.h>
+
+#include <endian.h>
+#include <netinet/ip6.h>
+
+//
+// srv6 pkt header definition
+//
+#define IP6_ROUTING_HEADER_TYPE_SRV6 4
```
rjarry, line 17: I am not sure which constant this is referring to, but there already is another one in DPDK with the same value:

https://github.com/DPDK/dpdk/blob/fd233ad17e5ffa42d50f1625165a7fce3f1cbc5f/lib/net/rte_ip6.h#L505-L506

modules/srv6/gr_srv6.h
```diff
@@ -0,0 +1,63 @@
...
+//
+struct srv6_localsid_data {
+	enum gr_srv6_behavior behavior;
+	uint16_t out_vrf_id;
+	uint8_t flags;
+};
+
+struct srv6_localsid_data *srv6_localsid_get(const struct rte_ipv6_addr *lsid, uint16_t vrf_id);
+
+extern rte_edge_t srv6_local_edge;
```
rjarry, line 47: Can you use `ip6_input_local_add_proto(IPPROTO_ROUTING, "srv6_local")` instead?

modules/srv6/gr_srv6.h
```diff
@@ -0,0 +1,63 @@
...
+struct srv6_steer_data {
+	struct nexthop *gr_nh;
+	uint16_t n_nh;
+	struct rte_ipv6_addr nh[];
+};
+
+struct srv6_steer_data *srv6_steer_get(const struct nexthop *nh);
+
+extern rte_edge_t srv6_steer_v4_edge;
+extern rte_edge_t srv6_steer_v6_edge;
```
rjarry, line 61: These two should be replaceable with a new `GR_NH_SRV6_SID` nexthop type along with:

```c
ip_output_register_nexthop_type(GR_NH_SRV6_SID, "srv6_steer");
ip6_output_register_nexthop_type(GR_NH_SRV6_SID, "srv6_steer");
```

I didn't look in detail what that would imply on your design. But I think it can work.

modules/srv6/gr_srv6_api.h
```diff
@@ -0,0 +1,103 @@
...
+//
+enum gr_srv6_behavior {
+	SR_BEHAVIOR_END = 0x0001,
+	SR_BEHAVIOR_END_T = 0x0009,
+	SR_BEHAVIOR_END_DT6 = 0x0012,
+	SR_BEHAVIOR_END_DT4 = 0x0013,
+	SR_BEHAVIOR_END_DT46 = 0x0014,
+
+	SR_BEHAVIOR_MAX,
+};
```
rjarry, line 26: Can you rewrite this as:

```c
typdef enum : uint16_t {
	SR_BEHAVIOR_END = 0x0001,
	SR_BEHAVIOR_END_T = 0x0009,
	SR_BEHAVIOR_END_DT6 = 0x0012,
	SR_BEHAVIOR_END_DT4 = 0x0013,
	SR_BEHAVIOR_END_DT46 = 0x0014,
	SR_BEHAVIOR_COUNT
} gr_sr6_behavior_t;
```

rjarry (changes requested):

modules/srv6/gr_srv6.h
ogournet: that's not enough, SRH (proto = IPPROTO_ROUTING with type=4) is not always present in packets "srv6_local".
I prefer a single entry point that cover all cases.

modules/infra/control/gr_nh_control.h
ogournet: It's done.
Srv6 endpoint also hook from ip6_output, because it does feel weird hooking from ip6_local.
BUT following rfc8986 4.1.1 Upper-Layer Header, the packet may re-enter ip6_input_local from sr6_local if proto is icmpv6, thus allowing ping of localSID.

rjarry:
@ogournet do you think you could split the `srv6: initial import` commit into multiple ones with some descriptions to help understanding the thought processes involved?

It would be great to have some references to RFC sections and also if needed some ASCII drawings to explain how everything articulates.

Here are some examples of what I am thinking about:

https://github.com/DPDK/grout/commit/0fc1cd222b8ec3b125cdfb168a09a240cc0debf0
https://github.com/DPDK/grout/commit/fb72a4d6f57c251ecdd43001033ff68dc1698b99
https://github.com/DPDK/grout/commit/7f068c47af42e2b5b663f7a22aeed2eb7e463455
https://github.com/DPDK/grout/commit/d50fc10a411c125328cf79a495a763568cbaa1c4

modules/srv6/datapath_headend.c
```diff
@@ -0,0 +1,203 @@
...
+	int proto, reduc;
+
+	for (uint16_t i = 0; i < nb_objs; i++) {
+		m = objs[i];
+
+		if (gr_mbuf_is_traced(m))
+			t = gr_mbuf_trace_add(m, node, sizeof(*t));
+
+		// dpdk mbuf packet_type MUST be correctly filled
+		if (m->packet_type & RTE_PTYPE_L3_IPV4) {
```
rjarry, line 78: This is a bit fragile. The mbuf packet_type field is dependent on DPDK driver implementations. Maybe you should enforce `m->packet_type |= RTE_PTYPE_L3_IPV4` and `m->packet_type |= RTE_PTYPE_L3_IPV6` from `ip_output` and `ip6_output` to avoid any surprises?

modules/srv6/datapath_headend.c
```diff
@@ -0,0 +1,203 @@
...
+			hdrlen += sizeof(*srh) + (d->n_seglist * sizeof(d->seglist[0]));
+
+		outer_ip6 = (struct rte_ipv6_hdr *)rte_pktmbuf_prepend(m, hdrlen);
+		if (unlikely(outer_ip6 == NULL)) {
+			edge = NO_HEADROOM;
+			goto next;
+		}
+		outer_ip6->vtc_flow = RTE_BE32(0x60000000);
+		outer_ip6->dst_addr = d->seglist[0];
+		outer_ip6->hop_limits = 64;
```
rjarry, line 127: can't you reuse `ip6_set_fields` in this case?

modules/srv6/gr_srv6.h
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
rjarry, line 1: If this is a non exported header, can you rename it to `srv6.h` without the `gr_` prefix and include it with double quotes?

modules/srv6/gr_srv6_api.h
```diff
@@ -0,0 +1,183 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
rjarry, line 1: Rename to `gr_srv6.h` to match other API headers.

modules/srv6/gr_srv6.h
```diff
@@ -0,0 +1,43 @@
...
+
+#ifndef _GR_DATAPATH_SRV6
+#define _GR_DATAPATH_SRV6
+
+#include "gr_srv6_api.h"
+
+#include <rte_ip6.h>
+
+#include <endian.h>
+#include <netinet/ip6.h>
```
rjarry, line 12: Are these includes needed for symbols defined in this header?

modules/srv6/gr_srv6_api.h
```diff
@@ -0,0 +1,183 @@
...
+
+	SR_H_ENCAPS_MAX,
+} gr_srv6_encap_behavior_t;
+
+struct gr_srv6_policy {
+	struct rte_ipv6_addr bsid;
+	uint16_t weight;
+	gr_srv6_encap_behavior_t encap_behavior;
+	uint8_t n_seglist;
+	struct rte_ipv6_addr seglist[/* n_nh */];
```
rjarry, line 56: nit in the comment: `s/n_nh/n_seglist/`

modules/srv6/gr_srv6_api.h
```diff
@@ -0,0 +1,183 @@
...
+#define GR_SRV6_POLICY_ADD REQUEST_TYPE(GR_SRV6_MODULE, 0x0001)
+
+struct gr_srv6_policy_add_req {
+	struct gr_srv6_policy p;
+};
+
+#define GR_SRV6_POLICY_DEL REQUEST_TYPE(GR_SRV6_MODULE, 0x0002)
+
+struct gr_srv6_policy_del_req {
+	struct rte_ipv6_addr bsid;
```
rjarry, line 68: For my culture, what is the `b` of `bsid` standing for?

smoke/srv6_test.sh
```diff
@@ -0,0 +1,295 @@
+#!/bin/bash
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2025 Olivier Gournet
+
+
```
rjarry, line 5: nit: only one empty line here.

smoke/srv6_test.sh
```diff
@@ -0,0 +1,295 @@
...
+    action=run
+    . $(dirname $0)/_init.sh
+
+else
+    # run from command line
+    fail() { exit 1; }
+    export PATH=$PATH:$(dirname $0)/../build
+    run_id=p
+    set -x
+fi
```
rjarry, line 19: Can you explain what this is for?

Also, please indent with tabs in shell scripts.

smoke/srv6_test.sh
```diff
@@ -0,0 +1,295 @@
...
+    ip -n $p1 link set $p1 up
+    ip -n $p1 addr add fd00:102::2/64 dev $p1
+    ip -n $p1 addr add 192.168.62.2/24 dev $p1
+}
+
+send_py_pkt() {
+    port=$1
+    script_head="
+#!/usr/bin/env python3
+from scapy.all import *
```
rjarry, line 93: Make sure to add python3-scapy to the dev dependencies in `README.md` and in `.github/workflows/*`.

smoke/srv6_test.sh
```diff
@@ -0,0 +1,295 @@
...
+"
+    ip netns exec $port python3 -c "$script_head $2 $script_tail"
+}
+send_py_pkt_dis() {
+    return
+}
+
+datapath_local() {
+    grcli show interface name p0 > /dev/null 2>&1
+    if [ $? -ne 0 ]; then
```
rjarry, line 106: Can you add a comment explaining what is this for?

Also, I prefer the shorter and more robust form:

```sh
if ! grcli show interface name p0 >/dev/null 2>&1; then
```

modules/srv6/datapath_local.c
```diff
@@ -0,0 +1,409 @@
...
+	[SR_BEHAVIOR_END] = "end",
+	[SR_BEHAVIOR_END_T] = "end.t",
+	[SR_BEHAVIOR_END_DT6] = "end.dt6",
+	[SR_BEHAVIOR_END_DT4] = "end.dt4",
+	[SR_BEHAVIOR_END_DT46] = "end.dt46",
+};
+
+static const char *behavior_to_str(gr_srv6_behavior_t b) {
+	return behavior_str[b];
+}
```
rjarry, line 47: probably not needed to have a function here. You can get the behavior name directly from the `behavior_str` array in `trace_srv6_format`.

modules/srv6/datapath_local.c
```diff
@@ -0,0 +1,409 @@
...
+};
+
+static int trace_srv6_format(char *buf, size_t len, const void *data, size_t /*data_len*/) {
+	const struct trace_srv6_data *t = data;
+	if (t->out_vrf_id < MAX_VRFS)
+		return snprintf(
+			buf,
+			len,
+			"action=%s segleft=%d out_vrf=%d",
+			behavior_to_str(t->behavior),
```
rjarry, line 62: `s/behavior_to_str(t->behavior)/behavior_str[t->behavior]/`

modules/srv6/datapath_local.c
```diff
@@ -0,0 +1,409 @@
...
+		return UNEXPECTED_UPPER;
+
+	// prepend ipv6 header without any ext, before entering ip6_input_local again.
+	ip6 = (struct rte_ipv6_hdr *)rte_pktmbuf_prepend(m, sizeof(*ip6));
+	ip6->proto = d->proto;
+	ip6->src_addr = d->src;
+	ip6->dst_addr = d->dst;
+	ip6->payload_len = rte_cpu_to_be_16(d->len);
+	ip6->hop_limits = d->hop_limit;
+	ip6->vtc_flow = RTE_BE32(0x60000000);
```
rjarry, line 110: maybe use `ip6_set_fields` here?

modules/srv6/control.c
```diff
@@ -0,0 +1,682 @@
...
+	struct srv6_policy_data ***d_listlist = NULL;
+	struct srv6_policy_data **d_list;
+	const struct nexthop **nh_list = NULL;
+	struct gr_srv6_steer_list_resp *resp;
+	const struct nexthop *nh;
+	struct gr_srv6_steer_entry *e;
+	void *data_ptr, *ptr;
+	const void *key_ptr;
+	uint32_t iter, i;
+	ssize_t j, len = sizeof(*resp);
```
rjarry, line 375: reverse xmas tree please

modules/srv6/control.c
```diff
@@ -0,0 +1,682 @@
...
+	gr_register_api_handler(&srv6_policy_add_handler);
+	gr_register_api_handler(&srv6_policy_del_handler);
+	gr_register_api_handler(&srv6_policy_get_handler);
+	gr_register_api_handler(&srv6_policy_list_handler);
+	gr_register_api_handler(&srv6_steer_add_handler);
+	gr_register_api_handler(&srv6_steer_del_handler);
+	gr_register_api_handler(&srv6_steer_list_handler);
+	gr_register_api_handler(&srv6_localsid_add_handler);
+	gr_register_api_handler(&srv6_localsid_del_handler);
+	gr_register_api_handler(&srv6_localsid_list_handler);
```
rjarry, line 680: Maybe it would make sense to split this file into steer, policy and localsid handlers?

rjarry (changes requested):
Hi Olivier,

this is looking good overall!

I have some remarks on the form but nothing critical.

Thank you again for your work!

rjarry:
Do you think you could split the `srv6: add srv6 module` commit into multiple ones to make it easier to review?

modules/srv6/gr_srv6_api.h
ogournet: [Binding SID](https://www.rfc-editor.org/rfc/rfc8402#section-5).

smoke/srv6_test.sh
ogournet: This was my 'unit test suite', written using part of smoke-test framework.
But it has nothing to do here, will remove it.

modules/srv6/control.c
ogournet: I will split them into control_local.c and control_headend.c, to match datapath source files.

rjarry:
> I prefer to keep it as a single commit.
> Every modification outside of modules/srv6 has its own commit.

Ok, fair enough. For future changes, it would be better to try and split it into incremental changes, even if it makes the features incomplete or not fully usable.

I see that you removed the scapy based tests. Why not keep them? They seemed like useful additions, especially for exotic protocols such as segment routing. For instance, I don't how the linux implementation of srv6 is compliant with the RFCs.

ogournet:
> I see that you removed the scapy based tests. Why not keep them?

It was not automatically run by smoke-tests, and didn't have the quality required for this PR.

rjarry (approved):
Looks good to me :+1:

rjarry:
Thank you so much for your contribution @ogournet !

PR #153, opened 2025-02-12
L1 xconnect

christophefontaine:
Introduce a l2 cross connect mode, which aims to replace testpmd io forwarding mode.

With this PR, you can switch to a "L2 XCONNECT" mode, to forward "as-is" all packets from 1 interface to another.
In order to create a full cross connect, the api must be called twice.

```
grout# set interface port dpdk-p0 mode xconnect dpdk-p1
grout# set interface port dpdk-p1 mode xconnect dpdk-p0
```

An interface can be cross connected with itself:
```
grout# set interface port dpdk-p0 mode xconnect dpdk-p0
```

Additional CLI commands are added as well:
```
grout# show interface all
NAME      ID  FLAGS       MODE      DOMAIN  TYPE      INFO
dpdk-p0   1   up running  XCONNECT  1       port      devargs=0000:05:00.0 mac=3c:fd:fe:a0:a4:c4
dpdk-p1   3   up running  L3    0       port      devargs=0000:05:00.1 mac=3c:fd:fe:a0:a4:c5
```

```
grout# show interface mode
NAME     MODE      PEER
dpdk-p0  XCONNECT  dpdk-p0
dpdk-p1  L3
```


modules/infra/cli/iface.c
```diff
@@ -262,11 +263,14 @@ static cmd_status_t iface_list(const struct gr_api_client *c, const struct ec_pn
 		char buf[BUFSIZ];
 		size_t n = 0;
 
+		// mode
+		scols_line_set_data(line, 0, iface->mode == GR_IFACE_MODE_L3 ? "*" : "!");
```
rjarry, line 267: I would prefer if this was moved as a MODE column before DOMAIN:

```
grout# show interface all
NAME      ID  FLAGS       MODE  DOMAIN  TYPE      INFO
dpdk-p0   1   up running  L2XC  3       port      devargs=0000:05:00.0 mac=3c:fd:fe:a0:a4:c4
gr-loop0  2   up running  L3    0       loopback
dpdk-p1   3   up running  L3    1       port      devargs=0000:05:00.1 mac=3c:fd:fe:a0:a4:c5
```

modules/infra/api/gr_infra.h
```diff
@@ -39,14 +39,23 @@
 
 #define GR_VRF_ID_ALL UINT16_MAX
 
+enum gr_iface_mode {
+	GR_IFACE_MODE_L3,
```
rjarry, line 43: Maybe set this to explicit = 0?

modules/infra/api/gr_infra.h
```diff
@@ -39,14 +39,23 @@
...
 	uint16_t type; // Interface type. Uses values from GR_IFACE_TYPE_*.
 	uint16_t flags; // Interface flags. Bit mask of GR_IFACE_F_*.
 	uint16_t state; // Interface state. Bit mask of GR_IFACE_S_*.
 	uint16_t mtu; // Maximum transmission unit size (incl. headers).
-	uint16_t vrf_id; // L3 addressing and routing domain
+	union {
+		uint16_t vrf_id; // L3 addressing and routing domain
+		uint16_t bridge_domain; // L2 xconnect peer interface id
+	};
+	enum gr_iface_mode mode;
```
rjarry, line 58: If you use enums in structures shared over the API, could you either pack them (or at least store them in well-known integer types)?

modules/infra/api/gr_infra.h
```diff
@@ -39,14 +39,23 @@
 
 #define GR_VRF_ID_ALL UINT16_MAX
 
+enum gr_iface_mode {
```
rjarry, line 42: I prefer `typedef enum { ... } __attribute__((packed)) gr_iface_mode_t;`

modules/infra/api/gr_infra.h
rjarry: Also, move the mode field before vrf_id, domain_id?

modules/infra/api/gr_infra.h
```diff
@@ -39,14 +39,23 @@
...
 struct gr_iface {
 	uint16_t id; // Interface unique index.
 	uint16_t type; // Interface type. Uses values from GR_IFACE_TYPE_*.
 	uint16_t flags; // Interface flags. Bit mask of GR_IFACE_F_*.
 	uint16_t state; // Interface state. Bit mask of GR_IFACE_S_*.
 	uint16_t mtu; // Maximum transmission unit size (incl. headers).
-	uint16_t vrf_id; // L3 addressing and routing domain
+	union {
+		uint16_t vrf_id; // L3 addressing and routing domain
+		uint16_t bridge_domain; // L2 xconnect peer interface id
```
rjarry, line 56: We don't have any bridge support yet. Maybe rename it to l2_domain_id? or simply domain_id?

modules/infra/datapath/rx.c
```diff
@@ -80,7 +81,18 @@ rx_process(struct rte_graph *graph, struct rte_node *node, void ** /*objs*/, uin
...
+		switch (iface->mode) {
+		case GR_IFACE_MODE_L3:
+			edge = ETH_IN;
+			break;
+		case GR_IFACE_MODE_L2_XCONNECT:
+			edge = L2XCONNECT;
+			break;
+		default:
+			edge = IFACE_MODE_UNKNOWN;
+			break;
```
rjarry, line 94: No need for dynamic registration? You are effectively adding a dependency from infra to the l2 module.

modules/l2/cli.c
```diff
@@ -0,0 +1,167 @@
...
+
+	return CMD_SUCCESS;
+}
+
+static int ctx_init(struct ec_node *root) {
+	int ret;
+
+	ret = CLI_COMMAND(
+		CLI_CONTEXT(root, CTX_SET, CTX_ARG("l2", "Enable L2 mode")),
+		"xconnect IFACE peer PEER",
```
rjarry, line 116: I am not fond of configuring this independently of interfaces. Since this is changing the mode of an interface regardless of its type (the mode is a generic field), it should be done in the interface context.

modules/l2/cli.c
```diff
@@ -0,0 +1,167 @@
...
+	if (gr_api_client_send_recv(c, GR_INFRA_IFACE_LIST, sizeof(req), &req, &resp_ptr) < 0) {
+		scols_unref_table(table);
+		return CMD_ERROR;
+	}
+	resp = resp_ptr;
+	qsort(resp->ifaces, resp->n_ifaces, sizeof(*resp->ifaces), iface_order);
+
+	scols_table_new_column(table, "NAME", 0, 0);
+	scols_table_new_column(table, "MODE", 0, 0);
+	scols_table_new_column(table, "PEER", 0, 0);
```
rjarry, line 79: No data is printed for the PEER column

modules/l2/cli.c
```diff
@@ -0,0 +1,167 @@
...
+		scols_line_set_data(line, 0, iface->name);
+		switch (iface->mode) {
+		case GR_IFACE_MODE_L3:
+			scols_line_set_data(line, 1, "L3");
+			break;
+		case GR_IFACE_MODE_L2_XCONNECT:
+			if (iface_from_id(c, iface->bridge_domain, &peer) < 0)
+				scols_line_sprintf(line, 1, "L2 XCONNECT -> BROKEN");
+			else
+				scols_line_sprintf(line, 1, "L2 XCONNECT -> %s", peer.name);
```
rjarry, line 99: You need to print two columns here.

rjarry (changes requested):

modules/infra/datapath/rx.c
christophefontaine: I'm wondering if I shouldn't put the cross connect between eth_in and eth_out.
This would enable cross connect to work between interfaces and not only ports, and would be more replicable for a (potential) bridge support.

modules/l2/cli.c
christophefontaine: yes, will fix it.

modules/infra/cli/iface.c
christophefontaine: ack

vjardin:
In order to avoid some confusions, when it is about xconnect some interfaces, it is a PHY xconnect or PHY port xconnect. It was the intend of testpmd.

A L2 xconnect, if used, could be applied for VLAN/VXLAN/or any ethernet bridging. I guess, it is not the intend. Am I wrong ?

christophefontaine:
@vjardin you're right, that's why I was wondering if it wouldn't be better to plug it on the eth_input/ouput nodes instead of the port_rx/port_tx nodes.
This would enable a true xconnect, and not a phy xconnect.
The 2 use cases are valid, and maybe I should rename the mode into `IFACE_MODE_PHY_XCONNECT` instead of L2_XCONNECT, so we can have a "real" l2 xconnect later.

Or would it be better to skip the phy_xconnect and implement l2_xconnect now ?

vjardin:
> Or would it be better to skip the phy_xconnect and implement l2_xconnect now ?

I foresee many benefits for testing for a phy_xconnect while I do not foresee using a generic l2 xconnect. L2 xconnect : stack wide, it is a simplified bridge ; performance wide it is does not help for any benchmarks. For the use cases, testpmd-like is the only case I would foresee (PHY to PHY, virtio to virtio, PHY to vhost, etc.)

Should I add a wish : it would be nice to have a rehash on tx so any set of  flows from a rx queue can be spread to a wide set of tx queue based on the mbuf' ingress RSS, flow hashes hard coded, flow hashes from any tbd models we could think adding (dynamic .so, userland eBPF, etc...).

Then, on tx it can combined with any of the DPDK tx qos (hardware schedulers, software ones, etc...) ; it would be a perfect benchmarking tools for DPDK and HW TX schedulers.

christophefontaine:
Ack. I'll continue on that path, and will create an issue to track your request.

modules/infra/datapath/rx.c
```diff
@@ -51,8 +59,11 @@ rx_process(struct rte_graph *graph, struct rte_node *node, void ** /*objs*/, uin
 		rx = rte_eth_rx_burst(
 			q.port_id, q.rxq_id, (struct rte_mbuf **)&node->objs[count], ctx->burst_size
 		);
+		if (rx == 0)
```
vjardin, line 62: hints: unlikely() ?

modules/infra/datapath/rx.c
```diff
@@ -51,8 +59,11 @@ rx_process(struct rte_graph *graph, struct rte_node *node, void ** /*objs*/, uin
...
+			continue;
+
 		iface = port_get_iface(q.port_id);
-		if (rx > 0 && iface == NULL) {
+		if (iface == NULL) {
```
vjardin, line 66: hints: unlikely() ?

vjardin (changes requested):

modules/infra/api/gr_infra.h
```diff
@@ -39,19 +39,31 @@
 
 #define GR_VRF_ID_ALL UINT16_MAX
 
+typedef enum gr_iface_mode {
+	GR_IFACE_MODE_L3 = 0,
+	GR_IFACE_MODE_XCONNECT,
+	GR_IFACE_MODE_COUNT,
+} __attribute__((packed)) gr_iface_mode_t;
```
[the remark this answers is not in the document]
christophefontaine: thanks ! That's exactly what I wished to do 👍 
Will fix that.

christophefontaine:
Depends on #159

modules/meson.build
```diff
@@ -5,4 +5,5 @@ subdir('infra')
 subdir('ip')
 subdir('ip6')
 subdir('ipip')
+subdir('l2')
```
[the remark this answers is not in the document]
christophefontaine: merged into the infra module.

modules/infra/control/gr_iface.h
```diff
@@ -31,9 +31,7 @@ typedef int (*iface_init_t)(struct iface *, const void *api_info);
 typedef int (*iface_reconfig_t)(
 	struct iface *,
 	uint64_t set_attrs,
-	uint16_t flags,
-	uint16_t mtu,
-	uint16_t vrf_id,
+	const struct gr_iface,
```
rjarry, line 38: Can you pass a `const struct gr_iface *` to be consistent with other arguments?

modules/infra/api/iface.c
```diff
@@ -30,14 +30,7 @@ static struct api_out iface_add(const void *request, void **response) {
...
 
-	iface = iface_create(
-		req->iface.type,
-		req->iface.flags,
-		req->iface.mtu,
-		req->iface.vrf_id,
-		req->iface.name,
-		req->iface.info
-	);
+	iface = iface_create(req->iface, req->iface.info);
```
rjarry, line 34:
```c
iface = iface_create(&req->iface, req->iface.info);
```

modules/infra/api/iface.c
```diff
@@ -112,15 +105,7 @@ static struct api_out iface_set(const void *request, void ** /*response*/) {
...
-	ret = iface_reconfig(
-		req->iface.id,
-		req->set_attrs,
-		req->iface.flags,
-		req->iface.mtu,
-		req->iface.vrf_id,
-		req->iface.name,
-		req->iface.info
-	);
+	ret = iface_reconfig(req->iface.id, req->set_attrs, req->iface, NULL);
```
rjarry, line 109:
```c
iface = iface_reconfig(req->iface.id, req->set_attrs, &req->iface, NULL);
```

modules/infra/control/gr_iface.h
```diff
@@ -57,21 +55,11 @@ struct iface_type {
...
 struct iface_type *iface_type_get(uint16_t type_id);
-struct iface *iface_create(
-	uint16_t type_id,
-	uint16_t flags,
-	uint16_t mtu,
-	uint16_t vrf_id,
-	const char *name,
-	const void *api_info
-);
+struct iface *iface_create(const struct gr_iface conf, const void *api_info);
```
rjarry, line 62:
```c
struct iface *iface_create(const struct gr_iface *, const void *api_info);
```

modules/infra/control/gr_iface.h
```diff
@@ -57,21 +55,11 @@ struct iface_type {
...
 int iface_reconfig(
 	uint16_t ifid,
 	uint64_t set_attrs,
-	uint16_t flags,
-	uint16_t mtu,
-	uint16_t vrf_id,
-	const char *name,
+	const struct gr_iface conf,
```
rjarry, line 66:
```c
const struct gr_iface *,
```

modules/infra/control/gr_port.h
```diff
@@ -47,9 +47,7 @@ uint32_t port_get_rxq_buffer_us(uint16_t port_id, uint16_t rxq_id);
 int iface_port_reconfig(
 	struct iface *iface,
 	uint64_t set_attrs,
-	uint16_t flags,
-	uint16_t mtu,
-	uint16_t vrf_id,
+	const struct gr_iface conf,
```
rjarry, line 50:
```c
const struct gr_iface *,
```

modules/infra/control/worker.c
```diff
@@ -284,14 +284,10 @@ int worker_rxq_assign(uint16_t port_id, uint16_t rxq_id, uint16_t cpu_id) {
...
-				GR_PORT_SET_N_TXQS,
-				iface->flags,
-				iface->mtu,
-				iface->vrf_id,
-				&p
-			);
+			struct gr_iface conf = {
+				.flags = iface->flags, .mtu = iface->mtu, .vrf_id = iface->vrf_id
+			};
+			ret = iface_port_reconfig(iface, GR_PORT_SET_N_TXQS, conf, &p);
```
rjarry, line 293: &iface

modules/infra/control/vlan.c
```diff
@@ -50,9 +50,7 @@ static int get_parent_port_id(uint16_t parent_id, uint16_t *port_id) {
 static int iface_vlan_reconfig(
 	struct iface *iface,
 	uint64_t set_attrs,
-	uint16_t flags,
-	uint16_t mtu,
-	uint16_t vrf_id,
+	const struct gr_iface conf,
```
rjarry, line 53:
```c
const struct gr_iface *,
```

modules/ipip/control.c
```diff
@@ -47,27 +47,25 @@ struct iface *ipip_get_iface(ip4_addr_t local, ip4_addr_t remote, uint16_t vrf_i
 static int iface_ipip_reconfig(
 	struct iface *iface,
 	uint64_t set_attrs,
-	uint16_t flags,
-	uint16_t mtu,
-	uint16_t vrf_id,
+	const struct gr_iface conf,
```
rjarry, line 50:
```c
const struct gr_iface *,
```

rjarry (changes requested):

modules/infra/control/gr_iface.h
christophefontaine: done.

docs/graph.svg
rjarry, line 1: I don't see any change in the diagram. Can this be removed from the patch?

modules/infra/api/gr_infra.h
```diff
@@ -33,25 +33,39 @@
...
 #define GR_IFACE_ID_UNDEF 0
 
 #define GR_VRF_ID_ALL UINT16_MAX
 
+typedef enum gr_iface_mode {
+	GR_IFACE_MODE_L3 = 0,
+	GR_IFACE_MODE_XCONNECT,
+	GR_IFACE_MODE_COUNT
+} __attribute__((mode(QI))) gr_iface_mode_t;
+
```
rjarry, line 49: I am starting to wonder if we could change the gr_iface_type_t to be mode(QI) too so that we don't have a 1 byte hole in the structure.

gr_iface_type_t does not need to be on 16 bit anyway and that would also remove the need of defining a GR_IFACE_TYPE_MAX to limit the values.

What do you think?

modules/infra/api/gr_infra.h
```diff
@@ -33,25 +33,39 @@
...
+	union {
+		uint16_t vrf_id; // L3 addressing and routing domain
+		uint16_t domain_id; // L2 xconnect peer interface id
+	};
 #define GR_IFACE_NAME_SIZE 64
 	char name[GR_IFACE_NAME_SIZE]; // Interface name (utf-8 encoded, nul terminated).
 	uint8_t info[256]; // Type specific interface info.
 };
 
+void register_interface_mode(gr_iface_mode_t mode, const char *next_node);
```
rjarry, line 67: This should go in the datapath headers. Not to be exposed in the API for clients.

modules/infra/cli/iface.c
```diff
@@ -283,21 +284,34 @@ static cmd_status_t iface_list(const struct gr_api_client *c, const struct ec_pn
 			SAFE_BUF(snprintf, sizeof(buf), " tracing");
 		scols_line_set_data(line, 2, buf);
 
+		// mode
+		switch (iface->mode) {
+		case GR_IFACE_MODE_L3:
+			scols_line_set_data(line, 3, "L3");
+			break;
+		case GR_IFACE_MODE_XCONNECT:
+			scols_line_set_data(line, 3, "XCONNECT");
```
rjarry, line 293: s/XCONNECT/XC/ (or L1XC) ?

modules/infra/cli/iface.c
```diff
@@ -283,21 +284,34 @@ static cmd_status_t iface_list(const struct gr_api_client *c, const struct ec_pn
...
+			break;
+		case GR_IFACE_MODE_COUNT:
+			scols_line_set_data(line, 3, "UNKNOWN");
+			break;
```
rjarry, line 297:
```c
default:
   scols_line_sprintf(line, 3, "%u", iface->mode);
   break;
```

modules/infra/cli/port.c
```diff
@@ -180,12 +198,75 @@ static cmd_status_t rxq_list(const struct gr_api_client *c, const struct ec_pnod
...
+	void *resp_ptr = NULL;
+
+	iface_from_name(c, arg_str(p, "IFACE"), &req_iface);
+
+	if (gr_api_client_send_recv(c, GR_INFRA_IFACE_LIST, sizeof(req), &req, &resp_ptr) < 0) {
+		scols_unref_table(table);
+		return CMD_ERROR;
+	}
+	resp = resp_ptr;
+	// qsort(resp->ifaces, resp->n_ifaces, sizeof(*resp->ifaces), iface_order);
```
rjarry, line 217: remove?

rjarry (changes requested):

docs/graph.svg
christophefontaine: There should be a change though, as I added a node between RX and TX ?

modules/infra/api/gr_infra.h
christophefontaine: Yes, I'll add that to its own patch though.

modules/infra/cli/port.c
```diff
@@ -24,12 +25,28 @@ static void port_show(const struct gr_api_client *, const struct gr_iface *iface
 	printf("n_txq: %u\n", port->n_txq);
 	printf("rxq_size: %u\n", port->rxq_size);
 	printf("txq_size: %u\n", port->txq_size);
+
+	if (iface->mode == GR_IFACE_MODE_L1_XC) {
+		if (iface_from_id(c, iface->domain_id, &peer) == 0)
+			printf("xc peer: %s\n", peer.name);
+		else
+			printf("xc peer: unknown\n");
```
rjarry, line 33: cosmetics:

s/xc peer/xc_peer/

And if iface_from_id returns an error, just print the domain_id as is.

modules/infra/cli/port.c
```diff
@@ -24,12 +25,28 @@ static void port_show(const struct gr_api_client *, const struct gr_iface *iface
...
-	snprintf(buf, len, "devargs=%s mac=" ETH_F, port->devargs, &port->mac);
+	struct gr_iface peer;
+	size_t s;
+
+	s = snprintf(buf, len, "devargs=%s mac=" ETH_F, port->devargs, &port->mac);
+	if (iface->mode == GR_IFACE_MODE_L1_XC) {
+		if (iface_from_id(c, iface->domain_id, &peer) == 0)
+			snprintf(&buf[s], len - s, " xc peer -> %s", peer.name);
+		else
+			snprintf(&buf[s], len - s, " xc peer -> unknown");
```
rjarry, line 48: Follow the same style as the rest of the attributes:

"xc_peer=$peer_name" or if iface_from_id fails, "xc_peer=$domain_id"

Can you use SAFE_BUF here?

rjarry (changes requested):

PR #154, opened 2025-02-12
Replace defines with enums

christophefontaine:
Make our lives a bit easier when debugging, switching from opaque flags value to an explicit enum.

Instead of 
```
(gdb) p nh->flags
$7 = 178
```
We have now:
```
(gdb) p nh->flags
$7 = (GR_NH_F_REACHABLE | GR_NH_F_STATIC | GR_NH_F_LOCAL | GR_NH_F_LINK)
```

### DIFF
--- a/modules/infra/api/gr_infra.h
+++ b/modules/infra/api/gr_infra.h
@@ -14,20 +14,27 @@
 #include <stdint.h>
 #include <sys/types.h>
 
-// Value for gr_iface.type
-#define GR_IFACE_TYPE_UNDEF 0x00
-#define GR_IFACE_TYPE_PORT 0x01
-#define GR_IFACE_TYPE_VLAN 0x02
-#define GR_IFACE_TYPE_LOOPBACK 0xFF
-#define GR_IFACE_TYPE_MAX 256
+typedef enum {
+	GR_IFACE_TYPE_UNDEF = 0x00,
+	GR_IFACE_TYPE_PORT = 0x01,
+	GR_IFACE_TYPE_VLAN = 0x02,
+	GR_IFACE_TYPE_IPIP = 0x03,
+	GR_IFACE_TYPE_LOOPBACK = 0xff,
+	GR_IFACE_TYPE_MAX = 256
+} __attribute__((mode(HI))) gr_iface_type_t;
 
 // Interface configure flags
-#define GR_IFACE_F_UP GR_BIT16(0)
-#define GR_IFACE_F_PROMISC GR_BIT16(1)
-#define GR_IFACE_F_ALLMULTI GR_BIT16(2)
-#define GR_IFACE_F_PACKET_TRACE GR_BIT16(3)
+typedef enum {
+	GR_IFACE_F_UP = GR_BIT16(0),
+	GR_IFACE_F_PROMISC = GR_BIT16(1),
+	GR_IFACE_F_ALLMULTI = GR_BIT16(2),
+	GR_IFACE_F_PACKET_TRACE = GR_BIT16(3),
+} __attribute__((mode(HI))) gr_iface_flags_t;
+
 // Interface state flags
-#define GR_IFACE_S_RUNNING GR_BIT16(0)
+typedef enum {
+	GR_IFACE_S_RUNNING = GR_BIT16(0),
+} __attribute__((mode(HI))) gr_iface_state_t;
 
 // Interface reconfig attributes
 #define GR_IFACE_SET_FLAGS GR_BIT64(0)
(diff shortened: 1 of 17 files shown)


modules/infra/api/gr_infra.h
```diff
@@ -15,19 +15,26 @@
...
-#define GR_IFACE_TYPE_VLAN 0x02
-#define GR_IFACE_TYPE_LOOPBACK 0xFF
-#define GR_IFACE_TYPE_MAX 256
+typedef enum {
+	GR_IFACE_TYPE_UNDEF = 0x00,
+	GR_IFACE_TYPE_PORT = 0x01,
+	GR_IFACE_TYPE_VLAN = 0x02,
+	GR_IFACE_TYPE_LOOPBACK = 0xFF,
+	GR_IFACE_TYPE_MAX = 256,
+} __attribute__((packed)) gr_iface_type_t;
```
vjardin, line 24: instead of packing, if you want 8 bits or 16 bits, you should used:

__attribute__((mode(HI))) gr_iface_type_t; // 16 bits
__attribute__((mode(QI))) gr_iface_type_t; // 8 bits

modules/infra/api/gr_infra.h
```diff
@@ -15,19 +15,26 @@
...
-#define GR_IFACE_F_UP GR_BIT16(0)
-#define GR_IFACE_F_PROMISC GR_BIT16(1)
-#define GR_IFACE_F_ALLMULTI GR_BIT16(2)
-#define GR_IFACE_F_PACKET_TRACE GR_BIT16(3)
+typedef enum {
+	GR_IFACE_F_UP = GR_BIT16(0),
+	GR_IFACE_F_PROMISC = GR_BIT16(1),
+	GR_IFACE_F_ALLMULTI = GR_BIT16(2),
+	GR_IFACE_F_PACKET_TRACE = GR_BIT16(3),
+} __attribute__((packed)) gr_iface_flags_t;
```
vjardin, line 32: same comment

modules/ipip/gr_ipip.h
```diff
@@ -10,7 +10,9 @@
 #include <gr_macro.h>
 #include <gr_net_types.h>
 
-#define GR_IFACE_TYPE_IPIP 0x03
+typedef enum {
+	GR_IFACE_TYPE_IPIP = 0x03,
+} __attribute__((packed)) gr_iface_type_ipip_t;
```
vjardin, line 15: same comment

vjardin (changes requested):

modules/ipip/gr_ipip.h
```diff
@@ -10,7 +10,9 @@
 #include <gr_macro.h>
 #include <gr_net_types.h>
 
-#define GR_IFACE_TYPE_IPIP 0x03
+typedef enum {
+	GR_IFACE_TYPE_IPIP = 0x03,
+} __attribute__((mode(QI))) gr_iface_type_ipip_t;
```
rjarry, line 15: That's what I was concerned with. How can we extend an enum without defining other types?

rjarry (changes requested):

modules/ipip/gr_ipip.h
christophefontaine: We could keep a #define, I don't have a strong preference.

We could define the iface type in the main module.
This doesn't create a `hard` link between the modules, as we just add a new value in the enum.
And we don't have any support of runtime loadable modules anyway.

modules/ipip/gr_ipip.h
christophefontaine: fixed, thanks !

modules/infra/api/gr_infra.h
christophefontaine: That's exactly why I love open source: thanks for that comment !
Fixed :)

modules/ipip/gr_ipip.h
rjarry: yes, it would be better with all values in the infra module.

christophefontaine: ack !
I'll also modify all function signatures to use the new type. (That's why I modified this PR as a draft)

christophefontaine: fixed.

modules/ipip/cli.c
```diff
@@ -73,7 +73,7 @@ static uint64_t parse_ipip_args(
 static cmd_status_t ipip_add(const struct gr_api_client *c, const struct ec_pnode *p) {
 	const struct gr_infra_iface_add_resp *resp;
 	struct gr_infra_iface_add_req req = {
-		.iface = {.type = GR_IFACE_TYPE_IPIP, .flags = GR_IFACE_F_UP}
+		.iface = {.type = (gr_iface_type_t)GR_IFACE_TYPE_IPIP, .flags = GR_IFACE_F_UP}
```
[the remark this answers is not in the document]
christophefontaine: I forgot to remove that cast: it was required when we had a dedicated enum for GR_IFACE_TYPE_IPIP.

modules/infra/api/gr_infra.h
```diff
@@ -15,19 +15,27 @@
...
-#define GR_IFACE_TYPE_PORT 0x01
-#define GR_IFACE_TYPE_VLAN 0x02
-#define GR_IFACE_TYPE_LOOPBACK 0xFF
-#define GR_IFACE_TYPE_MAX 256
+typedef enum {
+	GR_IFACE_TYPE_UNDEF = 0x00,
+	GR_IFACE_TYPE_PORT = 0x01,
+	GR_IFACE_TYPE_VLAN = 0x02,
+	GR_IFACE_TYPE_IPIP = 0x03,
+	GR_IFACE_TYPE_LOOPBACK = 0xFF,
```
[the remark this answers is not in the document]
christophefontaine: done !

rjarry (changes requested):

rjarry (approved):

PR #156, opened 2025-02-13
ip/ip6: fix ip(6) address deletion

christophefontaine:
ip and ip6 address deletion never worked, and returned EBUSY.
Only check that the nexthop isn't referenced by another entity instead of checking the flags.


rjarry (commented):
:+1:

rjarry (approved):

PR #158, opened 2025-02-13
ip6_local: allow registering handlers for extension headers

rjarry:
Allow registering a dynamic edge based on extension headers (e.g. IPPROTO_ROUTING for SRv6).

The next node will receive packets with their offsets adjusted at the beginning of the extension header (or at the beginning of the L4 header as it was before this patch).

Only verify the checksum for L4 protocols that actually carry one, namely: UDP, TCP, ICMPv6. There are other L4 protocols that carry a checksum field but grout does not support them yet (DCCP, SCTP, ...)


modules/ip6/datapath/ip6_local.c
```diff
@@ -77,17 +77,25 @@ static uint16_t ip6_input_local_process(
...
+		case IPPROTO_UDP:
+			// verify checksum if not already checked by hardware
+			switch (m->ol_flags & RTE_MBUF_F_RX_L4_CKSUM_MASK) {
+			case RTE_MBUF_F_RX_L4_CKSUM_NONE:
+			case RTE_MBUF_F_RX_L4_CKSUM_UNKNOWN:
+				if (rte_ipv6_udptcp_cksum_verify(
+					    ip, rte_pktmbuf_mtod_offset(m, void *, l3_hdr_size)
+				    )) {
+					edge = BAD_CHECKSUM;
+					goto next;
```
david-marchand, line 92: Nit: those jumps to next seem unneeded.

rjarry: These were a remnant of poorly ordered patches. It should make more sense now.

ogournet:
I don't think that relying on the presence of SRH in packet for local processing is enough.
The configured localSIDs should be matched against DA.

from RFC 8986 sec 4.
```
When an SRv6-capable node (N) receives an IPv6 packet whose
destination address matches a FIB entry that represents a locally
instantiated SRv6 SID (S), the IPv6 header chain is processed as
defined in Section 4 of [RFC8200].  For SRv6 SIDs associated with an
Endpoint behavior defined in this document, the SRH and Upper-Layer
header are processed as defined in the following subsections.
```
There is an example at  sec 4.16.1.3 that shows that SRH can be removed before getting to the node that does the decap.

rjarry:
@ogournet what do you think about using local ipv6 addresses as SID? This makes all the configuration and intermediate/endpoint processing much simpler.

For the steer/entrypoint case, we need custom handling in ip6_output.

ogournet:
> @ogournet what do you think about using local ipv6 addresses as SID? This makes all the configuration and intermediate/endpoint processing much simpler.

Simpler... it can not be simpler than my current PR on localsid.
But yes, we could try it, it makes sense.

modules/ip6/datapath/ip6_local.c
```diff
@@ -57,46 +56,65 @@ static uint16_t ip6_input_local_process(
...
-		case RTE_MBUF_F_RX_L4_CKSUM_UNKNOWN:
-			if (rte_ipv6_udptcp_cksum_verify(
-				    ip, rte_pktmbuf_mtod_offset(m, void *, l3_hdr_size)
-			    ))
-				edge = BAD_CHECKSUM;
+		switch (d->proto) {
+		case IPPROTO_HOPOPTS:
+		case IPPROTO_ROUTING:
+		case IPPROTO_DSTOPTS:
+		case IPPROTO_FRAGMENT:
```
david-marchand, line 88: This double switch + fallback + rte_pktmbuf_adj makes me unconfortable.
I hope we won't touch this code often...


Just a thought at the current code:

It sounds like DPDK is missing a helper like rte_ipv6_is_ext(int proto) (or a better name).
That could be used in rte_ipv6_get_next_ext() existing helper.

```
diff --git a/lib/net/rte_ip6.h b/lib/net/rte_ip6.h
index 92558a124a..fe78bba2b2 100644
--- a/lib/net/rte_ip6.h
+++ b/lib/net/rte_ip6.h
@@ -762,6 +762,21 @@ struct __rte_aligned(2) __rte_packed_begin rte_ipv6_fragment_ext {
 /* IPv6 fragment extension header size */
 #define RTE_IPV6_FRAG_HDR_SIZE sizeof(struct rte_ipv6_fragment_ext)
 
+static inline bool
+rte_ipv6_is_ext(int proto)
+{
+       switch (proto) {
+       case IPPROTO_AH:
+       case IPPROTO_HOPOPTS:
+       case IPPROTO_ROUTING:
+       case IPPROTO_DSTOPTS:
+       case IPPROTO_FRAGMENT:
+               return true;
+       default:
+               return false;
+       }
+}
+
 /**
  * Parse next IPv6 header extension
  *
@@ -783,6 +798,9 @@ rte_ipv6_get_next_ext(const uint8_t *p, int proto, size_t *ext_len)
 {
        int next_proto;
 
+       if (!rte_ipv6_is_ext(proto))
+               return -EINVAL;
+
        switch (proto) {
        case IPPROTO_AH:
                next_proto = *p++;
@@ -800,9 +818,6 @@ rte_ipv6_get_next_ext(const uint8_t *p, int proto, size_t *ext_len)
                next_proto = *p;
                *ext_len = RTE_IPV6_FRAG_HDR_SIZE;
                break;
-
-       default:
-               return -EINVAL;
        }
 
        return next_proto;
```

And that would translate as an updated change like:
```
                        d->len -= ext_size;
                        d->proto = next_proto;
                };
 
-               edge = edges[d->proto];
-               if (edge == UNKNOWN_PROTO)
+               // if last proto is an option, or we stopped and have a handler,
+               // there is no csum to check and no adjustment to do
+               if (rte_ip6_is_ext(d->proto))
                        goto next;
 
                // verify checksum if not already checked by hardware
@@ -88,7 +88,7 @@ static uint16_t ip6_input_local_process(
                case RTE_MBUF_F_RX_L4_CKSUM_NONE:
                case RTE_MBUF_F_RX_L4_CKSUM_UNKNOWN:
                        if (rte_ipv6_udptcp_cksum_verify(
-                                   ip, rte_pktmbuf_mtod_offset(m, void *, l3_hdr_size)
+                                   ip, rte_pktmbuf_mtod_offset(m, void *, d->ext_offset)
                            ))
                                edge = BAD_CHECKSUM;
                        break;
@@ -96,7 +96,8 @@ static uint16_t ip6_input_local_process(
                        edge = BAD_CHECKSUM;
                        break;
                }
-               rte_pktmbuf_adj(m, l3_hdr_size);
+               rte_pktmbuf_adj(m, d->ext_offset);
+               d->ext_offset = 0;
 next:
                rte_node_enqueue_x1(graph, node, edge, m);
        }
```

PR #160, opened 2025-02-13
infra: simplify show interface cli

christophefontaine:
Instead of using 'show interface all' to list all interfaces, we can just type 'show interface'.
Update doc and tests.


vjardin:
LGTM

david-marchand:
It should be fine.

It is a bit surprising to find that there is a "api handler" whose description string is "show interface ra" (which has nothing to do with cli commands..). Not sure this "name" for each API handler helps much, but that's a different topic.

rjarry (approved):

PR #161, opened 2025-02-14
cli: add arg_ip* helpers

ogournet:
Add helpers, only for cli:
  arg_ip4
  arg_ip6
  arg_ip4_net
  arg_ip6_net

And use them on ip/ip6 cli modules.

### DIFF
--- a/cli/ecoli.c
+++ b/cli/ecoli.c
@@ -162,3 +162,46 @@ int arg_eth_addr(const struct ec_pnode *p, const char *id, struct rte_ether_addr
 
 	return 0;
 }
+
+static int arg_ip(const struct ec_pnode *p, const char *id, void *addr, int family) {
+	const char *str = arg_str(p, id);
+	if (str == NULL)
+		return -errno;
+
+	if (inet_pton(family, str, addr) != 1)
+		return errno_set(EINVAL);
+
+	return 0;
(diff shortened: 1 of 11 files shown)


cli/ecoli.c
```diff
@@ -162,3 +162,44 @@ int arg_eth_addr(const struct ec_pnode *p, const char *id, struct rte_ether_addr
...
+	const char *str = arg_str(p, id);
+	if (str == NULL)
+		return -errno;
+
+	if (family == AF_INET)
+		return ip4_net_parse(str, net, zero_mask);
+	else
+		return ip6_net_parse(str, net, zero_mask);
+
+	return 0;
```
rjarry, line 196: nit pick, `return 0` is unreachable, can you do this instead?

```c
switch (family)
case AF_INET:
        return ip4_net_parse(str, net, zero_mask);
case AF_INET6:
        return ip6_net_parse(str, net, zero_mask);
default:
        return errno_set(EINVAL);
}
```

rjarry: I know this function is static and is not called with any other family value. But this cannot hurt and helps with readablility.

rjarry (changes requested):

rjarry (approved):

PR #163, opened 2025-02-17
infra: set iova mode

christophefontaine:
When running grout in a VM and hotplugging a port, the IOVA mode used is always VA, which leads to the impossibility to bind the port.

ERR: PCI_BUS:   Expecting 'PA' IOVA mode but current mode is 'VA', not initializing

Check if 'noiommu' is enabled, and sets iovamode to "PA".

Suggested-by: David Marchand <david.marchand@redhat.com>

### DIFF
--- a/main/dpdk.c
+++ b/main/dpdk.c
@@ -15,6 +15,7 @@
 #include <rte_log.h>
 #include <rte_mempool.h>
 #include <rte_version.h>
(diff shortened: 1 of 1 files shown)


david-marchand:
An option to override this detection may be needed in the future, but this should be enough for now.

Ack for me.

rjarry (approved):

PR #164, opened 2025-02-17
ci: add clang and older gcc checks

rjarry:
Add multiple jobs to ensure that the compilation works on gcc 11, 14 and clang 13 and 18.

Enable ccache to get results faster.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -25,14 +25,14 @@ jobs:
       MESON_EXTRA_OPTS: "-Ddpdk:platform=generic"
       DEBIAN_FRONTEND: noninteractive
       NEEDRESTART_MODE: l
-      CC: gcc-14
+      CC: ccache gcc-14
     steps:
       - name: install system dependencies
         run: |
           set -xe
           sudo apt-get update -qy
           sudo apt-get install -qy --no-install-recommends \
-            make gcc ninja-build meson git go-md2man libibverbs-dev \
+            make gcc ccache ninja-build meson git go-md2man libibverbs-dev \
             libasan8 libcmocka-dev libedit-dev libarchive-dev \
             libevent-dev libsmartcols-dev libnuma-dev python3-pyelftools \
             socat tcpdump traceroute graphviz iproute2 iputils-ping ndisc6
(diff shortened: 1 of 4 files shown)


.github/workflows/check.yml
```diff
@@ -25,14 +25,14 @@ jobs:
       MESON_EXTRA_OPTS: "-Ddpdk:platform=generic"
       DEBIAN_FRONTEND: noninteractive
       NEEDRESTART_MODE: l
-      CC: gcc-14
+      CC: ccache gcc-14
```
david-marchand: IIRC, meson detects ccache presence and appends it automatically to the compiler.
So you don't need to set ccache in the CC: variable.

.github/workflows/check.yml
```diff
@@ -36,6 +51,7 @@ jobs:
             libasan8 libcmocka-dev libedit-dev libarchive-dev \
             libevent-dev libsmartcols-dev libnuma-dev python3-pyelftools \
             socat tcpdump traceroute graphviz iproute2 iputils-ping ndisc6
+          mkdir -p ~/.ccache
```
david-marchand, line 54: Why is it needed?

rjarry: I thought it was after reading a warning from the cache action stating that the folder didn't exist and didn't cache anything.

.github/workflows/check.yml
rjarry: I copy pasted this from the dpdk GitHub workflow. I will try removing it.

.github/workflows/check.yml
david-marchand: In any case, if this is really needed, it should go to the previous patch.

modules/l4/l4_input_local.c
```diff
@@ -24,11 +24,12 @@ enum edges {
...
 
 void l4_input_register_port(uint8_t proto, rte_be16_t port, const char *next_node) {
-	LOG(DEBUG, "l4_input_register_port: proto=%hhu port=%hu-> %s", proto, port, next_node);
+	uint16_t p = rte_be_to_cpu_16(port);
+	LOG(DEBUG, "l4_input_register_port: proto=%hhu port=%u -> %s", proto, p, next_node);
 	switch (proto) {
 	case IPPROTO_UDP:
 		if (udp_edges[port] != MANAGEMENT)
-			ABORT("next node already registered for udp port=%hhu", port);
+			ABORT("next node already registered for udp port=%u", p);
```
christophefontaine, line 32: That should be "%hu" ?

rjarry: Yes, probably. I will need to update lots of other places accordingly.

PR #168, opened 2025-02-19
clang 15

christophefontaine:
fixes to be able to build with clang 15.


christophefontaine:
Clang 15 is available in RHEL9.2, so yes, I can add it to the build matrix if you wish.

rjarry:
Thanks! That will prevent breaking clang 15 in the future :)

rjarry (approved):

PR #169, opened 2025-02-19
ci: add clang 15

christophefontaine:
Add clang 15 on ubuntu 22.04 to the build matrix.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -25,16 +25,24 @@ jobs:
           - compiler: gcc-13
             sanitize: none
             buildtype: debugoptimized
+            os: ubuntu-24.04
           - compiler: gcc-14
             sanitize: address
             buildtype: debug
+            os: ubuntu-24.04
+          - compiler: clang-15
+            sanitize: none
+            buildtype: debugoptimized
+            os: ubuntu-22.04
           - compiler: clang-16
             sanitize: none
             buildtype: debugoptimized
+            os: ubuntu-24.04
           - compiler: clang-18
             sanitize: none
             buildtype: debugoptimized
-    runs-on: ubuntu-24.04
+            os: ubuntu-24.04
+    runs-on: ${{ matrix.conf.os }}
     env:
       SANITIZE: ${{ matrix.conf.sanitize }}
       BUILDTYPE: ${{ matrix.conf.buildtype }}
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yml
```diff
@@ -51,6 +59,11 @@ jobs:
             libasan8 libcmocka-dev libedit-dev libarchive-dev \
             libevent-dev libsmartcols-dev libnuma-dev python3-pyelftools \
             socat tcpdump traceroute graphviz iproute2 iputils-ping ndisc6
+          echo -e "$(meson --version)\n0.63.0" | sort -V
+          if [[ $0 -eq 0 ]] ; then
```
rjarry, line 63: Bashism, please change to:

```bash
if printf '%s\n0.63.0\n' "$(meson --version)" | sort -V; then
```

it should be more portable.

.github/workflows/check.yml
```diff
@@ -51,6 +59,11 @@ jobs:
             libasan8 libcmocka-dev libedit-dev libarchive-dev \
             libevent-dev libsmartcols-dev libnuma-dev python3-pyelftools \
             socat tcpdump traceroute graphviz iproute2 iputils-ping ndisc6
+          echo -e "$(meson --version)\n0.63.0" | sort -V
+          if [[ $0 -eq 0 ]] ; then 
+            sudo apt install -y python3-pip
+            pip3 install -U meson
```
rjarry, line 65: To make it explicit:

```bash
pip3 install --user -U meson
```

rjarry (changes requested):

.github/workflows/check.yml
```diff
@@ -51,6 +59,10 @@ jobs:
             libasan8 libcmocka-dev libedit-dev libarchive-dev \
             libevent-dev libsmartcols-dev libnuma-dev python3-pyelftools \
             socat tcpdump traceroute graphviz iproute2 iputils-ping ndisc6
+          if printf '%s\n0.63.0\n' "$(meson --version)" | sort -V -C ; then
```
rjarry, line 62: I think it would make it easier to understand in the reverse order:

```bash
if ! printf '0.63.0\n%s\n' "$(meson --version)" | sort -VC; then
```

In this form, it reads "if not properly sorted, install meson"

.github/workflows/check.yml
```diff
@@ -51,6 +59,10 @@ jobs:
...
+            sudo apt install -y python3-pip
+            pip3 install --user -U meson
```
rjarry, line 64: Maybe hardcode a meson version that is known to work so that we don't get surprises?

rjarry (changes requested):

.github/workflows/check.yml
christophefontaine: worst case scenario is that we will have to update to be compliant with the newer version of meson.
And we would need that anyway, wouldn't we ?

rjarry: The problem is that it could make the CI flaky depending on the "latest" version of meson available.

I prefer having something reproducible.

.github/workflows/check.yml
christophefontaine: I don't think this is easier to understand ;-) 
But anyway....

rjarry: ok nevermind ;)

rjarry (approved):

PR #170, opened 2025-02-19
ip,ip6: handle nexthop types

rjarry:
Allow registering dynamic next nodes after ip_output and ip6_output based on the nexthop type.

This will allow redirecting packets to different nodes for special processing, such as SRv6.


rjarry:
@ogournet this should help as a start to implement packet steering from `ip{,6}_output` into `sr6_output` based on a new nexthop type that you could add: `GR_NH_SRV6_SID`.

ogournet:
thanks, I'm going to give it a chance

PR #174, opened 2025-03-04
misc fixes: ip6

christophefontaine:
- Simple fix on interface creation.
- Cleanup ip6 rib when an IPv6 address is deleted

### DIFF
--- a/modules/ip6/control/route.c
+++ b/modules/ip6/control/route.c
@@ -393,9 +393,13 @@ void rib6_cleanup(struct nexthop *nh) {
 	struct rte_rib6 *rib;
 	uintptr_t nh_id;
 
-	rib6_delete(nh->vrf_id, nh->iface_id, &nh->ipv6, RTE_IPV6_MAX_DEPTH);
 	local_ip = nh->ipv6;
 	local_depth = nh->prefixlen;
+
+	if (nh->flags & (GR_NH_F_LOCAL | GR_NH_F_LINK))
+		rib6_delete(nh->vrf_id, nh->iface_id, &nh->ipv6, nh->prefixlen);
+	else
+		rib6_delete(nh->vrf_id, nh->iface_id, &nh->ipv6, RTE_IPV6_MAX_DEPTH);
 
 	rib = get_rib6(nh->vrf_id);
 	while ((rn = rte_rib6_get_nxt(rib, 0, 0, rn, RTE_RIB6_GET_NXT_ALL)) != NULL) {
--- a/modules/infra/datapath/main_loop.c
+++ b/modules/infra/datapath/main_loop.c
@@ -210,7 +210,8 @@ reconfig:
 shutdown:
 	log(NOTICE, "shutting down tid=%d", w->tid);
 	atomic_store(&w->stats, NULL);
-	rte_graph_cluster_stats_destroy(ctx.stats);
+	if (ctx.stats)
+		rte_graph_cluster_stats_destroy(ctx.stats);
 	rte_free(ctx.w_stats);
 	rte_thread_unregister();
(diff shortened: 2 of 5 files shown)


modules/ip6/control/address.c
```diff
@@ -236,6 +241,8 @@ static struct api_out addr6_del(const void *request, void ** /*response*/) {
 		return api_out(EBUSY, 0);
 
 	rib6_cleanup(nh);
+	rib6_delete(iface->vrf_id, iface->id, &nh->ipv6, nh->prefixlen);
+	nexthop_decref(nh);
```
rjarry, line 245: In the commit message, you say:

> Instead of calling rib6_cleanup, call rib6_delete and nexthop_decref.

But you are doing both. Is this intended? Maybe should be fixing `rib6_cleanup()` instead?

modules/ip6/control/address.c
```diff
@@ -325,8 +332,11 @@ static void ip6_iface_event_handler(uint32_t event, const void *obj) {
 	case IFACE_EVENT_PRE_REMOVE:
 		struct hoplist *addrs = &iface_addrs[iface->id];
 
-		gr_vec_foreach (nh, addrs->nh)
-			rib6_cleanup(nh);
+		gr_vec_foreach (nh, addrs->nh) {
+			rib6_delete(iface->vrf_id, iface->id, &nh->ipv6, nh->prefixlen);
+			nexthop_decref(nh);
```
rjarry, line 337: And here you're actually replacing rib6_cleanup with rib6_delete, as indicated in the commit message.

rjarry (changes requested):
I'm not sure about the fix. Let me know what you think.

modules/ip6/control/address.c
christophefontaine: Maybe we should fix rib6_cleanup indeed, this would fix  ip6_iface_event_handler as well.

modules/ip6/control/address.c
christophefontaine: I modified rib6_cleanup instead, I'm not 100% sure of the fix though, even if the updated smoke test pass.

modules/ip6/control/route.c
```diff
@@ -393,7 +393,7 @@ void rib6_cleanup(struct nexthop *nh) {
 	struct rte_rib6 *rib;
 	uintptr_t nh_id;
 
-	rib6_delete(nh->vrf_id, nh->iface_id, &nh->ipv6, RTE_IPV6_MAX_DEPTH);
+	rib6_delete(nh->vrf_id, nh->iface_id, &nh->ipv6, nh->prefixlen);
```
rjarry, line 396: I have a feeling that we should keep this at the beginning:

```c
rib6_delete(nh->vrf_id, nh->iface_id, &nh->ipv6, RTE_IPV6_MAX_DEPTH);
```

And only call the second one if `nh->prefixlen != 0` (i.e. it is a local interface address). Gateway nexthops do not have their prefixlen set to anything:

```c
if (nh->prefixlen != 0)
        rib6_delete(nh->vrf_id, nh->iface_id, &nh->ipv6, nh->prefixlen);
```

rjarry (changes requested):
I didn't check in detail, but we may need to audit rib*_cleanup functions to ensure no loose treads.

modules/ip6/control/route.c
christophefontaine: Yes, that's why I didn't want to merge this immediately (even if I need it !) that I put this PR in draft first.
If the gateway NH has not prefix len set, shouldn't we instead set it to RTE_IPv6_MAX_DEPTH / RTE_IPv4_MAX_DEPTH ?

rjarry (approved):

PR #177, opened 2025-03-08
ip: enhance ip header validation

maxime-leroy:
The current code assumes that the hardware performs the following checks:
- The packet length is sufficient to accommodate the minimum length of the IP header.
- The IP version is set to 4.
- The IP header length field is large enough to accommodate the minimum legal IP datagram length.

However, these assumptions are incorrect, at least with the tap driver. The following packets can be created using Scapy:

> p_badversion = Ether(dst='f0:0d:ac:dc:00:00', src='ba:d0:ca:ca:00:00')/IP(version=5, src='172.16.0.2', dst='172.16.1.2')
> p_badihl = Ether(dst='f0:0d:ac:dc:00:00', src='ba:d0:ca:ca:00:00')/IP(src='172.16.0.2', dst='172.16.1.2', ihl=3)
> p_truncated = Ether(dst='f0:0d:ac:dc:00:00', src='ba:d0:ca:ca:00:00')/IP(src='172.16.0.2', dst='172.16.1.2')[:14+10]

If the correct routes are configured, grout will forward these packets instead of discarding them, in violation of RFC1812.

The DPDK driver can provide hardware information about the checksum (which is already used to validate the IP checksum) and packet type. There are two packet types for IPv4:
- RTE_PTYPE_L3_IPV4: Ethernet type 0x0800, version 4, and IHL of 5.
- RTE_PTYPE_L3_IPV4_EXT: Ethernet type 0x0800, version 4, IHL in the range of 6–15, with options.

When one of these flags is set, it indicates that the IHL and version of the packet are valid. Otherwise, software validation is necessary.

As for the packet length requirement to accommodate the minimum legal IP datagram length (20 bytes), there is no guarantee that the hardware will validate it. Therefore, this validation must be performed in software.

Fixes: 5f5de2c05188 ("ip4_lookup: comply with rfc1812")

### DIFF
--- a/modules/ip/datapath/ip_input.c
+++ b/modules/ip/datapath/ip_input.c
@@ -25,6 +25,7 @@ enum edges {
 	NO_ROUTE,
 	BAD_CHECKSUM,
 	BAD_LENGTH,
+	BAD_VERSION,
 	OTHER_HOST,
 	EDGE_COUNT,
 };
@@ -53,7 +54,14 @@ ip_input_process(struct rte_graph *graph, struct rte_node *node, void **objs, ui
 		//
 		// (1) The packet length reported by the Link Layer must be large
 		//     enough to hold the minimum length legal IP datagram (20 bytes).
-		// XXX: already checked by hardware
+		if (rte_pktmbuf_data_len(mbuf) < sizeof(struct rte_ipv4_hdr)) {
+			// XXX: call rte_pktmuf_data_len is used to ensure that the IPv4 header
+			// is located on the first segment. IPv4 headers located on the second,
+			// third or subsequent segments, as well spanning segment boundaries, are
+			// not currently handled.
+			edge = BAD_LENGTH;
+			goto next;
+		}
 
 		// (2) The IP checksum must be correct.
 		switch (mbuf->ol_flags & RTE_MBUF_F_RX_IP_CKSUM_MASK) {
@@ -73,9 +81,17 @@ ip_input_process(struct rte_graph *graph, struct rte_node *node, void **objs, ui
 		// (3) The IP version number must be 4.  If the version number is not 4
 		//     then the packet may be another version of IP, such as IPng or
 		//     ST-II.
+		if (ip->version != IPVERSION) {
+			edge = BAD_VERSION;
+			goto next;
+		}
+
 		// (4) The IP header length field must be large enough to hold the
 		//     minimum length legal IP datagram (20 bytes = 5 words).
-		// XXX: already checked by hardware
+		if (rte_ipv4_hdr_len(ip) < sizeof(struct rte_ipv4_hdr)) {
+			edge = BAD_LENGTH;
+			goto next;
+		}
 
 		// (5) The IP total length field must be large enough to hold the IP
 		//     datagram header, whose length is specified in the IP header
@@ -148,6 +164,7 @@ static struct rte_node_register input_node = {
 		[NO_ROUTE] = "ip_error_dest_unreach",
 		[BAD_CHECKSUM] = "ip_input_bad_checksum",
 		[BAD_LENGTH] = "ip_input_bad_length",
+		[BAD_VERSION] = "ip_input_bad_version",
 		[OTHER_HOST] = "ip_input_other_host",
 	},
 };
@@ -162,4 +179,5 @@ GR_NODE_REGISTER(info);
 
 GR_DROP_REGISTER(ip_input_bad_checksum);
 GR_DROP_REGISTER(ip_input_bad_length);
+GR_DROP_REGISTER(ip_input_bad_version);
 GR_DROP_REGISTER(ip_input_other_host);


modules/ip/datapath/ip_input.c
```diff
@@ -25,10 +25,82 @@ enum edges {
...
+	uint8_t ptype_l3;
+
+	ip = rte_pktmbuf_mtod(mbuf, struct rte_ipv4_hdr *);
+
+	// RFC 1812 section 5.2.2 IP Header Validation
+	//
+	// (1) The packet length reported by the Link Layer must be large
+	//     enough to hold the minimum length legal IP datagram (20 bytes).
+	if (unlikely(rte_pktmbuf_data_len(mbuf) < sizeof(struct rte_ipv4_hdr))) {
+		// XXX: call rte_pktmbuf_linearize ?
```
rjarry, line 55: You can possibly use `rte_pktmbuf_pkt_len()` which takes care of multi-segment mbufs.

modules/ip/datapath/ip_input.c
```diff
@@ -25,10 +25,82 @@ enum edges {
...
 
+// Retrieves a pointer to the IP header from a mbuf.
+// The IP header is validated according to RFC 1812 section 5.2.2.
+//
+// @param mbuf
+//   A pointer to the mbuf. The IP header must be at the beginning (offset 0) and
+//   in the first segment of the mbuf.
+// @param edge
+//   If the IPv4 header does not comply with RFC 1812, this is set to one of the error codes:
+//   BAD_LENGTH, BAD_VERSION, or BAD_CHECKSUM. If the header is valid, it is set to 0.
```
rjarry, line 41: `if the header is valid, edge will be left untouched`

rjarry (changes requested):
Hi Maxime,

welcome to the project :)

I only have a couple of minor remarks. Please see below.

modules/ip/datapath/ip_input.c
maxime-leroy: Using rte_pktmbuf_data_len() ensures that the IP header is within the first segment, preventing out-of-bounds reads when the header spans multiple segments.

If the header is split between segments  can lead to incorrect parsing or crashes.

While this is unlikely with a 2K mbuf size, it’s not guaranteed.

To handle multi-segment packets, we could linearize the mbuf to ensureys the required header bytes are contiguous.

Or extracts just the necessary header portion without modifying the mbuf.

I have seen multiple places, we are presuming the data is always in the first segment. 

A future patch series could refactor these cases to properly handle larger MTUs and multi-segment packets.

modules/ip/datapath/ip_input.c
maxime-leroy: Good catch. Thanks

modules/ip/datapath/ip_input.c
rjarry: OK, fair points. As an alternative, we could use `rte_pktmbuf_read()` to avoid linearizing the whole mbuf. But as you said having an ipv4 header across a segment boundary is really unlikely.

maxime-leroy: I can use rte_pktmbuf_read(), but it only solves the issue for the ip_input node. In the ip_forward node, we modify the TTL directly by reading and writing the IP header without handling this unlikely case.

I looked into rte_pktmbuf_linearize(), which converts a multi-segment mbuf into a single contiguous buffer. However, if the mbuf size exceeds rx_mbuf_size (2048 bytes), such as in the case of jumbo frames, it will always fail.

What we need is a function that attempts to move as much data as possible into the first segment—essentially an rte_pktmbuf_compact() function.

modules/ip/datapath/ip_input.c
rjarry: Do you want to send an update for this typo? Or I can do it when applying the patch.

modules/ip/datapath/ip_input.c
rjarry: Yep, we'll deal with this when the time comes.

modules/ip/datapath/ip_input.c
maxime-leroy: I will do it. Add a comment for multi segement case. Thanks

modules/ip/datapath/ip_input.c
```diff
@@ -25,10 +25,84 @@ enum edges {
...
+//
+// @param mbuf
+//   A pointer to the mbuf. The IP header must be at the beginning (offset 0) and
+//   in the first segment of the mbuf.
+// @param edge
+//   If the IPv4 header does not comply with RFC 1812, this is set to one of the error codes:
+//   BAD_LENGTH, BAD_VERSION, or BAD_CHECKSUM. If the header is valid, edge will be left untouched.
+// @return
+//   A pointer to the IP header if it is valid; otherwise, returns NULL.
+static struct rte_ipv4_hdr *get_valid_ip_header(const struct rte_mbuf *mbuf, rte_edge_t *edge) {
```
david-marchand, line 44: <two_cents>
(I hate GH reviews..)

I get that only use of ptype is added, but then when create a separate helper?
Nothing else uses it, and I don't see what code would use it.
Keeping existing code, but just adding ptype consideration inline of current code would reduce the quantity of code churn. </two_cents>

david-marchand: Adding to this, passing &edge is awkward, as this value is linked to a jump label afterward.

maxime-leroy: I can send a patch to the mailing list instead of GitHub if needed.

The goal of moving the code into a new function was to improve readability by avoiding a too big function.

I agree that the edge parameter is not ideal.

There are more changes than just using ptype; the subfunction doesn't help to see that.

I can remove the subfunction if you prefer, or I can split the commit into two: one with fixes and another with the new function.

modules/ip/datapath/ip_input.c
```diff
@@ -25,10 +25,84 @@ enum edges {
...
+			*edge = BAD_CHECKSUM;
+			return NULL;
+		}
+		break;
+	case RTE_MBUF_F_RX_IP_CKSUM_BAD:
+		*edge = BAD_CHECKSUM;
+		return NULL;
+	}
+
+	ptype_l3 = (mbuf->packet_type & RTE_PTYPE_L3_MASK);
```
david-marchand, line 77: Not too sure, but I think ptype needs to be dropped when decapsulating traffic (in ipip node).
Otherwise, "inner" ip header will skip this check.

modules/ip/datapath/ip_input.c
david-marchand: Sorry, I was just grmbling about GH PR, you can keep on using it :-).

I would prefer without the helper indeed.

modules/ip/datapath/ip_input.c
maxime-leroy: Agree using ptype with break ip header validation when decapsulating ip tunnel.

But instead of dropping ptype, it will be more performant to shift inner ptype to outer.

Something like that:

static void update_ptype_inner_to_outer(struct rte_mbuf *mbuf)
{
mbuf->packet_type = (mbuf->packet_type & (RTE_PTYPE_INNER_L2_MASK | RTE_PTYPE_INNER_L3_MASK | RTE_PTYPE_INNER_L4_MASK)) >> RTE_PTYPE_INNER_SHIFT; 
}
With RTE_PTYPE_INNER_SHIFT define like that:
 (__builtin_ctz(RTE_PTYPE_INNER_L2_MASK) - __builtin_ctz(RTE_PTYPE_L2_MASK))

This kind of API should be in DPDK, not grout? 
I can split this commit into 2 parts: one adding the missing ip header check without using ptype, a second one using ptype, fixing ipip case. This second commit could be only merged when update_ptype_inner_to_outer is in DPDK ?

rjarry: > A last question... is ptype always provided by drivers? Is there anything to configure to get all the ptypes grout need?

I don't think there is any guarantee. Which is why the built-in nodes have a rx callback defined in case the drivers do not support it:

https://github.com/DPDK/dpdk/blob/v24.11/lib/node/ethdev_rx.c#L144-L176

maxime-leroy: if a driver doesn't provide ptype, the code will fallback in software validation. Just expect that the driver doesn't provide wrong ptype.

rjarry (approved):
Looks good to me :+1:

PR #181, opened 2025-03-12
Cpu affinity

rjarry:


modules/infra/control/port.c
```diff
@@ -163,6 +163,11 @@ static int port_configure(struct iface_info_port *p) {
 	if ((ret = rte_eth_dev_info_get(p->port_id, &info)) < 0)
 		return errno_log(-ret, "rte_eth_dev_info_get");
 
+	if (strcmp(info.driver_name, "net_tap") == 0) {
+		p->n_txq = RTE_MAX(p->n_txq, p->n_rxq);
+		p->n_rxq = p->n_txq;
+	}
+
 	rxq_size = get_rxq_size(p, &info);
```
maxime-leroy, line 103: Does it mean if number of worker (dataplane core) is inferior to n_rxq. A given worker thread is going to poll N queues of the same port ? Seems port_queue_assign always give rxq to same worked ?

What am I am missing ?

maxime-leroy (approved):

maxime-leroy (approved):
atomic is useless. Look good for me.

modules/infra/control/port.c
rjarry: Yes, that is correct. In some cases, (tap, vhost, virtio) the number of TXQs needs to be equal to the number of RXQs (and same goes both ways).

Since we now impose to allocate a fixed number of TXQs (equal to the number of CPUs in the datapath affinity mask), it also implies that we have a fixed number of RXQs for tap interfaces. See https://github.com/DPDK/grout/pull/181/commits/36eba17fd2ef2c4a85e4c8e763807e43b3a5d4fc

maxime-leroy: not the best, as it means we are going to poll several queue of a tap on the same worker. Example with 2 worker (dp core). One tap port and other physical port. If affinity is set to poll tap port on the first worker. It means that first worker is going to poll q0 of tap, q1 of tap, and q0 of other device. It's unfair. Anyway, as you said tap is mostly used for testing purpose, it's enough for now.

maxime-leroy (commented):
"
1) Use the first available CPU in the grout process affinity for *all*
   control plane threads (including ones started by DPDK).
"

I don't know yet grout enough. But limiting to only one CPUS for control threads seems  not optimal. Providing a list of data cores  in args of grout, using other cores could be an other way.

Does it mean if we don't use taskset, all cores except the first one are used by grout ?

modules/infra/datapath/main_loop.c
```diff
@@ -133,7 +133,7 @@ void *gr_datapath_loop(void *priv) {
 	}
 
 	w->lcore_id = rte_lcore_id();
-	snprintf(name, 15, "gr:loop-c%d", w->cpu_id);
+	snprintf(name, 15, "gr:worker-c%d", w->cpu_id);
```
maxime-leroy, line 136: We have already:    grout,  gr:ctrl-output.
Worker is internal api to grout. For DPDK user, it will be more logical to have dataplane (i.e. gr:dataplane-c%'). 
Top or ps -T $(pidof grout) show the thread name, it maybe be better to use a name explicit for user, not only for grout developer.

maxime-leroy (commented):
cli: sort affinity qmaps by cpu

Could give an output example of command before this commit after ? it make easier the life of the reviewer ;-)

maxime-leroy (approved):
look good

maxime-leroy (approved):
Look good to me.

maxime-leroy (commented):
"1) Destruction of all existing workers to ensure all RXQs are not polled
   anymore. This also ensures reproducible and deterministic RXQ/TXQ
   distribution across different runs.
"

When doing that, it means we lost packets on other ports.
As restarting worker thread will take times.

Restarting all worker seems acceptable when we change number of workers.

It should be possible to just update qmap of every worker, without stopping this ones and restarting this one when we only reconfigure a port.

smoke/affinity_test.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2025 Robin Jarry
+
+. $(dirname $0)/_init.sh
+
+# Start with a single datapath CPU.
+grcli set affinity cpus control 0 datapath 1
```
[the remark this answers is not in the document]
rjarry: In some cases, you may run grout with `isolcpus` and the affinity of the process on startup will not include isolated cores. If you want to move datapath threads on isolated cores, you need to have two independent masks.

rjarry:
> It should be possible to just update qmap of every worker, without stopping this ones and restarting this one when we only reconfigure a port.

I will update the code to only stop threads if they are not used anymore and avoid stopping them to restart them immediately.

modules/infra/datapath/main_loop.c
rjarry: Thread names are limited to 16 characters (including the terminating `\0`) so I wanted something short. If you have more than 100 CPUs, this pattern will overflow:

```python
>>> len("gr:datapath-c666")
16
```

For consistency, we could rename the main thread to have the same `gr:` prefix as the others.

rjarry:
> I don't know yet grout enough. But limiting to only one CPUS for control threads seems not optimal. Providing a list of data cores in args of grout, using other cores could be an other way.

Grout needs to run in different environments, each with its own limitations.

* Bare-metal or in VMs with `isolcpus` started as a systemd service. In this case, the startup CPU affinity should only be used for control plane, and the datapath CPU set should be forced via the configuration.
* Containers with a custom CPU affinity and without the ability to use CPUs outside of that startup affinity.

These two setups are somewhat hard to reconcile. To make things simpler, I've decided to put DPDK control threads (`dpdk-intr`, `dpdk-telemet-v2`) on CPU 0 since we have no way to move them after `rte_eal_init` has finished.

> Does it mean if we don't use taskset, all cores except the first one are used by grout ?

That is correct. If you don't specify a custom affinity, `grout` will reserve the first CPU (0) for control plane operations and will **allow** datapath workers to run on all other CPUs. However, the other CPUs are only provisionned, no thread will be running on them until there are RXQs to poll. If you only have one RXQ to poll, only one of these other CPUs will be used.

maxime-leroy (commented):
" :1) Destruction of all workers running on CPUs outside of the datapath
   affinity mask. Workers that are running on CPUs inside the affinity
   are kept alive but without any RXQ/TXQ assigned. This is required to
   ensure that no RXQ is being polled while distribution happens. This
   also ensures reproducible and deterministic RXQ/TXQ distribution
   across different runs.
"

First point:

It seems odd to run a worker on a non-datapath core, since a worker is inherently a datapath thread.
Why not simply allow workers to run only on datapath cores?
I'm not sure I understand the reason behind having a special case for workers running on non-datapath cores.


Second point:

2)

    Good workers are not destroyed, but bad RX queues aren’t polled during queue reassignment,
    which means packets can be lost during this operation.

I don't fully understand why a worker can’t continue polling RX queues from the old queue map during the transition.

Wouldn’t it be possible to provide the new queue map directly to the worker thread?
For example:

for worker in all_workers:
    new_rxq_map = generate_rxq_map(worker)
    send_async_cmd(UPDATE_QMAP, new_rxq_map, worker)

wait_all_workers_finish(UPDATE_QMAP)

on worker thread:
    UPDATE_QMAP:
        old_rxq_map = worker->rxq_map
        worker->rxq_map = new_rxq_map
        release(old_rxq_map)

A mechanism like this would reduce the disruption during queue map updates.
However, I don’t know enough about GROUT to say whether it supports sending async commands from the control thread to the workers.

We also need to ensure that no RX queue is ever polled by two different workers — one using the old map, and another using the new one.
From what I understand of the commit, the goal is to configure all RX/TX queues across all ports at once,
ensuring that the queue map stays consistent. But in reality, if we’re only adding or removing queues for a new or deleted port,
this guarantee might not hold.

Since we don’t seem to have that guarantee, we probably need a synchronization step where all workers stop polling queues briefly.
This step should be as short as possible to minimize packet loss.

In the end, it's a tradeoff between:

    always generating a deterministic queue map based on the current ports and cores (independent of add/remove order),
    vs

    minimizing packet loss during the update.

Unfortunately, I don’t see a way to achieve both goals perfectly at the same time.

"2) Distribution of every port RXQ to *as many workers as possible* in
   the datapath cpu affinity mask. Port RXQs will be associated with
   workers running on CPUs on the same NUMA socket if possible. If not,
   a warning will be logged and the RXQs will be affected to any
   available CPU.
"

Polling many rxqs can reduce the performance, it could be convenient to have a parameter to limit the number of rx queues.

"
3) Since there is a guarantee that all ports have enough TXQs by
   construction, one TXQ of each port will be affected to each worker.
"
In fact, some nics have a limited number of TXQ, there is no such guarantee.
This could be address a later stage.

modules/infra/control/worker.c
```diff
@@ -277,27 +257,118 @@ int worker_rxq_assign(uint16_t port_id, uint16_t rxq_id, uint16_t cpu_id) {
...
+
+	STAILQ_FOREACH_SAFE (worker, &workers, next, tmp) {
+		if (CPU_ISSET(worker->cpu_id, affinity)) {
+			// Remove all RXQ/TXQ from that worker to have a clean slate.
+			gr_vec_free(worker->rxqs);
+			gr_vec_free(worker->txqs);
+			if ((ret = worker_graph_reload(worker)) < 0) {
+				errno_log(errno, "worker_graph_reload");
+				goto end;
+			}
```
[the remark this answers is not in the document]
rjarry: This is not possible if we want to keep the algorithm trivial. If we recompute all RXQ/TXQ mappings first and then reload every worker one at a time, there is a chance that one RXQ may still be polled by one worker while it is being assigned to another one.

We need to ensure the source worker has released the RXQ before assigning it to another one.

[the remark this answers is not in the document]
rjarry: Mind that this will only happen when changing the datapath CPU affinity which should only happen once at startup.

I am considering adding another mechanism for moving RXQs between threads which should be much quicker.

PR #183, opened 2025-03-17
ip: add unit tests to validate ip header

maxime-leroy:
This patch introduces unit tests to validate the correctness of the IPv4 header processing in the ip_process_input function. The tests check for key properties such as IP version, checksum calculation, packet length, etc ...

### DIFF
--- a/modules/ip/datapath/ip_input.c
+++ b/modules/ip/datapath/ip_input.c
@@ -18,6 +18,19 @@
 #include <rte_mbuf.h>
 #include <rte_mbuf_dyn.h>
 
+#ifdef __GROUT_UNIT_TEST__
+#include <gr_cmocka.h>
+
+// The function is defined as static inline in the original code, so it cannot be wrapped directly
+// using CMocka's function wrapping mechanism.
+static inline void
+__wrap_rte_node_enqueue_x1(struct rte_graph *, struct rte_node *, rte_edge_t next, void *) {
+	check_expected(next);
+}
+
+#define rte_node_enqueue_x1 __wrap_rte_node_enqueue_x1
+#endif
+
 enum edges {
 	FORWARD = 0,
 	OUTPUT,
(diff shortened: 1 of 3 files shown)


maxime-leroy:
I will fix the compilation issue on gcc14. I must mock these two functions to avoid reading random memory during the unit test.

modules/ip/datapath/ip_input.c
```diff
@@ -181,3 +203,134 @@ GR_DROP_REGISTER(ip_input_bad_checksum);
...
+	fake_mbuf->ipv4_hdr.type_of_service = 0;
+	fake_mbuf->ipv4_hdr.packet_id = 1;
+	fake_mbuf->ipv4_hdr.fragment_offset = 0;
+	fake_mbuf->ipv4_hdr.time_to_live = 64;
+	fake_mbuf->ipv4_hdr.next_proto_id = IPPROTO_RAW;
+	fake_mbuf->ipv4_hdr.hdr_checksum = 0;
+	fake_mbuf->ipv4_hdr.src_addr = RTE_IPV4(0, 3, 0, 1);
+	fake_mbuf->ipv4_hdr.dst_addr = RTE_IPV4(1, 9, 8, 6);
+
+	fake_mbuf->mbuf.buf_addr = (void *)&fake_mbuf->ipv4_hdr;
```
rjarry, line 254: is the explicit cast required here?

modules/ip/datapath/ip_input.c
```diff
@@ -30,6 +30,28 @@ enum edges {
...
+//
+// To work around this limitation, we use `#define` to replace calls to these functions
+// with our own wrapped implementations. This ensures that all occurrences of the
+// functions in the tested code are redirected to the wrapped versions during unit testing.
+static void __wrap_rte_node_enqueue_x1(
+	struct rte_graph *graph,
+	struct rte_node *node,
+	rte_edge_t next,
+	void *obj
+);
```
rjarry, line 48: For now we only have one single unit test for datapath code. But thinking ahead, I wonder if we should move this function definition into `gr_graph.h` and force all code to include it instead of `rte_graph{_worker}.h`.

Is `static` required on a declaration? It feels weird.

modules/ip/datapath/ip_input.c
```diff
@@ -30,6 +30,28 @@ enum edges {
...
+static bool __wrap_gr_mbuf_is_traced(struct rte_mbuf *m);
```
rjarry, line 49: This plus the `#define` below should go directly in `gr_mbuf.h` so that it can be reused by other tests in the future.

modules/ip/datapath/ip_input.c
```diff
@@ -30,6 +30,28 @@ enum edges {
 	EDGE_COUNT,
 };
 
+#ifdef __GROUT_UNIT_TEST__
+#include <gr_cmocka.h>
```
rjarry, line 34: Move the include in the unit test block at the bottom.

modules/ip/datapath/ip_input.c
```diff
@@ -181,3 +203,134 @@ GR_DROP_REGISTER(ip_input_bad_checksum);
...
+	fake_mbuf->ipv4_hdr.ihl = 5;
+	fake_mbuf->ipv4_hdr.version = 4;
+	fake_mbuf->ipv4_hdr.total_length = rte_cpu_to_be_16(sizeof(struct rte_ipv4_hdr));
+	fake_mbuf->ipv4_hdr.type_of_service = 0;
+	fake_mbuf->ipv4_hdr.packet_id = 1;
+	fake_mbuf->ipv4_hdr.fragment_offset = 0;
+	fake_mbuf->ipv4_hdr.time_to_live = 64;
+	fake_mbuf->ipv4_hdr.next_proto_id = IPPROTO_RAW;
+	fake_mbuf->ipv4_hdr.hdr_checksum = 0;
+	fake_mbuf->ipv4_hdr.src_addr = RTE_IPV4(0, 3, 0, 1);
```
rjarry, line 251: Too late to say happy birthday. Maybe next year :)

modules/ip/datapath/ip_input.c
```diff
@@ -181,3 +203,134 @@ GR_DROP_REGISTER(ip_input_bad_checksum);
...
+mock_func(void, loopback_input_add_type(rte_be16_t, const char *));
+
+static void __wrap_rte_node_enqueue_x1(
+	struct rte_graph *graph,
+	struct rte_node *node,
+	rte_edge_t next,
+	void *obj
+) {
+	(void)graph;
+	(void)node;
```
rjarry, line 224: I'm not sure if having named arguments is required by cmocka. If not, use use unnamed args:

https://github.com/DPDK/grout/blob/main/CONTRIBUTING.md#unused-function-parameters

rjarry (changes requested):
Hi Maxime, thanks for the patch! I have been wanting to test datapath code for a long time and it is a great starting point.

modules/ip/datapath/ip_input.c
maxime-leroy: In fact, it's not needed anymore with the fake mbuf,  as we have private data. This one are memset, so 	traces->next is NULL. Thus gr_mbuf_is_traced will always returned false. This code can be dropped. Except if we prefer to explicit disable with a wrap function. Let met know.

rjarry: if you are dealing with it, no need to mock it :+1:

modules/ip/datapath/ip_input.c
```diff
@@ -18,6 +18,17 @@
...
 
+#ifdef __GROUT_UNIT_TEST__
+// The function is defined as static inline in the original code, which means they are
+// expanded at compile-time and do not have linkable symbols. This makes them impossible
+// to wrap directly using CMocka's function wrapping mechanism.
+//
+// To work around this limitation, `#define` is used to replace calls to this function
+// with a wrapped implementation (defined in gr_graph.h). This ensures that all occurrences of the
+// function in the tested code are redirected to the wrapped versions during unit testing.
+#define rte_node_enqueue_x1 __wrap_rte_node_enqueue_x1
```
rjarry, line 29: Can you also move this in `gr_graph.h` and replace all direct includes of `rte_graph.h` or `rte_graph_worker.h` with `gr_graph.h` only so that everyone can benefit from the __wrap_rte_node_enqueue_x1 function?

rjarry (changes requested):

modules/ip/datapath/ip_input.c
rjarry: OK, after some discussions with @maxime-leroy we concluded that it is impossible to define this in gr_graph.h otherwise it would replace all occurrences of rte_node_enqueue_x1 in rte_graph.h.

modules/ip/datapath/ip_input.c
```diff
@@ -18,6 +18,17 @@
...
 #include <rte_mbuf_dyn.h>
 
+#ifdef __GROUT_UNIT_TEST__
+// The function is defined as static inline in the original code, which means they are
+// expanded at compile-time and do not have linkable symbols. This makes them impossible
+// to wrap directly using CMocka's function wrapping mechanism.
+//
+// To work around this limitation, `#define` is used to replace calls to this function
+// with a wrapped implementation (defined in gr_graph.h). This ensures that all occurrences of the
+// function in the tested code are redirected to the wrapped versions during unit testing.
```
rjarry, line 28: Can you put this in the commit message?

modules/infra/control/gr_graph.h
```diff
@@ -53,4 +53,12 @@ extern struct node_infos node_infos;
...
 	}
 
+#ifdef __GROUT_UNIT_TEST__
+#include <gr_cmocka.h>
+
+static inline void
+__wrap_rte_node_enqueue_x1(struct rte_graph *, struct rte_node *, rte_edge_t next, void *) {
+	check_expected(next);
+}
+#endif
```
rjarry, line 63: Since the #define needs to be in .c files, having this function here is confusing. Sorry about the double reversal but could you move it back next to `#define rte_node_enqueue_x1 __wrap_rte_node_enqueue_x1` in ip_input.c?

rjarry (changes requested):

rjarry (approved):

PR #187, opened 2025-03-21
rpm: fix errors on recent fedora

rjarry:
Fix the following errors when building on fedora 41:

> error: line 23: Dependency tokens must begin with alpha-numeric, '_'
> or '/': BuildRequires: %toolset

> error: line 5: %package debuginfo: package grout-debuginfo already
> exists


rjarry:
stale

PR #194, opened 2025-04-07
Abstract socket

maxime-leroy:
POC : For frr, dplane plugin must connect on socket unix. But frr don't run with root user.

This series is a dirty poc to support this usecase.


GNUmakefile
```diff
@@ -32,10 +32,9 @@ smoke-tests: all
 .PHONY: update-graph
 update-graph: all
 	$Q set -xe; tmp=`mktemp -d`; \
-	trap "killall grout; wait; rm -rf $$tmp" EXIT; \
-	export GROUT_SOCK_PATH="$$tmp/sock"; \
+	trap "killall grout; wait;" EXIT; \
 	$(BUILDDIR)/grout -t & \
-	socat FILE:/dev/null UNIX-CONNECT:$$GROUT_SOCK_PATH,retry=10 && \
+	socat FILE:/dev/null ABSTRACT-CONNECT:grout.sock,retry=10 && \
```
rjarry, line 37: Could you preserve the ability to force a different name for the abstract socket. Possibly via a `GROUT_SOCK_NAME` variable.

api/gr_api.h
```diff
@@ -25,11 +25,12 @@ struct gr_api_response {
 #define EVENT_TYPE(module, id) (((uint32_t)(0xffff & module) << 16) | (0xffff & id))
 #define EVENT_TYPE_ALL UINT32_C(0xffffffff)
 
-#define GR_DEFAULT_SOCK_PATH "/run/grout.sock"
+#define GR_SOCK_NAME "\0grout.sock"
+#define GR_SOCK_NAME_LEN (sizeof(GR_SOCK_NAME) - 1)
```
rjarry, line 29: Why does this need to start with a NULL byte? It would be more convenient if this was handled in the function where the abstract socket name is actually used.

api/gr_api_client_impl.h
```diff
@@ -39,9 +39,12 @@ struct gr_api_client *gr_api_client_connect(const char *sock_path) {
 		goto err;
 
 	addr.un.sun_family = AF_UNIX;
-	memccpy(addr.un.sun_path, sock_path, 0, sizeof(addr.un.sun_path) - 1);
+	memcpy(addr.un.sun_path, GR_SOCK_NAME, GR_SOCK_NAME_LEN);
```
rjarry, line 42: Here, you can keep:

```c
addr.un.sun_path[0] = '\0';
memccpy(addr.un.sun_path + 1, sock_name, 0, sizeof(addr.un.sun_path) - 2);
```

api/gr_api_client_impl.h
```diff
@@ -39,9 +39,12 @@ struct gr_api_client *gr_api_client_connect(const char *sock_path) {
...
 
-	if (connect(client->sock_fd, &addr.a, sizeof(addr.un)) < 0)
+	if (connect(client->sock_fd,
+		    &addr.a,
+		    offsetof(struct sockaddr_un, sun_path) + GR_SOCK_NAME_LEN)
```
rjarry, line 46: I don't understand why you need to change this?

cli/ec_node_dyn.c
```diff
@@ -30,24 +30,13 @@ static void disconnect_client(void *ptr) {
 
 static struct gr_api_client *connect_client(struct ec_comp *comp) {
 	const struct ec_pnode *pstate = ec_comp_get_cur_pstate(comp);
-	const char *sock_path = NULL;
```
rjarry, line 33: s/sock_path/sock_name/

cli/gr_cli.h
```diff
@@ -95,6 +95,5 @@ typedef int (*ec_node_dyn_comp_t)(
 struct ec_node *ec_node_dyn(const char *id, ec_node_dyn_comp_t cb, void *cb_arg);
 
 #define CLIENT_ATTR "gr_api_client"
-#define SOCK_PATH_ID "gr_api_sock_path"
```
rjarry, line 98: s/PATH/NAME/ and s/path/name/

cli/main.c
```diff
@@ -37,9 +37,6 @@ static void help(void) {
 	puts("  -e, --err-exit             Abort on first error.");
 	puts("  -f PATH, --file PATH       Read commands from file instead of stdin.");
 	puts("  -h, --help                 Show this help message and exit.");
-	puts("  -s PATH, --socket PATH     Path to the control plane API socket.");
-	puts("                             Default: GROUT_SOCK_PATH from env or");
-	printf("                             %s).\n", GR_DEFAULT_SOCK_PATH);
```
rjarry, line 42: Keep this but s/PATH/NAME/ and s/path/name/

main/api.c
```diff
@@ -374,25 +417,9 @@ int api_socket_start(struct event_base *base) {
...
-		if (ret < 0 && errno != ECONNREFUSED)
-			return errno_log(errno, "connect");
-		// remove socket file, and try to bind again
-		if (unlink(addr.un.sun_path) < 0)
-			return errno_log(errno, "unlink");
-		ret = bind(fd, &addr.a, sizeof(addr.un));
-	}
+	memcpy(addr.un.sun_path, GR_SOCK_NAME, GR_SOCK_NAME_LEN);
+
+	ret = bind(fd, &addr.a, offsetof(struct sockaddr_un, sun_path) + GR_SOCK_NAME_LEN);
```
rjarry, line 422: I had a look at https://man7.org/linux/man-pages/man7/unix.7.html and it seems you should rewrite it as:

```c
addr.un.sun_path[0] = '\0';
memccpy(addr.un.sun_path + 1, gr_config.api_sock_name, 0, sizeof(addr.un.sun_path) - 2);

ret = bind(fd, &addr.a, sizeof(addr.un));
```

main/api.c
```diff
@@ -413,7 +440,7 @@ int api_socket_start(struct event_base *base) {
 	// keep a reference for callbacks
 	ev_base = base;
 
-	LOG(INFO, "listening on API socket %s", path);
+	LOG(INFO, "listening on API socket");
```
rjarry, line 443:
```c
LOG(INFO, "listening on API abstract socket %s", gr_config.api_sock_name);
```

main/gr_config.h
```diff
@@ -8,7 +8,6 @@
 #include <stdbool.h>
 
 struct gr_config {
-	const char *api_sock_path;
```
rjarry, line 11: Keep it as `api_sock_name`

main/main.c
```diff
@@ -29,7 +29,7 @@
 // Please keep options/flags in alphabetical order.
 
 static void usage(const char *prog) {
-	printf("Usage: %s [-h] [-L <type>:<lvl>] [-p] [-s <path>] [-t] [-T <regexp>]\n", prog);
```
rjarry, line 32: Keep it with s/path/name/

smoke/_init.sh
```diff
@@ -3,7 +3,7 @@
 
 set -e
 
-if [ -S "$GROUT_SOCK_PATH" ]; then
+if socat -T 0.1 - ABSTRACT-CONNECT:grout.sock < /dev/null; then
```
rjarry, line 6: Keep the test for `$GROUT_SOCK_NAME` here:

```bash
if [ -n "$GROUT_SOCK_NAME" ]; then
```

smoke/_init.sh
```diff
@@ -74,7 +74,7 @@ set -x
 if [ "$run_grout" = true ]; then
 	taskset -c 0,1 grout -tvx &
 fi
-socat FILE:/dev/null UNIX-CONNECT:$GROUT_SOCK_PATH,retry=10
+socat FILE:/dev/null ABSTRACT-CONNECT:grout.sock,retry=10
```
rjarry, line 77:
```bash
socat FILE:/dev/null ABSTRACT-CONNECT:$GROUT_SOCK_NAME,retry=10
```

smoke/_init.sh
rjarry: You also need to change line 58 as follows:

```bash
if [ "$run_grout" = true ]; then
	export GROUT_SOCK_NAME="grout-$run_id.sock"
fi
```

main/main.c
```diff
@@ -79,9 +76,6 @@ static int parse_args(int argc, char **argv) {
 
 	opterr = 0; // disable getopt default error reporting
 
-	gr_config.api_sock_path = getenv("GROUT_SOCK_PATH");
-	if (gr_config.api_sock_path == NULL)
-		gr_config.api_sock_path = GR_DEFAULT_SOCK_PATH;
```
rjarry, line 84: Keep this but s/PATH/NAME/ and s/path/name/

main/main.c
```diff
@@ -100,9 +94,6 @@ static int parse_args(int argc, char **argv) {
 		case 'S':
 			gr_config.log_syslog = true;
 			break;
-		case 's':
-			gr_config.api_sock_path = optarg;
-			break;
```
rjarry, line 105: Keep:

```c
		case 's':
			gr_config.api_sock_name = optarg;
			break;
```

main/api.c
```diff
@@ -329,6 +330,43 @@ static void read_cb(evutil_socket_t sock, short what, void * /*priv*/) {
...
+	cap_t caps;
+
+	if (getsockopt(fd, SOL_SOCKET, SO_PEERCRED, &cred, &len) == -1) {
+		LOG(ERR, "getsockopt(SO_PEERCRED): %s", strerror(errno));
+		return -1;
+	}
+
+	// root user, allowed
+	if (cred.uid == 0)
+		return 0;
```
rjarry, line 346: This shortcut is not required.

main/api.c
```diff
@@ -329,6 +330,43 @@ static void read_cb(evutil_socket_t sock, short what, void * /*priv*/) {
...
+
+	// check if peer process has CAP_NET_ADMIN
+	caps = cap_get_pid(cred.pid);
+	if (caps == NULL) {
+		LOG(ERR, "cap_get_pid: %s", strerror(errno));
+		return -1;
+	}
```
rjarry, line 353: replace with:

```c
return errno_log(errno, "cap_get_pid");
```

main/api.c
```diff
@@ -329,6 +330,43 @@ static void read_cb(evutil_socket_t sock, short what, void * /*priv*/) {
...
+
+	if (cap_get_flag(caps, CAP_NET_ADMIN, CAP_EFFECTIVE, &has_cap) == -1) {
+		LOG(ERR, "cap_get_flags: %s", strerror(errno));
+		cap_free(caps);
+		return -1;
+	}
```
rjarry, line 359: replace with:

```c
cap_free(caps);
return errno_log(errno, "cap_get_flags");
```

main/api.c
```diff
@@ -329,6 +330,43 @@ static void read_cb(evutil_socket_t sock, short what, void * /*priv*/) {
...
+static int check_permission(int fd) {
+	cap_flag_value_t has_cap = CAP_CLEAR;
+	struct ucred cred;
+	socklen_t len = sizeof(cred);
+	cap_t caps;
+
+	if (getsockopt(fd, SOL_SOCKET, SO_PEERCRED, &cred, &len) == -1) {
+		LOG(ERR, "getsockopt(SO_PEERCRED): %s", strerror(errno));
+		return -1;
+	}
```
rjarry, line 342: replace with:

```c
return errno_log(errno, "getsockopt(SO_PEERCRED)");
```

main/api.c
```diff
@@ -329,6 +330,43 @@ static void read_cb(evutil_socket_t sock, short what, void * /*priv*/) {
 		event_free_finalize(0, ev, finalize_fd);
 }
 
+static int check_permission(int fd) {
+	cap_flag_value_t has_cap = CAP_CLEAR;
+	struct ucred cred;
+	socklen_t len = sizeof(cred);
+	cap_t caps;
```
rjarry, line 337: reverse xmas tree please

main/api.c
```diff
@@ -329,6 +330,43 @@ static void read_cb(evutil_socket_t sock, short what, void * /*priv*/) {
...
+	}
+
+	if (cap_get_flag(caps, CAP_NET_ADMIN, CAP_EFFECTIVE, &has_cap) == -1) {
+		LOG(ERR, "cap_get_flags: %s", strerror(errno));
+		cap_free(caps);
+		return -1;
+	}
+	cap_free(caps);
+
+	if (has_cap != CAP_SET) {
```
rjarry, line 362: I would prefer if the check was more dynamic.

* If the `grout` process is started as `root` or as any user that has the `CAP_NET_ADMIN` capability, then require that all clients have that same capability.
* Else, if `grout` is started as a regular user without `CAP_NET_ADMIN`, then require that client processes are started with the same `gid` than the `grout` process.

main/api.c
```diff
@@ -329,6 +330,43 @@ static void read_cb(evutil_socket_t sock, short what, void * /*priv*/) {
...
+		LOG(ERR, "unauthorized peer (pid=%d, uid=%d)", cred.pid, cred.uid);
+		return -1;
```
rjarry, line 364:
```c
return errno_set(EPERM);
```

main/api.c
```diff
@@ -346,6 +384,12 @@ static void listen_cb(evutil_socket_t sock, short what, void * /*priv*/) {
 		return;
 	}
 
+	/* check if peer is root or has CAP_NET_ADMIN */
+	if (check_permission(fd) != 0) {
+		close(fd);
+		return;
```
rjarry, line 390: For consistency with the rest of the code:

```c
if (check_client_permissions(fd) < 0) {
```

Can we return an explicit error to the client?

rjarry (changes requested):
Hi Maxime,

this seems like a good improvement. Especially not needing to clean the file up is nice.

I'd like to preserve the ability to change the abstract socket name and also to allow more flexible permission checking when `grout` is not started as `root`.

api/gr_api_client_impl.h
maxime-leroy: an abstract socket address is distinguished (from a
  pathname socket) by the fact that sun_path[0] is a null
  byte ('\0').  The socket's address in this namespace is
  given by the additional bytes in sun_path that are covered
  by the specified length of the address structure.  (Null
  bytes in the name have no special significance.

from man unix 7.

see: https://stackoverflow.com/questions/11640826/can-not-connect-to-linux-abstract-unix-socket

main/api.c
maxime-leroy: need to dig into grout to understand API msg. I will check that.

rjarry:
Thinking again about my previous message.

Could you rearrange the commits so that the check for permissions is performed in the first commit using `SO_PEERCRED` to ensure that `cred.uid == 0 || cred.uid == geteuid()`.

That way you will get the exact same security level than before since `grout` was creating the socket with `-rw-------` permissions.

Then, in the second commit, you can introduce `libcap` and allow **any user** to connect as long as they have `CAP_NET_ADMIN`.

maxime-leroy:
> Thinking again about my previous message.
> 
> Could you rearrange the commits so that the check for permissions is performed in the first commit using `SO_PEERCRED` to ensure that `cred.uid == 0 || cred.uid == geteuid()`.
> 
> That way you will get the exact same security level than before since `grout` was creating the socket with `-rw-------` permissions.
> 
> Then, in the second commit, you can introduce `libcap` and allow **any user** to connect as long as they have `CAP_NET_ADMIN`.

Same user instead of same group ?

rjarry:
> Same user instead of same group ?

Yes, I was looking at the code and I was wrong, we are **not** changing the default permissions of the socket after `bind()`. So it depends on `umask` which is most likely the default `0022`. So the permissions of the socket are always: `-rw-r--r--` and the owner/group is the one from the user running `grout`.

So, in order to have no functional/security change after the first commit is applied, we should ensure that the peer user ID is equal to `geteuid()` (or equal to 0).

rjarry (approved):

PR #195, opened 2025-04-08
cli: use scols_line_sprintf if available

rjarry:
Since libsmartcols 2.41, scols_line_sprintf is available and does not need to be implemented by grout. Add the required meson code to avoid building our local version when the symbol is available in the library.

Link: https://github.com/util-linux/util-linux/commit/f0d0817c37931b504a7


maxime-leroy (approved):
Look good. Thanks

PR #202, opened 2025-04-17
modules/ip[6]: fix uninitialized nexthop in "addr del" event

maxime-leroy:
When an IP address is removed, the RIB cleanup process (rib4/6_cleanup) resets associated nexthop memory before the addr del event is sent. As a result, the event includes a nexthop structure with all fields set to zero, even though a valid nexthop was previously in use.

To reproduce:

grcli show events
> addr add: iface[1] 9.6.6.7
> route del: 9.6.6.7/24 via 9.6.6.7 origin link
> nh del: iface 1 vrf 0 9.6.6.7 d2:f0:0c:bc:a4:10
> addr del: iface[0] 0x5d279f4eb1ac

Or from another instance:

grout# add ip address 9.6.6.7/24 iface p4
grout# del ip address 9.6.6.7/24 iface p4

Fix this by emitting the addr del event before the nexthop is cleaned up, ensuring the event still carries valid nexthop information.

### DIFF
--- a/modules/ip/control/address.c
+++ b/modules/ip/control/address.c
@@ -123,10 +123,11 @@ static struct api_out addr_del(const void *request, void ** /*response*/) {
 		return api_out(ENOENT, 0);
 	}
 
+	gr_event_push(IP_EVENT_ADDR_DEL, nh);
+
 	rib4_cleanup(nh);
 
 	gr_vec_del(addrs->nh, i);
-	gr_event_push(IP_EVENT_ADDR_DEL, nh);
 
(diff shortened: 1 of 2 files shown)


maxime-leroy:
To check where is the best place to send the event in IPv6.

rjarry (changes requested):
Hi Maxime,

this looks good to me.

can you just change the commit title to:

`ip,ip6: fix uninitialized nexthop in "addr del" event`

Thanks!

PR #207, opened 2025-04-29
Socket group

maxime-leroy:
RFC about socket permission for frr


rjarry:
We should probably allow commit titles that look like `Revert ".*"`

rjarry (changes requested):
I would prefer two different options:

`-o, --socket-owner <user>:<group>`

And:

`-p, --socket-permissions <0660>`

But that will require more parsing. Maybe there are functions that do this already in the libc.

rjarry:
I have made a few changes, let me know if that suits you.

rjarry (approved):
NB: I have not retested after my cosmetic changes.

maxime-leroy:
> I have made a few changes, let me know if that suits you.

Look better. Thanks. I will retest it on Friday.

maxime-leroy:
Working fine with last version.

PR #209, opened 2025-04-30
Ifindex rt

maxime-leroy:
Add capacity to configure route with interface and gateway.

For now, it's just a WIP series. 

We need that for FRR plugin.

### DIFF
--- a/modules/ip/api/gr_ip4.h
+++ b/modules/ip/api/gr_ip4.h
@@ -19,6 +19,7 @@ struct gr_ip4_ifaddr {
 struct gr_ip4_route {
 	struct ip4_net dest;
 	ip4_addr_t nh;
+	uint16_t iface_id;
 	uint16_t vrf_id;
 	gr_rt_origin_t origin;
 };
@@ -65,6 +66,7 @@ struct gr_ip4_route_add_req {
 	uint16_t vrf_id;
 	struct ip4_net dest;
 	ip4_addr_t nh;
+	uint16_t iface_id;
 	gr_rt_origin_t origin;
 	uint8_t exist_ok;
 };
--- a/modules/ip/cli/route.c
+++ b/modules/ip/cli/route.c
@@ -16,12 +16,30 @@
 #include <errno.h>
 
 static cmd_status_t route4_add(const struct gr_api_client *c, const struct ec_pnode *p) {
-	struct gr_ip4_route_add_req req = {.exist_ok = true, .origin = GR_RT_ORIGIN_USER};
+	struct gr_ip4_route_add_req req = {
+		.exist_ok = true, .origin = GR_RT_ORIGIN_USER, .iface_id = GR_IFACE_ID_UNDEF
+	};
+	int ret;
 
 	if (arg_ip4_net(p, "DEST", &req.dest, true) < 0)
 		return CMD_ERROR;
-	if (arg_ip4(p, "NH", &req.nh) < 0)
+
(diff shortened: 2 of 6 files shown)


christophefontaine:
I'm not sure rejecting unconditionally an existing route with a different next hop is the correct behavior. 

From a FIB perspective, I agree that *now* we should only have 1 next hop for any given route. 

Yet, from a RIB perspective, we can have different next hop for the same route.

We would need a differentiator (eg: metric, link speed, ...) to decide the best route to render from the ROB to the FIB. 

With dynamic routing, won't we have multiple gateways for a single route ?

vjardin:
Let's keep it simple for the current design. Mid term, next hop group shall be considered in order to support multiple gateways and interfaces with their updates. Dynamic routing will benefit of nhg for such Grout datapath.

Rejecting helps to avoid a miss configuration for the time being and I understand that it is enough for the routing/FIB manager.

maxime-leroy:
> I'm not sure rejecting unconditionally an existing route with a different next hop is the correct behavior. 
It's the current behavior. My commit doesn't change that. 
> 
> From a FIB perspective, I agree that *now* we should only have 1 next hop for any given route. 
> 
> Yet, from a RIB perspective, we can have different next hop for the same route.
> 
> We would need a differentiator (eg: metric, link speed, ...) to decide the best route to render from the ROB to the FIB

There is different topic like ecmp,  route selection from rib. 
 
But it's not the topic of this commit.
> 
> With dynamic routing, won't we have multiple gateways for a single route ?

When you do:
add ip route  10.0.0.0/24 via 1.1.1.1
add ip route  10.0.0.0/24  via 2.2.2.2
-> there is no error
show ip route -> only show the first route

My patch doesn't change anything. Grout don't support multiple next nop in the rib now. It just return an error when we tried to do that. 

When multipe nexthop will be support, this commit is not need. Until that i believe it useful to get an error when a route is not added in the rib.

christophefontaine:
Understood, looks nice !

maxime-leroy:
> Yes sure, can you push --force without the last commit? I'll merge the PR ASAP.

https://github.com/DPDK/grout/pull/225 just create this one. Keep this one to update last patch by using gr_nexthop.

And frr plugin must be updated to use it also (if merged)

maxime-leroy:
> possible to merge the two fixes ?
> 
> Last commit must be updated to use gr_nexthop @rjarry ? (i will do it later)

//! Nexthop structure exposed to the API.
struct gr_nexthop {
        gr_nh_flags_t flags; //!< bit mask of GR_NH_F_*
        uint16_t vrf_id; //!< L3 VRF domain
        uint16_t iface_id; //!< interface associated with this nexthop
        struct rte_ether_addr mac; //!< link-layer address
        union {
                struct {
                } addr;
                ip4_addr_t ipv4;
                struct rte_ipv6_addr ipv6;
        };
        gr_nh_type_t type; //!< nexthop type
        uint8_t prefixlen; //!< only has meaning with GR_NH_F_LOCAL
        uint16_t held_pkts; //!< number of packets waiting for resolution
        clock_t last_reply; //!< timestamp when last update was received
};
There are many fields like prefixlen, held_pkts, last_reply,flags etc. I don't why we should expose this fields on gr_ip4_route ? there are no even use by nh4_add. Not sure why we expose internal fields in the gr_api.

maxime-leroy:
It's done anymore. It has implemented in other PR.

PR #210, opened 2025-05-03
port: allow attaching already probed devices

rjarry:
At startup, we pass -a 0000:00:00.0 to rte_eal_init() to disable automatic PCI probing. However, this only affects the PCI bus devices. Other buses (e.g. FSMLC) are not affected by this argument and are always probed in rte_eal_init().

When configuring a port with such a device, we get an error from the API saying that the port is already attached to another interface:

	grout# add interface port dpni2 devargs fslmc:dpni.2
	error: File exists

Instead of iterating over the DPDK probed devices, look at grout ports to determine whether a port matching this devargs value is already configured.

If we cannot find one, check if the device has already been probed and only probe it if it was not.

### DIFF
--- a/modules/infra/control/port.c
+++ b/modules/infra/control/port.c
@@ -326,21 +326,31 @@ static int iface_port_init(struct iface *iface, const void *api_info) {
 	const struct gr_iface_info_port *api = api_info;
 	uint16_t port_id = RTE_MAX_ETHPORTS;
 	struct rte_dev_iterator iterator;
+	const struct iface *i;
 	struct gr_iface conf;
 	int ret;
 
-	RTE_ETH_FOREACH_MATCHING_DEV(port_id, api->devargs, &iterator) {
-		rte_eth_iterator_cleanup(&iterator);
-		return errno_set(EEXIST);
+	i = NULL;
+	while ((i = iface_next(GR_IFACE_TYPE_PORT, i)) != NULL) {
+		const struct iface_info_port *p = (const struct iface_info_port *)i->info;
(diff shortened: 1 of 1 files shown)


vjardin:
Tested with:
```
ls-delete dpni.2
ls-delete dpni.3
/usr/sbin/qoriq-dpdk/dpaa2/dynamic_dpl.sh dpmac.1 dpmac.2
DPRC=dprc.2 grout -v --log-level="eal.*,debug" --log-level="pmd.*,debug" --log-level="bus.fslmc,debug"
```

```
$ grcli
Welcome to the graph router CLI version custom.
Use ? for help and <tab> for command completion.
grout# add interface port dpni2 devargs fslmc:dpni.2 up mode l3
Created interface 1

grout# show interface
NAME   ID  FLAGS       MODE  DOMAIN  TYPE  INFO
dpni2  1   up running  L3    0       port  devargs=fslmc:dpni.2 mac=00:04:9f:05:79:e0
```

```
DEBUG: GROUT: listen_cb: new connection fd=203
DEBUG: GROUT: read_cb: fd=203 id=1 req_type=0xcafe1981 len=128
DEBUG: GROUT: send_response: fd=203 for_id=1 len=0 status=0 Success
DEBUG: GROUT: read_cb: fd=203 id=2 req_type=0xacdc0001 len=332
DEBUG: DPAA2_NET: dpaa2_dev_info_get(): >>
DEBUG: GROUT: gr_pktmbuf_pool_get: allocate mempool mbuf_0:0 reserved 16383 (size 65535)
DEBUG: DPAA2_NET: dpaa2_dev_info_get(): >>
DEBUG: DPAA2_NET: dpaa2_eth_dev_configure(): >>
```

PR #211, opened 2025-05-04
meson: generate grout.pc - pkg-config

vjardin:
A pc/pkg-config file is needed in order to provide helpers for integration. For instance, it can be used when FRRouting is crosscompiled in order to avoid an unsafe path such as /usr/include/grout.

Note: if the build directory has no .git metadata, the current version probe yields the placeholder string 'Custom' rather than a tag such as 'vA.B.C'. So, currently, I did copy version.


vjardin:
The purpose is to avoid such dirty path https://github.com/FRRouting/frr/pull/18750/commits/cdf857813e7004c403362d4fa56138b94fcd9960#diff-5e5d29cd44d2aaa72ecac8b3c8c04605a31042986152ffd3f6ac82236a90d578R279

meson.build
```diff
@@ -149,3 +149,14 @@ if cmocka_dep.found()
     test(name, executable(name, kwargs: t), suite: 'unit')
   endforeach
 endif
+
+# generate grout.pc
+grout_pkg = import('pkgconfig')
+grout_pkg.generate(
+  name        : 'grout',
+  description : 'Helpers for grout integration and management',
+  version     : 'v0.9.1',
```
rjarry, line 158: Please remove this to let meson use the project version.

There is a way to override the version from the environment already by exporting `GROUT_VERSION` in the build system. I use it for Fedora: https://src.fedoraproject.org/rpms/grout/blob/rawhide/f/grout.spec#_66

meson.build
```diff
@@ -149,3 +149,14 @@ if cmocka_dep.found()
     test(name, executable(name, kwargs: t), suite: 'unit')
   endforeach
 endif
+
+# generate grout.pc
+grout_pkg = import('pkgconfig')
+grout_pkg.generate(
```
rjarry, line 155: nit pick:

```meson
import('pkgconfig').generate(
```

And do not align colons.

meson.build
```diff
@@ -149,3 +149,14 @@ if cmocka_dep.found()
...
+  name        : 'grout',
+  description : 'Helpers for grout integration and management',
+  version     : 'v0.9.1',
+  url         : 'https://github.com/DPDK/grout',
+  subdirs     : 'grout',  # provides -I${includedir}/grout
```
rjarry, line 160: This should follow what is set in `install_headers()` line 129 (i.e. no subdir)

meson.build
```diff
@@ -149,3 +149,14 @@ if cmocka_dep.found()
...
+  # library   : []        # none
```
rjarry, line 161: No need for this line.

meson.build
```diff
@@ -149,3 +149,14 @@ if cmocka_dep.found()
     test(name, executable(name, kwargs: t), suite: 'unit')
   endforeach
 endif
+
+# generate grout.pc
+grout_pkg = import('pkgconfig')
+grout_pkg.generate(
+  name        : 'grout',
+  description : 'Helpers for grout integration and management',
```
rjarry, line 157: I'm not sure about this description. Do we need it?

rjarry (changes requested):

meson.build
vjardin: pkgconfig.generate: if a library is not passed as a positional argument, the 'description' keyword argument is required.

meson.build
vjardin: lt leads to the value "custom" on Buildroot when some local folders are used for development because of https://gitlab.com/buildroot.org/buildroot/-/blob/master/package/pkg-generic.mk#L565

It could work if I can get the following accepted by Buildroot maintainers:
```
--- a/package/grout/grout.mk
+++ b/package/grout/grout.mk
@@ -4,13 +4,15 @@
 #
 ################################################################################

-GROUT_VERSION = 0.9.1
+GROUT_REPOVERSION = 0.9.1
+GROUT_VERSION = $(GROUT_REPOVERSION)
 GROUT_SITE = $(call github,DPDK,grout,v$(GROUT_VERSION))
 GROUT_LICENSE = BSD-3-Clause
 GROUT_LICENSE_FILES = LICENSE
+GROUT_INSTALL_STAGING = YES

 # Avoid using buildroot commit hash
-GROUT_CONF_ENV = GROUT_VERSION=$(GROUT_VERSION)
+GROUT_CONF_ENV = GROUT_VERSION=$(GROUT_REPOVERSION)
```

My update will assume that it'll be accepted.

vjardin: let's try with 5f3ea83e188e0b6b955fa8c505120ad5c856250c

meson.build
```diff
@@ -128,6 +128,15 @@ grcli_exe = executable(
 
 install_headers(api_headers)
 
+# generate grout.pc for those headers
+import('pkgconfig').generate(
+  name: 'grout',
+  description: 'Helpers for grout integration and management',
+  version: meson.project_version(),
```
rjarry, line 135: This is redundant. It is already the default value when version is not specified.

meson.build
```diff
@@ -128,6 +128,15 @@ grcli_exe = executable(
...
+  url: 'https://github.com/DPDK/grout',
+  subdirs: 'grout',  # provides -I${includedir}/grout
```
rjarry, line 137: This is incorrect. The headers are installed flat in /usr/include.

rjarry (changes requested):

meson.build
rjarry: ack

meson.build
rjarry: I am not sure to follow the problem. If you are using a local grout repo, isn't it a good thing that the version *is* actually `custom` and not `0.9.1`?

vjardin:
Bottom line: a grout.pc/pkg-config is not needed. configure.ac can use `AC_CHECK_HEADER([gr_api.h]` instead of `PKG_CHECK_MODULES([GROUT], [grout >= v0.9.2]`

Let's drop this pull request.

PR #213, opened 2025-05-05
iface: use base structures for copying to api

rjarry:
There is no need to copy individual fields. Use the base structures to copy common fields from the internal and public API messages.

Fixes: c5f94a9386ae ("iface: reuse public structures")


maxime-leroy (approved):
Fix missing speed not with 'gr-cli show interface name XX'

PR #214, opened 2025-05-05
cli: autocomplete support on fslmc CPUs (QorIQ)

vjardin:
Add support for parsing and adding to the autocomplete the contents of the fslmc bus (for example for a LX2160 QorIQ CPU).

Example:
```
 grout# add interface port dpni.2 devargs <?>
 fslmc:dpni.2       net_null           net_tap            net_tun
 net_vhost          net_virtio_user
 grout# add interface port dpni.2 devargs f<tab>
 grout# add interface port dpni.2 devargs fslmc:dpni.2
```

The bus cases (PCI, fslmc, vdev) have not been redesigned.

### DIFF
--- a/docs/grcli.1.md
+++ b/docs/grcli.1.md
@@ -48,6 +48,17 @@ Print version and exit.
 
 Print executed commands.
 
+# ENVIRONMENT
+
+#### **DPRC**
+
+Set the DPRC - Datapath Resource Container: This value should match the
+one used by DPDK during the scan of the fslmc bus. It is recommended to
+set this on any NXP QorIQ targets. This serves as the entry point for
+grcli to enable autocompletion of fslmc devices manageable by grout.
+While grcli can configure grout without this environment setting,
+autocompletion of the devargs will not be available.
+
 # SEE ALSO
 
 **grout**(8)
--- a/cli/ec_node_devargs.c
+++ b/cli/ec_node_devargs.c
@@ -22,6 +22,7 @@ ec_node_devargs_parse(const struct ec_node *, struct ec_pnode *, const struct ec
 	return 1;
 }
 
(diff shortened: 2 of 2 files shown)


docs/grcli.1.md
```diff
@@ -48,6 +48,17 @@ Print version and exit.
...
+
+#### **DPRC**
+
+Set the DPRC - Datapath Resource Container: This value should match the
+one used by DPDK during the scan of the fslmc bus. It is recommended to
+set this on any NXP QorIQ targets. This serves as the entry point for
+grcli to enable autocompletion of fslmc devices manageable by grout.
+While grcli can configure grout without this environment setting,
+autocompletion of the devargs will not be available.
+
```
rjarry, line 61: How does rte_eal_init() manage to detect the devices without this variable set?

Could you do the same in grcli?

vjardin: nope, I cannot, any DPDK application have to be started with DPRC env set.
```
static int
rte_fslmc_scan(void)
{
        int ret;
        char fslmc_dirpath[PATH_MAX];
        DIR *dir;
        struct dirent *entry;
        static int process_once;
        int groupid;
        char *group_name;

        if (process_once) {
                DPAA2_BUS_DEBUG("Fslmc bus already scanned. Not rescanning");
                return 0;
        }
        process_once = 1;

        /* Now we only support single group per process.*/
        group_name = getenv("DPRC");
        if (!group_name) {
                DPAA2_BUS_DEBUG("DPAA2: DPRC not available");
                ret = -EINVAL;
                goto scan_fail;
        }
       ...
```

rjarry: Ok that's very sad :(

PR #215, opened 2025-05-05
port: add port description

vjardin:
According to the RFC1213 a port description can be up to 255 bytes.


maxime-leroy:
Why is it specific to port ? Not to any interfaces type ?

Could be just allocated dynamically as it's optional field and to not to have a hardcorded limit for this description?

rjarry (changes requested):
I would also prefer if the description could be attached to all interfaces. Not just ports.

PR #222, opened 2025-05-20
main: use locale from environment

rjarry:
Change explicit setlocale(LC_CTYPE, "C.UTF-8") to setlocale(LC_ALL, "") to initialize the wide char encoding to whatever locale is enabled in the environment.

Rename utf8_check to charset_check to avoid confusion since the encoding may now be a non-UTF-8 charset.


vjardin:
LGTM. It'll simplify the integration.

Issue #115, opened 2024-12-23
Discard NDP packets containing null length option


david-marchand:
Ref. NDP rfc:

```
[4.6]  Option Formats
...
      Length         8-bit unsigned integer.  The length of the option
                     (including the type and length fields) in units of
                     8 octets.  The value 0 is invalid.  Nodes MUST
                     silently discard an ND packet that contains an
                     option with length zero.
```

david-marchand:
A first part of the fix would be to break in the loop parsing options.
Yet, there are parts of grout that continue dealing with this NDP packet where the rfc says to discard the whole packet.

PR #228, opened 2025-05-29
Discard NDP packets containing null length option

spk-hebbar:
According to https://datatracker.ietf.org/doc/html/rfc4861#section-4.6,  neighbor discovery packets with option length 0 is not dropped, currently it is not so. Added fix for that along with unit tests for NDP NA input process. 

Fixes: https://github.com/DPDK/grout/issues/115

### DIFF
--- a/modules/ip6/datapath/meson.build
+++ b/modules/ip6/datapath/meson.build
@@ -24,5 +24,9 @@ tests += [
   {
     'sources': files('ip6_input.c'),
     'link_args': [],
-  }
+  },
+  {
+    'sources': files('ndp_na_input.c'),
+    'link_args': [],
+  },
 ]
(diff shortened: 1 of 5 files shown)


modules/ip6/datapath/meson.build
```diff
@@ -22,7 +22,18 @@ inc += include_directories('.')
...
+      ),
+    'link_args': [],
+  }
+]
+
+tests += [
+  {
+    'sources': files(
+      'ndp_na_input.c',
+      ),
```
rjarry, line 36: This is a bit awkward. Can you do a single list?

```meson
tests += [
  {
    'sources': files('ip6_input.c'),
    'link_args': [],
  },
  {
    'sources': files('ndp_na_input.c'),
    'link_args': [],
  },
]
```

spk-hebbar:
@rjarry Thanks for the review. I have updated the commits according to your comments.

rjarry (approved):

PR #231, opened 2025-06-05
ip: fix crash when receiving broadcast packets

rjarry:
When receiving a broadcast or multicast ethernet frame (e.g. a DHCP request), the ip_output_mbuf_data->iface field wrongfully is reset to NULL.

This leads to a crash in ip_input_local when accessing iface->vrf_id:

```
Thread 6 "grout:w20" received signal SIGSEGV, Segmentation fault. 0x00000000004fe924 in ip_input_local_process
	(graph=0x7ff7e95f5dc0, node=0x7ff7e95fb140, objs=0x7ff7f39ad740, nb_objs=1)
	at ../modules/ip/datapath/ip_local.c:48
48                              data->vrf_id = iface->vrf_id;
```

Make sure that iface is always preserved to its original value. Homogenize the code with ip6_input.

Fixes: 2e7d5f6f1aef ("ip: clarify usage of mpriv in ip_input_process")
Reported-by: Spoorthi K <spk@redhat.com>

### DIFF
--- a/modules/ip/datapath/ip_input.c
+++ b/modules/ip/datapath/ip_input.c
@@ -41,10 +41,11 @@ ip_input_process(struct rte_graph *graph, struct rte_node *node, void **objs, ui
 	uint16_t i;
 
 	for (i = 0; i < nb_objs; i++) {
-		iface = NULL;
-		nh = NULL;
 		mbuf = objs[i];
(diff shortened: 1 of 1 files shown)


spk-hebbar (approved):
Thanks, LGTM

PR #244, opened 2025-06-25
rx,tx: enable multi segment to support higher mtu

rjarry:
In order to support receiving packets larger than 2048 bytes, we need to enable multi-segment rx and tx offload in the drivers.

Enable RTE_ETH_RX_OFFLOAD_SCATTER and RTE_ETH_TX_OFFLOAD_MULTI_SEGS in the requested offload flags when configuring DPDK ports.

Update the basic IP forward test to check we are indeed receiving 5000 bytes packets with a configured MTU of 9000 bytes.

Suggested-by: David Marchand <david.marchand@redhat.com>

### DIFF
--- a/modules/infra/control/port.c
+++ b/modules/infra/control/port.c
@@ -72,7 +72,12 @@ static struct rte_eth_conf default_port_config = {
 		},
 	},
 	.rxmode = {
-		.offloads = RTE_ETH_RX_OFFLOAD_CHECKSUM | RTE_ETH_RX_OFFLOAD_VLAN,
+		.offloads = RTE_ETH_RX_OFFLOAD_CHECKSUM
+			| RTE_ETH_RX_OFFLOAD_VLAN
+			| RTE_ETH_RX_OFFLOAD_SCATTER,
+	},
+	.txmode = {
+		.offloads = RTE_ETH_TX_OFFLOAD_MULTI_SEGS,
 	},
 };
 
@@ -123,6 +128,7 @@ int port_configure(struct iface_info_port *p, uint16_t n_txq_min) {
 	else
 		conf.rxmode.mq_mode = RTE_ETH_MQ_RX_RSS;
 	conf.rxmode.offloads &= info.rx_offload_capa;
+	conf.txmode.offloads &= info.tx_offload_capa;
 	if (info.dev_flags != NULL && *info.dev_flags & RTE_ETH_DEV_INTR_LSC) {
 		conf.intr_conf.lsc = 1;
 	}
--- a/smoke/ip_forward_test.sh
+++ b/smoke/ip_forward_test.sh
@@ -7,8 +7,8 @@
 p0=${run_id}0
 p1=${run_id}1
 
-grcli add interface port $p0 devargs net_tap0,iface=$p0 mac f0:0d:ac:dc:00:00
-grcli add interface port $p1 devargs net_tap1,iface=$p1 mac f0:0d:ac:dc:00:01
+grcli add interface port $p0 devargs net_tap0,iface=$p0 mac f0:0d:ac:dc:00:00 mtu 9000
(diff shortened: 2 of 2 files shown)


modules/infra/control/port.c
```diff
@@ -95,8 +95,9 @@ int port_configure(struct iface_info_port *p, uint16_t n_txq_min) {
 	if ((ret = rte_eth_dev_info_get(p->port_id, &info)) < 0)
 		return errno_log(-ret, "rte_eth_dev_info_get");
 
-	if (strcmp(info.driver_name, "net_tap") == 0) {
-		p->n_txq = RTE_MAX(p->n_txq, p->n_rxq);
```
david-marchand, line 99: This patch has nothing to do with the series about larger packets.

In any case, I am not following.
Why remove this part above?

With the previous code, if a user requests more rxq, this would result in unneeded txq, but I don't see a problem with this.

david-marchand (commented):
LGTM.

modules/infra/control/port.c
rjarry: I'll drop this patch for now.

maxime-leroy (commented):
We should be very careful on that.

Most of code doing operation on mbuf supposes there is only one segment, data are contiguous.

Most of time network header will be in first segment, but anyway to have safe code if we active mutli segs, you can not anymore present that all data are in the first segments.

maxime-leroy:
> Use of rte_pktmbuf_read would be the safest, but blindly adding checks would impact performance.
> 
> If multiseg scares you, another approach is to allow bigger frames in the packet pool (for which there is no configuration knob atm).
 rte_pktmbuf_read is just for read. You need to modify the packet, when grout doing memmove in a pktmbuf data without knowing if data is spread on several segment, it's just a dangerous.

Of course, we can use jumbo mbuf to support jumbo packet. The side effect is increased memory consumption.

Else it's to support multi segs. In this case, doing the check about multi segments when writing a mbuf or reading this one must be done. Of course, the side effect is an extra cost for cpu. 

Nothing is free, you must choose if you want to increase memory cpu cycles or cpu cost to support jumbo frames.

Just enabling the feature without wondering if some operations done on muf are safe or not with multiseg is the  perfect way to get crash.

I am against this commit, It should not have been  merged without answering properly about my point.

Prove me that all operations are safe in grout with multi segments or else fix it.

rjarry:
> Just enabling the feature without wondering if some operations done on muf are safe or not with multiseg is the perfect way to get crash.
> 
> I am against this commit, It should not have been merged without answering properly about my point.
> 
> Prove me that all operations are safe in grout with multi segments or else fix it.

Operations aren't safe for sure.

I don't have any specific preference. The only thing I want is for jumbo MTUs to work.

We could use bigger MBUFs. Is there any downside to this besides high memory consumption?

maxime-leroy:
> > Just enabling the feature without wondering if some operations done on muf are safe or not with multiseg is the perfect way to get crash.
> > I am against this commit, It should not have been merged without answering properly about my point.
> > Prove me that all operations are safe in grout with multi segments or else fix it.
> 
> Operations aren't safe for sure.
> 
> I don't have any specific preference. The only thing I want is for jumbo MTUs to work.
> 
> We could use bigger MBUFs. Is there any downside to this besides high memory consumption?

This commits will drop performance (at least > 10% au doigt mouillé). Don't have time to do the test. But by enabling scatter in rx and multi segment in tx, you use less performance function of driver for RX/TX (i.e. vector vs non vector).

By using jumbo mbuf, you don't have the problem. I don't see any sideeffect, except the increaze of memory consumption. We should do the math to see if it's acceptable or not. Maybe an option --jumbo could be added at grout. To not force any one to have these big mbuf.

At least, if we need multi segments, we should enable only when mtu is > mbuf headroom. To keep the best performance for standard mtu.

PR #245, opened 2025-06-26
Revert "rx,tx: enable multi segment to support higher mtu"

maxime-leroy:
Reverts DPDK/grout#244

To support jumbo mbuf:
- we can increase mbuf size, the side effect is increased memory consumption.
-  To support multi segs. In this case, doing the check about multi segments when writing a mbuf or reading this one must be done. Of course, the side effect is an extra cost for cpu.

Choose 2 have been done without fixing/checking that operation done on mbuf are safe or not with multiseg. It just the perfect way to get  segfault.

Some work has done on my side in ip_input and ip6_input to check that data are on the first segment. I have added some unit tests. But it's not enough at all to support multi segments.

### DIFF
--- a/modules/infra/control/port.c
+++ b/modules/infra/control/port.c
@@ -72,12 +72,7 @@ static struct rte_eth_conf default_port_config = {
 		},
 	},
 	.rxmode = {
-		.offloads = RTE_ETH_RX_OFFLOAD_CHECKSUM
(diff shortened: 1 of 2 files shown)


rjarry:
Closing this. The revert was done in https://github.com/DPDK/grout/pull/246

PR #246, opened 2025-06-27
Max mtu

rjarry:


smoke/_init.sh
```diff
@@ -81,7 +81,6 @@ if [ -n "${builddir}" ]; then
 	export PATH=$builddir:$PATH
 fi
 
-grout_extra_options=""
 if [ "$test_frr" = true ] && [ "$run_frr" = true ]; then
 	chmod 0755 $tmp # to access on tmp
 	grout_extra_options+="-m 0666"
```
maxime-leroy, line 86: grout_extra_options="" is not set, no sure if going  to work when doing that grout_extra_options+="-m 0666"

main/dpdk.c
```diff
@@ -143,7 +143,7 @@ int dpdk_init(void) {
 		gr_vec_add(eal_args, "--no-shconf");
 		gr_vec_add(eal_args, "--no-huge");
 		gr_vec_add(eal_args, "-m");
-		gr_vec_add(eal_args, "2048");
+		gr_vec_add(eal_args, "8192");
 	} else {
```
maxime-leroy, line 147: Too much memory ? it will not work on my laptop, neither on buildroot CI (only 4G of memory, most of test are limited to 1G for the VM).

At the end, in CI, the tap driver is tested, not sure it has many value to test it with jumbo frames....

maxime-leroy (changes requested):

modules/ipip/control.c
```diff
@@ -102,7 +102,7 @@ static int iface_ipip_fini(struct iface *iface) {
 static int iface_ipip_init(struct iface *iface, const void *api_info) {
 	const struct gr_iface conf = {
 		.flags = iface->flags,
-		.mtu = iface->mtu,
+		.mtu = iface->mtu ?: 1480,
```
maxime-leroy, line 105: In which case, iface->mtu is zero ?

rjarry:
> ICMP too big must also sent for IPV4 with DF bit set.
> 
> We don't have any function to send packets ?

We do but at the moment, we don't support neither fragmentation nor sending ICMP too big packets.

This will be implemented later.

modules/ipip/control.c
rjarry: Unless users have specified a value for mtu in the initial interface creation message, it will be 0. It was OK until now since we'll drop packets that are bigger than output interface MTU.

I chose to initialize it to 1480 which seems like a good default.

PR #261, opened 2025-07-15
smoke: fail RS test early when missing rdisc6

david-marchand:
Check for rdisc6 command availability so that a user gets an explicit failure reason rather than a "command not found" lost into many other debug logs.

Fixes: 2e45b6c9844f ("ip6: answer to router sollicit message")

Note: this is not really important, more like a PR to test if sourcery is (correctly) setup.


DPDK-admin:
@coderabbitai configuration

maxime-leroy:
Too many IA to review commit, please choose one ... Too garbage to handle.

PR #267, opened 2025-07-30
route: do not duplicate NH

christophefontaine:
A duplicate gateway NH is created when we add 2 routes with the same gateway IP, such as the following config:
```
add ip route 16.0.0.0/24 via 192.168.210.2
add ip route 48.0.0.0/24 via 192.168.211.2
add ip route 16.0.1.0/24 via 192.168.210.2
add ip route 48.0.1.0/24 via 192.168.211.2
```

```
grout# show nexthop
VRF  ID  TYPE  FAMILY  IP                         MAC                IFACE       STATE      FLAGS              ORIGIN
0        L3    IPv6    fe80::48df:37ff:fe68:d710  48:df:37:68:d7:10  cv0	 reachable  static local link  link
0        L3    IPv6    fe80::48df:37ff:fe68:d718  48:df:37:68:d7:18  cv1	 reachable  static local link  link
0        L3    IPv4    192.168.210.1              48:df:37:68:d7:10  cv0	 reachable  static local link  link
0        L3    IPv4    192.168.211.1              48:df:37:68:d7:18  cv1	 reachable  static local link  link
0        L3    IPv4    192.168.210.2              ?                  cv0	 new        gateway            user
0        L3    IPv4    192.168.211.2              ?                  cv1	 new        gateway            user
0        L3    IPv4    192.168.210.2              90:e2:ba:14:2f:8c  cv0	 reachable  gateway            user
0        L3    IPv4    192.168.211.2              90:e2:ba:14:2f:8d  cv1	 reachable  gateway            user
```
Lookup for that nexthop, and if it doesn't exist, create it.

## Summary by Sourcery

Bug Fixes:
- Check for existing nexthop by address before falling back to RIB lookup in IPv4 and IPv6 route addition to prevent duplicate gateway next hops.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the logic for adding IPv4 and IPv6 routes to handle nexthop lookups more robustly, providing clearer error handling when a nexthop cannot be found or a gateway is unreachable.

PR #272, opened 2025-08-01
ip6: define a preferred address

christophefontaine:
addr6_get_preferred costs a lot as it loops over the configured addresses. If we have only 1 configured address in addition to the local (fe80::) address, use it immediately.

With the following configuration:
```
grout# show ip6 address
IFACE  ADDRESS
cv0    fe80::48df:37ff:fe68:d710/64
cv0    fc00:210::1/64
cv1    fe80::48df:37ff:fe68:d718/64
cv1    fc00:211::1/64
```
Before:
```
grout# show stats software
NODE                     CALLS   PACKETS  PKTS/CALL  CYCLES/CALL  CYCLES/PKT
sr6_output              126208  32309242      256.0      47265.6       184.6
```
After:
```
grout# show stats software
NODE                      CALLS   PACKETS  PKTS/CALL  CYCLES/CALL CYCLES/PKT
sr6_output               290585  35604326      122.5      15925.6      130.0
```

## Summary by Sourcery

Cache and quickly return the single non-link-local IPv6 address per interface when only one global address is configured to improve addr6_get_preferred performance

Enhancements:
- Introduce a per-interface cache for the preferred global IPv6 address to avoid repeated loops in addr6_get_preferred
- Automatically update the preferred-address cache when IPv6 addresses are added or removed

### DIFF
--- a/modules/ip6/control/address.c
+++ b/modules/ip6/control/address.c
@@ -42,12 +42,33 @@ struct hoplist *addr6_get_all(uint16_t iface_id) {
 	return addrs;
 }
 
+static struct nexthop *preferred_addr[MAX_IFACES];
+
+static void addr6_update_preferred(uint16_t iface_id) {
+	struct hoplist *addrs = addr6_get_all(iface_id);
+	struct nexthop *pref = NULL, *nh;
+
+	if (addrs == NULL)
+		return;
+
+	if (gr_vec_len(addrs->nh) == 2) {
+		gr_vec_foreach (nh, addrs->nh) {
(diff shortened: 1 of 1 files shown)


maxime-leroy (changes requested):
It's show a real performance problem but solution is only worked with one ipv6 configured, so it's not a good solution.
in frr, there are using tun src api, see: https://github.com/DPDK/grout/pull/267
example:
 ip -6 sr tunsrc show
tunsrc addr ::

PR #283, opened 2025-08-07
ip/ip6: fix display nexthop

christophefontaine:
With the ability to set a nexthop with an ID, we now support a IPv4 gateway for a IPv6 network, and the other way around.

Fix the command 'show ip(6) route' to print the address with the proper format.

```
grout# show ip route
VRF  DESTINATION       NEXT_HOP       ORIGIN  ID  NEXT_HOP_VRF
0    16.0.1.0/24       fc00:210::2    user    11  0

```
Fixes: a1e4087b ("all: add support for gateway-less nexthops")


maxime-leroy (commented):
Having an example of output with show ip route before commit will help to understand the issue better.

maxime-leroy (approved):
look good.

PR #298, opened 2025-08-16
treewide: refactor gr_nexthop

christophefontaine:
Huge refactor of gr_nexthop structure to split
the different fields according to the nexthop type.

Only did it for the exercise of doing it, to understand if this is relevant.

### DIFF
--- a/modules/infra/api/gr_nexthop.h
+++ b/modules/infra/api/gr_nexthop.h
@@ -75,20 +75,26 @@ typedef enum : uint8_t {
 //! Nexthop structure exposed to the API.
 struct gr_nexthop {
 	gr_nh_type_t type;
-	addr_family_t af;
 	gr_nh_state_t state;
 	gr_nh_flags_t flags; //!< bit mask of GR_NH_F_*
+	gr_nh_origin_t origin;
 	uint32_t nh_id; //!< Arbitrary ID set by user. Zero means "unset".
 	uint16_t vrf_id; //!< L3 VRF domain
-	uint16_t iface_id; //!< interface associated with this nexthop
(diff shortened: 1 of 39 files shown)


modules/infra/api/gr_nexthop.h
```diff
@@ -75,20 +75,26 @@ typedef enum : uint8_t {
...
+			uint8_t prefixlen; //!< only has meaning with GR_NH_F_LOCAL
+			union {
+				struct {
+				} addr;
+				ip4_addr_t ipv4;
+				struct rte_ipv6_addr ipv6;
+			};
+		} l3;
+		struct { // GR_NH_T_BLACKHOLE
+		} bh;
```
christophefontaine, line 97: @vjardin This is the kind of refactor for gr_nexthop, to have per type related data.
Seeing how this impacts the whole tree, I don't know if I'm ready to push this kind of huge modification :) 

@rjarry this is only the file that matter, everything else was to comply...

If we use only unnamed structs, this wouldn't impact as much the whole tree, and we would still group the fields by nexthop type.

```
union {
		struct { // GR_NH_T_L3
			addr_family_t af;
			uint16_t iface_id; //!< interface associated with this nexthop
			struct rte_ether_addr mac; //!< link-layer address
			uint8_t prefixlen; //!< only has meaning with GR_NH_F_LOCAL
			union {
				struct {
				} addr;
				ip4_addr_t ipv4;
				struct rte_ipv6_addr ipv6;
			};
		} ;
		struct { // GR_NH_T_BLACKHOLE
		} ;
		struct { // GR_NH_T_GROUP
		} ;
};
```

PR #318, opened 2025-09-24
telemetry

rjarry:
- **nexthop: only expose statistics per type**
- **rib,rib6: expose all origins in telemetry stats**


modules/infra/api/gr_nexthop.h
```diff
@@ -70,7 +70,6 @@ typedef enum : uint8_t {
 	GR_NH_ORIGIN_OPENFABRIC = 197, // (RTPROT_OPENFABIC from zebra)
 	GR_NH_ORIGIN_SRTE = 198, // (RTPROT_SRTE from zebra)
 	GR_NH_ORIGIN_INTERNAL = 255, //!< Reserved for internal use by grout.
-#define _GR_NH_ORIGIN_COUNT (GR_NH_ORIGIN_INTERNAL + 1)
```
christophefontaine, line 74: maybe keeping _GR_NH_ORIGIN_COUNT (eventually defined to UINT8_MAX +1 to match the underlying enum type) will avoid the magic number "256" below ?

PR #328, opened 2025-10-01
overhead stats

rjarry:
* **stats: add overhead cycles accounting**
* **main-loop: move reset stats to ensure consistency**
* **stats: always expose idle and overhead stats**

## Summary by CodeRabbit

- New Features
  - Added an “Overhead” runtime stat to show non-node processing time.
  - Per-worker loop and cycle counters added for finer performance insight.

- Refactor
  - Stats now report packets and batches (replacing previous object/call metrics); UI/CLI columns updated accordingly.
  - Unified zero-value stat filtering for consistent display.
  - Standardized periodic housekeeping to stabilize stats windows and aggregation.

### DIFF
--- a/modules/infra/api/gr_infra.h
+++ b/modules/infra/api/gr_infra.h
@@ -125,8 +125,8 @@ struct gr_port_rxq_map {
 struct gr_infra_stat {
 	char name[64];
 	uint64_t topo_order;
-	uint64_t objs;
-	uint64_t calls;
+	uint64_t packets;
+	uint64_t batches;
 	uint64_t cycles;
 };
 
--- a/modules/infra/api/stats.c
+++ b/modules/infra/api/stats.c
@@ -44,13 +44,13 @@ static gr_vec struct gr_infra_stat *graph_stats(uint16_t cpu_id) {
 			const char *name = rte_node_id_to_name(n->node_id);
 			s = find_stat(stats, name);
 			if (s != NULL) {
-				s->objs += n->objs;
-				s->calls += n->calls;
+				s->packets += n->packets;
+				s->batches += n->batches;
 				s->cycles += n->cycles;
 			} else {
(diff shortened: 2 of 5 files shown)


rjarry:
> * how can we display stats per worker ?

We could do that with an extra argument to filter stats of a single worker.

> * on the stats ordering, it would be great to have an option to order them by order of processing the the graph.
>   This would simplify a lot the performance comparison between 2 versions, and just makes sense for the user.
>   WDYT ?

That is more tricky but could be doable by adding a "topo_order" field to all stats objects.

I have some reservations about this new "overhead" stats though. There may be something I missed, or the overhead is really much higher than the actual graph processing.

This is when grout is mostly idle:

```
NODE        CALLS  PACKETS  PKTS/CALL  CYCLES/CALL  CYCLES/PKT
idle        31398        0        0.0     148387.7         0.0
overhead  8037888        0        0.0       4568.2         0.0
```

And this is when it is processing packets at almost maximum capacity:

```
NODE            CALLS   PACKETS  PKTS/CALL  CYCLES/CALL  CYCLES/PKT
overhead      3496192         0        0.0      65477.1         0.0 <---- this does not make any sense
ip_input       333214  44544257      133.7       5060.4        37.9
eth_output     333191  44541058      133.7       2982.2        22.3
port_tx-p0q1   160183  22272448      139.0       2895.8        20.8
ip_output      333214  44544257      133.7       2783.2        20.8
ip_forward     333214  44544257      133.7       1434.7        10.7
port_output    333191  44541058      133.7       1278.0         9.6
port_rx-p0q0  1887440  44537220       23.6        691.4        29.3
port_rx-p1q0  1781504  22272448       12.5        463.1        37.0
eth_input     3495936  44541058       12.7        444.3        34.9
idle                0         0        0.0          0.0         0.0
```

rjarry:
I finally got it right...

PR #334, opened 2025-10-06
rx,tx: use one node per port queue

rjarry:
- **dpdk: update to v25.07**
- **rx,tx: initialize node context when creating graph**
- **vec: add gr_strvec_free**
- **vec: add gr_vec_extend**
- **infra: rename enum to name functions**
- **infra: rename {rx,tx}.c with a port_ prefix**
- **port: pass whole structure to plug,unplug**
- **eth_output: make output dependent on interface type**
- **rx,tx: use one node per port queue**


rjarry:
@coderabbitai full review

PR #350, opened 2025-10-20
Add GARP and unsolicited NDP support

rjarry:
- **nexthop: add constant for local address flags**
- **ip: send gratuitous arp requests when iface comes up**
- **ip6: allow forging ndp na replies**
- **ip6: send neigh advert for every address when iface is up**
- **port: only unplug/stop if necessary**
- **trace: remove function name from logs**
- **trace: format garp packets**
- **trace: display neigh advert solicited flag**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved interface-up handling to trigger ARP and IPv6 neighbor advertisements
  * Refined gratuitous ARP handling to avoid updating probe/timing state
  * Adjusted local nexthop handling during deletion to correct flag checks

* **Performance Improvements**
  * More granular port reconfiguration to avoid unnecessary full restarts
  * IPv6 neighbor advertisements sent immediately rather than deferred

* **Refactor**
  * Standardized logging format and backend across datapath components

### DIFF
--- a/modules/infra/datapath/trace.c
+++ b/modules/infra/datapath/trace.c
@@ -266,7 +266,7 @@ int trace_icmp6_format(char *buf, size_t len, const struct icmp6 *icmp6, size_t 
 		const struct icmp6_neigh_advert *na = PAYLOAD(icmp6);
 		payload_len -= sizeof(*na);
 		inet_ntop(AF_INET6, &na->target, dst, sizeof(dst));
-		SAFE_BUF(snprintf, len, "neigh advert %s is at", dst);
+		SAFE_BUF(snprintf, len, "neigh advert %s is at (solicited=%u)", dst, na->solicited);
 		opt = PAYLOAD(na);
 		break;
 	}


modules/infra/api/nexthop.c
```diff
@@ -84,7 +83,8 @@ static struct api_out nh_del(const void *request, struct api_ctx *) {
 
 	if (nh->type == GR_NH_T_L3) {
 		struct nexthop_info_l3 *l3 = nexthop_info_l3(nh);
-		if ((l3->flags & addr_flags) == addr_flags || nh->origin == GR_NH_ORIGIN_LINK)
+		if ((l3->flags & NH_LOCAL_ADDR_FLAGS) == NH_LOCAL_ADDR_FLAGS
```
christophefontaine: There is a behavior diff, as we check on `GR_NH_F_LINK` as well. Is that intentional ?

rjarry: That is a fair question and I don't recall why we would check GR_NH_F_LINK separately. I was probably a mistake.

modules/ip6/datapath/ndp_na_output.c
```diff
@@ -19,13 +19,38 @@ enum {
...
 
+struct advertise_context {
+	const struct nexthop *local;
+	const struct nexthop *remote;
+	const struct iface *iface;
+};
+
+static control_input_t na_output;
+
+int nh6_advertise(const struct nexthop *local, const struct nexthop *remote) {
```
christophefontaine: isn't there an overlap on the RA messages ?
I'm wondering if we can't factorise a bit of code here, as we may need to send a RA in addition to a NA message.

PR #351, opened 2025-10-20
cli: remove redundant interface type name field

rjarry:
The `cli_iface_type` structure duplicated the interface type name which is already available via `gr_iface_type_name()`. Remove the name field and use `gr_iface_type_name()` directly to avoid maintaining duplicate information.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized interface type name handling to use a centralized helper function instead of direct struct field access, consolidating type identification logic across multiple components.


aharivel (approved):
LGTM

PR #353, opened 2025-10-21
Dpdk force static link

maxime-leroy:
## Summary by CodeRabbit

* **Chores**
  * Updated DPDK dependency to 24.11.3 (minor patch release).
  * Introduced a new build option to enable static linking against DPDK and applied it across build and packaging configurations (build system, Debian packaging, RPM spec, and makefile invocations) to produce consistent static-linked artifacts.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -54,7 +54,8 @@ clean:
 install: $(BUILDDIR)/build.ninja
 	$Q meson install -C $(BUILDDIR) --skip-subprojects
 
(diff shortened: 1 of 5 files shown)


rjarry (changes requested):
Static linking should be disabled by default. You can add a meson option to enable it if you want.

Also, do not change the minimum required version in `meson.build` since it does not add any crucial features. You can simply bump the version in `subprojects/dpdk.wrap`.

PR #359, opened 2025-10-23
smoke: cleanup and refactor tests

rjarry:
This series refactors the smoke test infrastructure to reduce code duplication and speed up execution.

Port creation and namespace setup are now centralized in helper functions. Repetitive code blocks have been replaced with loops. MAC addresses are no longer hardcoded, and link-local addresses are queried dynamically when needed.

Test execution time has been reduced by lowering sleep intervals from 1s to 200ms and FRR startup timeout from 15s to 5s.

## Summary by CodeRabbit

* **Tests**
  * Streamlined smoke test framework with automatic per-port provisioning and MAC isolation.
  * Moved tests to explicit per-namespace topologies for clearer, consistent host contexts.
  * Added automatic link-local IPv6 resolution for more reliable neighbor checks.
* **Improvements**
  * Improved robustness with retry-based readiness checks, faster backoff intervals (0.2s), centralized failure reporting, and accumulated cleanup steps.

### DIFF
--- a/smoke/cross_vrf_forward_frr_test.sh
+++ b/smoke/cross_vrf_forward_frr_test.sh
@@ -9,12 +9,13 @@ create_interface p1 vrf 2
 
 for n in 0 1; do
 	p=p$n
-	netns_add $p
-	ip link set $p netns $p
-	ip -n $p link set $p up
-	ip -n $p addr add 172.16.$n.2/24 dev $p
-	ip -n $p addr add 16.$n.0.1/16 dev lo
-	ip -n $p route add default via 172.16.$n.1
+	ns=n$n
+	netns_add $ns
+	ip link set $p netns $ns
+	ip -n $ns link set $p up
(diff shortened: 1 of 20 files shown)


smoke/cross_vrf_forward_frr_test.sh
```diff
@@ -7,11 +7,12 @@
 p0=${run_id}0
 p1=${run_id}1
 
-create_interface $p0 vrf 1
-create_interface $p1 vrf 2
-
 for n in 0 1; do
 	p=$run_id$n
+	vrf=$((n + 1))
```
christophefontaine: Not fan of that commit: we're now interleaving linux and grout interface creation and configuration.
While the previous way of configuring was more verbose, it feels a bit cleaner as we configure 1 part, and then the other part.

rjarry: Fair point. I removed it.

christophefontaine (approved):

PR #377, opened 2025-11-03
icmp6: replace endian-dependent bitfields with flag enums

rjarry:
The `icmp6_router_advert` and `icmp6_neigh_advert` structures used bitfields to represent flag bits. The order of bitfield members in memory is implementation-defined and requires `#ifdef BYTE_ORDER` guards to ensure correctness across different architectures.

Replace bitfields with `uint8_t` flag members and define flag constants as typed enums using the `GR_BIT8()` macro. This approach is portable, explicit, and does not rely on compiler-specific behavior.

Update all code accessing these flags to use bitwise operations instead of direct member access.

## Summary by CodeRabbit

* **Refactor**
  * IPv6 router and neighbor advertisement flag handling consolidated: individual boolean flag fields replaced by a single explicit flags field for clarity and consistency across components.
  * Internal flag sourcing consolidated where diagnostics/trace output report solicited status.
  * No change to observable packet semantics; behavior preserved.

### DIFF
--- a/modules/infra/datapath/gr_icmp6.h
+++ b/modules/infra/datapath/gr_icmp6.h
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <gr_bitops.h>
 #include <gr_macro.h>
 
 #include <rte_byteorder.h>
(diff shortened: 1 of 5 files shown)


modules/infra/datapath/gr_icmp6.h
```diff
@@ -74,14 +75,15 @@ struct icmp6_router_solicit {
...
+	ICMP6_RA_F_OTHER_CONFIG = GR_BIT8(1),
+} icmp6_ra_flags_t;
+
 // ICMP6_TYPE_ROUTER_ADVERT
 struct icmp6_router_advert {
 	uint8_t cur_hoplim;
-#if BYTE_ORDER == BIG_ENDIAN
-	uint8_t managed_addr : 1, other_config : 1, __unused_flags : 6;
-#else
-	uint8_t __unused_flags : 6, managed_addr : 1, other_config : 1;
```
maxime-leroy: Does it mean we should not allow any bit-fields in grout ?

maxime-leroy (approved):

modules/infra/datapath/gr_icmp6.h
christophefontaine: We're trading a known complexity in 1 header with a more complex code to write, read and debug in all other files.
I'm not sure if I'm a fan of that.
@rjarry Do we have a test that this works on big endian arch ? Otherwise, I'd refrain merging that.

rjarry: The C standard only mentions **byte** ordering. Not **bit** ordering. Compilers can store them in any order they want.

PR #381, opened 2025-11-04
port: prevent disabling forced promiscuous and allmulti modes

rjarry:
When a port MAC filter table reaches its hardware limit or if the driver does not support individual address filtering, the code falls back to enabling promisc or allmulti mode to ensure all required traffic is still received. However, other code paths that manage these modes were unaware of this fallback mechanism and could inadvertently disable them, breaking MAC address filtering.

Introduce two new interface state flags to track when promisc or allmulti modes have been force-enabled due to MAC filter overflow/unsupported. The port_promisc_set() and port_allmulti_set() functions now check these forced states and refuse to modify the hardware state when the mode is required for filtering.

Update the CLI to distinguish between user-requested and filter-forced modes by displaying "promisc(forced)" or "allmulti(forced)" when applicable.

## Summary by CodeRabbit

* **New Features**
  - Added indicators for interfaces forced into promiscuous or all-multicast modes and display of "promisc(fixed)" / "allmulti(fixed)".

* **Behavioral Fixes**
  - Prevent changing promiscuous/all-multicast when marked fixed; ensures MAC list reinstallation keeps hardware and software state consistent.

* **Chores**
  - Exposed port-level control APIs for MAC add/remove and promisc/allmulti control; updated coverage/build settings.

* **Tests**
  - Added unit tests covering MAC add/remove and forced-mode flows.

### DIFF
--- a/modules/infra/control/meson.build
+++ b/modules/infra/control/meson.build
@@ -41,5 +41,19 @@ tests += [
       '-Wl,--wrap=rte_pktmbuf_pool_create',
       '-Wl,--wrap=rte_zmalloc',
     ],
-  }
+  },
+  {
+    'sources': files('port_test.c', 'port.c'),
+    'link_args': [
+      '-Wl,--wrap=rte_eth_allmulticast_disable',
+      '-Wl,--wrap=rte_eth_allmulticast_enable',
+      '-Wl,--wrap=rte_eth_allmulticast_get',
+      '-Wl,--wrap=rte_eth_dev_mac_addr_add',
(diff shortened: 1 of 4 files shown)


maxime-leroy:
What about ospf in FRR ? 

These ones use a socket to listen multicast groupe. The cp port will not be enough in this case.

rjarry:
> What about ospf in FRR ?
> 
> These ones use a socket to listen multicast groupe. The cp port will not be enough in this case.

I don't understand what you mean.

christophefontaine (approved):

PR #387, opened 2025-11-06
bond: add lacp support

rjarry:
Implement Link Aggregation Control Protocol (802.3ad/802.1AX) for bond interfaces. LACP provides dynamic link aggregation with automatic member discovery, synchronization, and failover based on protocol negotiation rather than simple link state monitoring.

Add a new GR_BOND_MODE_LACP operational mode alongside the existing active-backup mode. LACP mode includes three load balancing algorithms for distributing egress traffic across active members:

- rss: Reuse hardware RSS hash when available for zero-cost balancing
- l2: Hash based on destination MAC address and VLAN tag
- l3+l4: Hash based on IP addresses and TCP/UDP port numbers

The control plane implementation handles LACP protocol state machines including receiving and validating LACP PDUs, tracking partner state, managing synchronization and collecting/distributing flags, and handling fast/slow periodic timers and timeout detection. A periodic timer runs every second to send LACP PDUs and detect partner timeouts based on the negotiated timeout intervals.

NB: port numbers in LACP PDUs are 1-based rather than 0-based to ensure interoperability with switches that reject port ID zero as invalid.

Members transition to active state when LACP negotiation succeeds and both sides report synchronized state.

Active members are added to a 256-entry redirection table that maps hash values to member indices. The table is populated by distributing active member IDs in round-robin fashion across all entries, ensuring even load distribution while maintaining flow consistency.

For example, with 3 active members, entry 0 maps to member 0, entry 1 to member 1, entry 2 to member 2, entry 3 back to member 0, and so on. Traffic is directed by computing a hash from packet headers, taking modulo 256 to get a table index, and using the stored member ID at that position. This provides deterministic per-flow member selection with minimal disruption when members are added or removed.

The bond interface transitions to running state when at least one member becomes active.

## Summary by CodeRabbit

* **New Features**
  * Added LACP (802.3ad) bond mode with selectable balancing algorithms (rss, l2, l3+l4) and per‑packet load distribution; full LACP send/receive and active-member management.
  * Packet traces now include detailed LACP PDU state.

* **CLI**
  * Configure LACP mode and balancing algorithm with mode-specific validation and help text.

* **Tests**
  * Added smoke test validating LACP bond setup, member sync, and connectivity.


modules/infra/datapath/bond_output.c
```diff
@@ -26,13 +33,154 @@ static int bond_trace_format(char *buf, size_t len, const void *data, size_t /*d
...
+			tuple.l2.vlan_id = 0;
+			eth_type = eth->ether_type;
+			l3_offset = sizeof(*eth);
+		}
+		switch (eth_type) {
+		case RTE_BE16(RTE_ETHER_TYPE_IPV4): {
+			l3.ip4 = rte_pktmbuf_mtod_offset(m, struct rte_ipv4_hdr *, l3_offset);
+			tuple.v4.src_addr = l3.ip4->src_addr;
+			tuple.v4.dst_addr = l3.ip4->dst_addr;
+			switch (l3.ip4->next_proto_id) {
```
aharivel: Problem:
  1. Fragmented IP packets won't have L4 headers in non-first fragments - accessing them causes undefined behavior
  2. No bounds check that L4 header is within packet data

Maybe:
```
  // Check for fragmentation
  if (l3.ip4->fragment_offset & RTE_BE16(RTE_IPV4_HDR_MF_FLAG | RTE_IPV4_HDR_OFFSET_MASK)) {
      // Fragmented packet - only hash on L3
      tuple.v4.sport = 0;
      tuple.v4.dport = 0;
      len = sizeof(tuple.v4);
      break;
  }

  case IPPROTO_UDP:
      if (unlikely(m->data_len < l3_offset + rte_ipv4_hdr_len(l3.ip4) + sizeof(*l4.udp))) {
          tuple.v4.sport = 0;
          tuple.v4.dport = 0;
      } else {
          l4.udp = rte_pktmbuf_mtod_offset(...);
          tuple.v4.sport = l4.udp->src_port;
          tuple.v4.dport = l4.udp->dst_port;
      }
```

modules/infra/datapath/bond_output.c
```diff
@@ -26,13 +33,154 @@ static int bond_trace_format(char *buf, size_t len, const void *data, size_t /*d
...
+				tuple.v4.dport = l4.tcp->dst_port;
+				break;
+			default:
+				tuple.v4.sport = 0;
+				tuple.v4.dport = 0;
+			}
+			len = sizeof(tuple.v4);
+			break;
+		}
+		case RTE_BE16(RTE_ETHER_TYPE_IPV6): {
```
aharivel: Problem:
  1. IPv6 extension headers are not handled - l3.ip6->proto might be a routing header or fragment header, not the final protocol
  2. No bounds checks before accessing IPv6 and L4 headers
  3. IPv6 fragments (when present in fragment extension header) can't have L4 in non-first fragments

Maybe:
```
  case RTE_BE16(RTE_ETHER_TYPE_IPV6): {
      if (unlikely(m->data_len < l3_offset + sizeof(*l3.ip6))) {
          len = sizeof(tuple.l2);
          break;
      }
      l3.ip6 = rte_pktmbuf_mtod_offset(m, const struct rte_ipv6_hdr *, l3_offset);
      tuple.v6.src_addr = l3.ip6->src_addr;
      tuple.v6.dst_addr = l3.ip6->dst_addr;

      // TODO: Handle IPv6 extension headers properly
      // For now, only hash on L3 if extension headers present
      uint8_t next_proto = l3.ip6->proto;
      if (next_proto != IPPROTO_UDP && next_proto != IPPROTO_TCP) {
          tuple.v6.sport = 0;
          tuple.v6.dport = 0;
          len = sizeof(tuple.v6);
          break;
      }
```

modules/infra/datapath/bond_output.c
```diff
@@ -26,13 +33,154 @@ static int bond_trace_format(char *buf, size_t len, const void *data, size_t /*d
...
+			tuple.l2.vlan_id = 0;
+		}
+		len = sizeof(tuple.l2);
+		break;
+	case GR_BOND_ALGO_RSS:
+		if (m->ol_flags & RTE_MBUF_F_RX_RSS_HASH) {
+			hash = m->hash.rss;
+			goto out;
+		}
+		// fallthrough
```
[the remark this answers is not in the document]
christophefontaine: We may want to sync with a common pattern across the code base, `[[fallthrough]]` is accepted as well for C23, and shorter.
Just to have a consistent code base everywhere.

modules/infra/datapath/bond_output.c
christophefontaine: This raises a much bigger concern: the subsequent fragments may go through a different link from the 1st fragment.
For fragmented packets, what about using the `ip.src + ip.dst + identification` fields ?

rjarry: That is a good point. Note that if RSS is enabled, it will be used anyway. Regardless of fragmentation.

PR #390, opened 2025-11-07
smoke: simplify bgp test

rjarry:
The previous test architecture was overly complex with two separate network namespaces simulating end hosts connected through routers. This complexity made the test harder to understand and maintain while not providing additional validation of BGP functionality.

Simplify the test to focus on the core requirement: verify that grout can learn BGP routes from an FRR peer and use them for forwarding. The test now uses an iBGP session between grout and a single FRR instance running in a network namespace. The FRR peer advertises a route to its loopback interface which grout must learn and use to successfully ping the remote address.

This reduces the number of interfaces from three to one and eliminates the need for two additional network namespaces and veth pairs.

## Summary by CodeRabbit

* **Tests**
  * Streamlined BGP routing test: replaced multi-host topology with a single contained BGP peer namespace, moved interface and route setup into that namespace, and removed prior preconfigured host/loopback routing.
  * Simplified readiness and validation: replaced multi-step route-learning checks with a repeated route-presence loop and switched end-to-end validation to an in-namespace ping, improving reliability and speed.

### DIFF
--- a/smoke/bgp_frr_test.sh
+++ b/smoke/bgp_frr_test.sh
@@ -2,158 +2,83 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2025 Andrea Panattoni
 
-# This test leverages the following architecture to test the BGP route exchange between an
-# FRR+Grout setup and a pure FRR BGP peer. At the end of the configuration phase, Host-A
-# and Host-B must be able to ping each other, as the respective default gateway routers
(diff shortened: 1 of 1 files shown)


maxime-leroy:
For vpn srv6 with bgp, it could be useful to have 2 netns to simulate 2 hosts.

christophefontaine (approved):

PR #392, opened 2025-11-09
smoke: use hw ports in grout namespace

rsafrono:
Since commits 4d5fa1c and 4853d4d grout is running in it's own namespace.  
When tests are using hardware ports we should ensure that the ports are accessible from that namespace.

## Summary by CodeRabbit

* **New Features**
  * Optional hardware-port mode to attach physical network ports (disabled by default).

* **Improvements**
  * Prevents creating more hardware ports than available, returning a clear error when exhausted.
  * Physical interfaces are moved into isolated network namespaces and temporarily renamed during setup.
  * Cleanup reliably restores original names and namespaces (includes short stabilization delay).
  * Port addition flow preserved after setup.

### DIFF
--- a/smoke/_init.sh
+++ b/smoke/_init.sh
@@ -125,13 +125,22 @@ port_add() {
 	local name="$1"
 	shift
         if [ "$use_hardware_ports" = true ]; then
+                if [ $tap_counter -ge ${#net_interfaces[@]} ]; then
+			fail "Can not create port. No more hardware ports available."
+		fi
+		nsenter -t 1 -n ip link set "${net_interfaces[$tap_counter]}" netns $(ip netns identify)
 		ip link set "${net_interfaces[$tap_counter]}" name "x-$name"
+                # When test fails prematurely due to insufficient number of ports
+                # we need to return them back to default namespace and wait a little
+                # before proceeding to ensure reliable execution
+		echo "ip link set x-$name netns 1 || :" >> $tmp/restore_interfaces
+		echo "sleep 1" >> $tmp/restore_interfaces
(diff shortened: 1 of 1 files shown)


smoke/_init.sh
```diff
@@ -125,13 +125,22 @@ port_add() {
 	local name="$1"
 	shift
         if [ "$use_hardware_ports" = true ]; then
+                if [ $tap_counter -ge ${#net_interfaces[@]} ]; then
+			fail "Can not create port. No more hardware ports available."
+		fi
```
rjarry: Is this related to this commit?

smoke/_init.sh
```diff
@@ -125,13 +125,22 @@ port_add() {
...
+		nsenter -t 1 -n ip link set "${net_interfaces[$tap_counter]}" netns grout
 		ip link set "${net_interfaces[$tap_counter]}" name "x-$name"
+                # When test fails prematurely due to insufficient number of ports
+                # we need to return them back to default namespace and wait a little
+                # before proceeding to ensure reliable execution
+		echo "ip link set x-$name netns 1 || :" >> $tmp/restore_interfaces
+		echo "sleep 1" >> $tmp/restore_interfaces
```
rjarry: I don't understand this. Could you explain with an example?

smoke/_init.sh
rsafrono: The tests can work without this, just error will be less clear, like 'no such index'. But I thought if the comment below relates to 'insufficient number of ports' scenario it worth to add a more clear message. I can omit it if it looks redundant here.

PR #396, opened 2025-11-18
frr: update version to 10.5.0

rjarry:
We don't need to patch FRR to build our out of tree plugin anymore. We just need a local patch when building FRR for our local tests.

Hopefully soon, we'll have frr-headers packages that we can install from Fedora and we'll be able to use them instead of building.

## Summary by CodeRabbit

* **Chores**
  * Upgraded FRR (Free Range Routing) dependency to 10.5.0 across builds, packaging, and container sources.
  * Removed an unnecessary patch-application step from the install/build flow.
  * Bumped Meson/external dependency checks to require the newer FRR version and aligned packaging metadata accordingly.

### DIFF
--- a/subprojects/frr.wrap
+++ b/subprojects/frr.wrap
@@ -1,6 +1,6 @@
 [wrap-git]
 url = https://github.com/FRRouting/frr
-revision = frr-10.4.1
+revision = frr-10.5.0
 depth = 1
 diff_files =
 	frr/meson-add-dependency-definition.patch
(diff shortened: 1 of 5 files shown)


maxime-leroy:
Just rebase , it should be successful. Can we use the new pkgconfig and kept compilation of FRR ?

subprojects/frr.wrap
```diff
@@ -1,6 +1,6 @@
 [wrap-git]
 url = https://github.com/FRRouting/frr
-revision = frr-10.4.1
+revision = frr-10.5.0
 depth = 1
 diff_files =
 	frr/meson-add-dependency-definition.patch
```
christophefontaine: This patch file must be updated to match FRR version: (line 64)
```
-project('frr', 'c', version: '10.4.1', license: 'GPL-2.0-or-later', meson_version: '>= 0.63.0')
+project('frr', 'c', version: '10.5.0', license: 'GPL-2.0-or-later', meson_version: '>= 0.63.0')
```

christophefontaine (approved):

PR #399, opened 2025-11-19
lacp fixes

rjarry:
- **bond: fix init with lacp mode**
- **port: silently ignore multicast mac filters**
- **lacp: only use tcp/udp ports for non-fragmented packets**

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified bond initialization and cleanup by removing unnecessary LACP MAC configuration steps.
  * Improved MAC handling to return errors for null inputs and gracefully ignore multicast addresses without changing filters.
  * Enhanced IPv4 fragmented-packet handling so transport ports are considered only for non-fragmented packets.

### DIFF
--- a/modules/infra/datapath/bond_output.c
+++ b/modules/infra/datapath/bond_output.c
@@ -104,22 +104,34 @@ hash_tx_member(const struct rte_mbuf *m, const struct iface_info_bond *bond) {
 			tuple.v4.dst_addr = l3.ip4->dst_addr;
 			switch (l3.ip4->next_proto_id) {
 			case IPPROTO_UDP:
-				l4.udp = rte_pktmbuf_mtod_offset(
-					m,
-					struct rte_udp_hdr *,
-					l3_offset + rte_ipv4_hdr_len(l3.ip4)
-				);
-				tuple.v4.sport = l4.udp->src_port;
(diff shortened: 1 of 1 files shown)


modules/infra/datapath/bond_output.c
```diff
@@ -104,22 +104,34 @@ hash_tx_member(const struct rte_mbuf *m, const struct iface_info_bond *bond) {
...
 			switch (l3.ip4->next_proto_id) {
 			case IPPROTO_UDP:
-				l4.udp = rte_pktmbuf_mtod_offset(
-					m,
-					struct rte_udp_hdr *,
-					l3_offset + rte_ipv4_hdr_len(l3.ip4)
-				);
-				tuple.v4.sport = l4.udp->src_port;
-				tuple.v4.dport = l4.udp->dst_port;
+				if (l3.ip4->fragment_offset == 0) {
```
aharivel: The **fragment_offset** field includes both flags (upper 3 bits) and offset (lower 13 bits). Maybe we should be more clearer on what we are testing by using something like: 

`l3.ip4->fragment_offset & RTE_BE16(RTE_IPV4_HDR_OFFSET_MASK | RTE_IPV4_HDR_MF_FLAG)) == 0`

PR #419, opened 2025-11-27
Ip frag reorder

christophefontaine:
Depends on #415.

## Summary by CodeRabbit

* **New Features**
  * Added a fragmentation limit and a new "too many fragments" drop reason to prevent excessive fragment generation.
  * Fragments are now emitted in batches for more efficient processing and queuing.

* **Bug Fixes**
  * Consolidated error/cleanup path to correctly handle partial fragment failures and preserve tracing.
  * Improved per-fragment tracing and cleanup to ensure all generated fragments are tracked and freed on error.

### DIFF
--- a/modules/ip/datapath/ip_fragment.c
+++ b/modules/ip/datapath/ip_fragment.c
@@ -17,6 +17,12 @@
 #include <stdint.h>
 #include <stdio.h>
 
+// According to RFC 791:
+// Every internet destination must be able to receive a datagram of 576 octets
+// either in one piece or in fragments to be reassembled.
+// With a jumbo frame (MTU set to 9000), we'd need 16 fragments.
+#define MAX_FRAGMENTS 16
+
 struct ip_fragment_trace_data {
 	uint16_t packet_id;
 	uint16_t frag_num;
@@ -28,6 +34,7 @@ enum {
 	IP_OUTPUT = 0,
 	NO_MBUF,
 	ALREADY_FRAGMENTED,
+	TOO_MANY_FRAGMENTS,
 	ERROR,
 	EDGE_COUNT,
 };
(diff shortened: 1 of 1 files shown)


modules/infra/datapath/trace.c
```diff
@@ -618,6 +618,38 @@ void *gr_mbuf_trace_add(struct rte_mbuf *m, struct rte_node *node, size_t data_l
...
+	struct gr_trace_item *src_trace, *dst_trace;
+	void *data;
+
+	// Reset trace head
+	STAILQ_INIT(dst_traces);
+
+	// Copy each trace item from source to destination
+	STAILQ_FOREACH (src_trace, src_traces, next) {
+		// Allocate new trace item for destination
+		while (rte_mempool_get(trace_pool, &data) < 0) {
```
aharivel: if `trace_pool` is exhausted and `traced_packets` ring is empty, this could loop forever. Maybe put a safety check ?

modules/ip/datapath/ip_fragment.c
```diff
@@ -34,7 +40,7 @@ enum {
 
 static uint16_t
 ip_fragment_process(struct rte_graph *graph, struct rte_node *node, void **objs, uint16_t nb_objs) {
-	struct rte_mbuf *mbuf, *frag_mbuf;
+	struct rte_mbuf *mbuf, *frag_mbuf, *frags[MAX_FRAGMENTS];
```
david-marchand: Since the node enqueue helper expects a void * array, and this frags[] array is not used for anything else, I would declare it as void *frags[].
This avoids the cast later.

modules/ip/datapath/ip_fragment.c
```diff
@@ -68,31 +76,23 @@ ip_fragment_process(struct rte_graph *graph, struct rte_node *node, void **objs,
...
 		if (unlikely(frag_size < 8)) {
 			edge = ERROR;
-			goto drop;
+			goto next;
 		}
 
 		num_frags = (data_len + frag_size - 1) / frag_size;
 		assert(num_frags > 1);
+		if (num_frags > MAX_FRAGMENTS) {
+			edge = ERROR; // XXX: add xstats
```
david-marchand: Why would you need a xstat? Can't you add a GR_DROP_REGISTER() node?

PR #423, opened 2025-11-29
deb,rpm: enable lto

rjarry:
Try and improve performance and binary size slightly by enabling link time optimization.

## Summary by CodeRabbit

* **Chores**
  * Enabled link-time optimization during the build process.
  * Added a compiler option to suppress specific string-operation overflow warnings.
  * Made the automatic test invocation a no-op during package build.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -1,7 +1,7 @@
 #!/usr/bin/make -f
 
 # Hardening
-export DEB_BUILD_MAINT_OPTIONS = hardening=+all reproducible=+all optimize=-lto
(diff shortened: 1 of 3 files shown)


christophefontaine:
I'm not really concerned about binary size, but what is the objective ?
Is there any side effects, if we have to debug the binary ?
Also, what is the performance improvement provided by LTO ?

christophefontaine:
I'm well aware of how it works. 
But I'm looking for facts, and how much % is improved, rather than hoping that it _may_ improve the performances.

rjarry:
I will do some benchmarks and let you know.

PR #425, opened 2025-12-01
port: disable hardware vlan filter for bifurcated drivers

rjarry:
When using grout with mlx5 VFs, asking to offload VLAN filter does not return an error but it causes all unicast packets to be steered to the kernel net device RXQs instead of the ones configured for grout.

Disabling VLAN filter offload fixes the issue. Unfortunately, there is no way of knowing whether a given device is using a "bifurcated" driver in a generic way. For now, rely on the driver name containing "mlx5". It is the only driver that we know of that has this limitation so far.

Add a new features field on the iface_info_port structure. Store a PORT_F_VLAN_FILTER flag when the port was successfully configured with RTE_ETH_RX_OFFLOAD_VLAN_FILTER offload.

If PORT_F_VLAN_FILTER isn't available, skip adding/removing any VLAN filter in the hardware.

## Release Notes

* **New Features**
  * Added VLAN filtering capability detection and management for network ports.

* **Bug Fixes**
  * Improved hardware VLAN offload handling for better device compatibility.
  * Added validation to prevent VLAN filter operations on unsupported devices.

### DIFF
--- a/modules/infra/control/port.c
+++ b/modules/infra/control/port.c
@@ -129,9 +129,19 @@ int port_configure(struct iface_info_port *p, uint16_t n_txq_min) {
 	if (info.dev_flags != NULL && *info.dev_flags & RTE_ETH_DEV_INTR_LSC) {
 		conf.intr_conf.lsc = 1;
 	}
+	if (strncmp(info.driver_name, "mlx5", 4) == 0) {
+		// Hardware VLAN filtering is incompatible with mlx5 bifurcated driver mode.
+		// Skip it for mlx5 devices and rely on software filtering instead.
+		conf.rxmode.offloads &= ~RTE_ETH_RX_OFFLOAD_VLAN;
+	}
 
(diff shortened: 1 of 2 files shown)


rjarry:
I think #428 will be cleaner. Let's close this.

modules/infra/control/port.c
```diff
@@ -129,10 +129,20 @@ int port_configure(struct iface_info_port *p, uint16_t n_txq_min) {
 	if (info.dev_flags != NULL && *info.dev_flags & RTE_ETH_DEV_INTR_LSC) {
 		conf.intr_conf.lsc = 1;
 	}
+	if (strncmp(info.driver_name, "mlx5", 4) == 0) {
+		// Hardware VLAN filtering is incompatible with mlx5 bifurcated driver mode.
+		// Skip it for mlx5 devices and rely on software filtering instead.
+		conf.rxmode.offloads &= ~RTE_ETH_RX_OFFLOAD_VLAN;
+	}
```
maxime-leroy: I would prefer a little bit, this approach.

Does it miss some tests on DPDK side ?

PR #427, opened 2025-12-01
infra/nexthop: validate VRF before creating nexthop, and drop nh ID if nexthop creation fails

maxime-leroy:
nexthop_new() didn’t verify whether the provided VRF or interface IDs were valid.

This could result in creating a nexthop with an invalid interface, causing Grout to crash when routing packets through it.

This patch adds checks to ensure the VRF and interface exist before allocating the nexthop.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of VRF and network interface lookups to detect invalid configurations earlier and return clearer errors.
  * Centralized error handling to ensure resources are released on failure, preventing leaks while preserving existing success behavior.

### DIFF
--- a/modules/infra/control/nexthop.c
+++ b/modules/infra/control/nexthop.c
@@ -390,6 +390,9 @@ int nexthop_update(struct nexthop *nh, const struct gr_nexthop_base *base, const
 			goto err;
 		}
 		nh->vrf_id = iface->vrf_id;
+	} else if (get_vrf_iface(nh->vrf_id) == NULL) {
+		ret = -errno;
+		goto err;
 	}
 
 	// Import type-specific info using callback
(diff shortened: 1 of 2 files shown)


modules/infra/control/nexthop.c
```diff
@@ -345,6 +345,12 @@ struct nexthop *nexthop_new(const struct gr_nexthop_base *base, const void *info
 	if (rte_lcore_has_role(rte_lcore_id(), ROLE_NON_EAL))
 		ABORT("nexthop created from datapath thread");
 
+	if (base->iface_id != GR_IFACE_ID_UNDEF && iface_from_id(base->iface_id) == NULL)
+		return errno_set_null(EINVAL);
```
maxime-leroy: We should probably check that the interface is the same vrf ?

Does it make any sense to be able to send a request for a new nexthop with vrf ID and  iface ID set ?

PR #428, opened 2025-12-02
port: remove hardware vlan filter

rjarry:
Depending on drivers, adding a VLAN filter can cause all other traffic (e.g. traffic not tagged with that precise VLAN) to be dropped before it even reaches the RX queues.

Grout has VLAN sub interfaces and needs to receive both tagged and untagged traffic.

Since there is no generic way of knowing how the hardware/driver will behave, remove VLAN filter offload completely. Grout will now receive *all* traffic and the VLAN filter will be done in software. This is acceptable since the switches should do their job and steer traffic in a sensible way.

Keep RTE_ETH_RX_OFFLOAD_VLAN_STRIP which isn't affected by VLAN filter.

## Summary by CodeRabbit

* **Refactor**
  * Removed VLAN management from bond and port interfaces and from the public API; VLANs must be managed via alternative methods.
  * Port default RX offload adjusted for VLAN strip behavior.
  * VLAN reconfiguration and teardown no longer perform automatic per-parent VLAN add/delete operations, changing how VLAN filters are maintained.

### DIFF
--- a/modules/infra/control/bond.c
+++ b/modules/infra/control/bond.c
@@ -141,34 +141,6 @@ static int bond_promisc_set(struct iface *iface, bool enabled) {
 	return bond_all_members_set_flag(iface, GR_IFACE_F_PROMISC, enabled, iface_set_promisc);
 }
 
-static int bond_vlan_add(struct iface *iface, uint16_t vlan_id) {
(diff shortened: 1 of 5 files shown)


rjarry:
This is an alternative fix to #425 which does not involve hard-coding a driver name.

PR #432, opened 2025-12-02
control-output: prevent use-after-free on object deletion

rjarry:
When an interface or nexthop is deleted while packets referencing them are still queued in the control output ring, callbacks would be invoked with dangling pointers, causing heap-use-after-free errors.

Example backtrace from smoke tests:

```
ERROR: AddressSanitizer: heap-use-after-free on address 0x7f1a06511100
READ of size 2 at 0x7f1a06511100 thread T0
  #0 in ndp_router_sollicit_input_cb ../modules/ip6/control/router_advert.c:109
  #1 in control_output_poll ../modules/infra/control/control_output.c:33
  #2 in event_base_loop
  #3 in main ../main/main.c:314
```

Add a drain parameter to all control_output_cb_t callbacks. When interfaces or nexthops are deleted, the control output ring is drained with a control_output_drain context containing the event type and deleted object pointer. Callbacks check this context and free packets that reference the deleted object without processing them.

Both GR_EVENT_IFACE_REMOVE and GR_EVENT_NEXTHOP_DELETE events are moved to occur after RCU synchronization. This ensures all datapath threads have seen that the objects are no longer accessible (removed from their respective lookup tables before the sync). At that point, no new packets referencing the deleted objects can be enqueued and only packets already in the control output ring may still hold references. The events trigger a drain before the object memory is freed, preventing callbacks from accessing freed memory.

### DIFF
--- a/modules/infra/control/control_output.c
+++ b/modules/infra/control/control_output.c
@@ -129,7 +129,7 @@ static void control_output_fini(struct event_base *) {
 
 static struct gr_module control_output_module = {
 	.name = "control_output",
-	.depends_on = "graph",
+	.depends_on = "*",
 	.init = control_output_init,
 	.fini = control_output_fini,
 };


modules/ip6/control/nexthop.c
```diff
@@ -240,6 +240,23 @@ void ndp_probe_input_cb(struct rte_mbuf *m) {
...
+	case GR_EVENT_IFACE_REMOVE:
+		if (mbuf_data(m)->iface == priv)
+			return true;
+		break;
+	case GR_EVENT_NEXTHOP_DELETE: {
+		struct nexthop *nh;
+		memcpy(&nh, control_output_mbuf_data(m)->cb_data, sizeof(struct nexthop *));
+		if (nh == priv)
+			return true;
+		break;
```
christophefontaine: maybe `return nh == priv;` ?

modules/ip6/control/nexthop.c
```diff
@@ -28,14 +28,29 @@
...
+		case GR_EVENT_IFACE_REMOVE:
+			if (mbuf_data(m)->iface == drain->obj)
+				goto free;
+			break;
+		case GR_EVENT_NEXTHOP_DELETE: {
+			if (nh == drain->obj)
+				goto free;
+			break;
+		}
+		}
```
maxime-leroy: why {} in GR_EVENT_NEXTHOP_DELETE and not GR_EVENT_IFACE_REMOVE ?

maxime-leroy (changes requested):

PR #435, opened 2025-12-03
frr: patch staticd to use non-uSID SRv6 behaviors

maxime-leroy:
Grout does not support uSID/CSID SIDs, but FRR staticd configures them by default, causing errors in SRv6 smoke tests ("not supported next-c-sid for srv6 local").

Apply a local FRR patch switching DT behaviors to their non-uSID variants (End, End_X, End_DT{4,6,46}) until upstream FRR adds native support.

(not merge until patch is accepted on FRR)


christophefontaine:
@maxime-leroy As the FRR PR is closed, (and as grout supports uSID) can we close this PR as well ?

PR #461, opened 2025-12-12
Graph SVG output fixes

rjarry:
- **graph: fix color of edges to port_output**
- **graph: add missing type info for new nodes**

### DIFF
--- a/modules/infra/control/gr_graph.h
+++ b/modules/infra/control/gr_graph.h
@@ -8,6 +8,7 @@
 #include <rte_common.h>
 #include <rte_graph.h>
 
+#include <assert.h>
 #include <sys/queue.h>
 
 #ifdef __GROUT_UNIT_TEST__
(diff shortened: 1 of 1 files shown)


modules/dhcp/datapath/dhcp_input.c
```diff
@@ -135,6 +135,7 @@ static struct rte_node_register node = {
 
 static struct gr_node_info info = {
 	.node = &node,
+	.type = GR_NODE_T_L4 | GR_NODE_T_CONTROL,
```
christophefontaine: Maybe we could an assert in GR_NODE_REGISTER to catch future missing type info ?

```c
#define GR_NODE_REGISTER(info)                                                                     \
        RTE_INIT(gr_node_register_##info) {                                                        \
                assert(info.type != 0);
                STAILQ_INSERT_TAIL(&node_infos, &info, next);                                      \
        }
```

christophefontaine (changes requested):
nit: modify the commit message, ack otherwise.

PR #464, opened 2025-12-15
port: handle RTE_ETH_EVENT_INTR_RESET events

rjarry:
Register a callback for RTE_ETH_EVENT_INTR_RESET which is raised by some DPDK drivers when the hardware requires a reset. This typically happens when the PF driver modifies VF settings from the host side, for example when changing a VF MAC address via ip link:

    ip link set $pf vf $vf_num mac $mac

When the reset event is received, the port is fully reset and reconfigured via rte_eth_dev_reset() followed by rte_eth_dev_configure() and rte_eth_dev_start(). The MAC address is re-read from hardware to pick up any changes made by the PF.

The callback is invoked from DPDK internal threads, so we use event_active() to defer the actual reset handling to the main control plane event loop.

Note that RTE_ETH_EVENT_INTR_RESET support varies by driver. The Intel iavf driver reports this event when the VF MAC is changed from the PF. The mlx5 driver does not report this event in the same scenario.

### DIFF
--- a/modules/infra/control/port.c
+++ b/modules/infra/control/port.c
@@ -28,6 +28,7 @@
 #include <rte_malloc.h>
 
 #include <arpa/inet.h>
+#include <pthread.h>
 #include <stdio.h>
 #include <string.h>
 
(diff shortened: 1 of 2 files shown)


modules/infra/control/port.c
```diff
@@ -668,6 +675,51 @@ static int lsc_port_cb(
...
+	pthread_mutex_unlock(&reset_ports_lock);
+
+	gr_vec_foreach (uint16_t pid, port_ids) {
+		struct iface *iface = (struct iface *)port_get_iface(pid);
+		if (iface != NULL) {
+			struct iface_info_port *port = iface_info_port(iface);
+			struct gr_iface_info_port api = {.mac = {{0}}};
+			LOG(INFO, "%s: port %u reset", iface->name, pid);
+			port->needs_reset = true;
+			iface_port_reconfig(iface, GR_PORT_SET_MAC, NULL, &api);
```
christophefontaine: IIUC, this will override any manually configured MAC on the port, won't it ?

On a side note, on a mac change event, do we trigger a gratuitous ARP and a NA message to tell all neighbors that the hardware address is modified ?

rjarry: > IIUC, this will override any manually configured MAC on the port, won't it ?

Yes indeed. Unfortunately, we don't have a way of knowing if the user had configured a specific mac address.

> On a side note, on a mac change event, do we trigger a gratuitous ARP and a NA message to tell all neighbors that the hardware address is modified ?

We don't, but we should. Not only when there is a VF reset by the way. I can address this in a separate patch if that is OK.

PR #516, opened 2026-02-24
l2: add bridge and FDB statistics

fabiendupont:
## Summary

Expose per-bridge forwarding, drop and learning counters aggregated
across all datapath cores. Each counter category is maintained in
per-core arrays to avoid cache contention in the datapath, then
summed on demand when the control plane queries them.

Depends on #504.

The bridge_stats structure tracks unicast, broadcast, multicast and
flood forwarding, various drop reasons (no FDB entry, hairpin,
interface down), and learning outcomes (success, update, failure,
limit exceeded). The fdb_stats structure tracks lookup hits, misses
and aged entries.

Statistics are cleared when a bridge is destroyed. A reset API
allows clearing counters without destroying the bridge.

CLI commands are exposed under `bridge stats show/reset`.

## Test plan

- [ ] Unit tests pass (`meson test`)
- [ ] Smoke tests pass (`smoke/run.sh`)
- [ ] `grcli bridge stats show` displays counters after traffic
- [ ] `grcli bridge stats reset` clears counters

## Summary by CodeRabbit

* **New Features**
  * VXLAN interface support (VNI, encap VRF, local/dst port, MAC) with ingress/egress datapath and tracing
  * Bridge interface support and VXLAN-aware flooding path
  * Flood VTEP management and flood list/events; VTEP shown in FDB listings
  * Per-bridge L2 FDB statistics (hit/miss/flood) with show/reset commands
  * UDP/L4 port aliasing for input ports
  * Enhanced CLI: VXLAN, flood (VTEP) and FDB stats commands; improved FDB display

* **Tests**
  * Added VXLAN and EVPN/VXLAN integration smoke tests


rjarry (changes requested):
I wonder if using rte_graph xstats would be a better direction here. What do you think?

https://doc.dpdk.org/api/structrte__node__xstats.html

rjarry (changes requested):

PR #558, opened 2026-03-16
infra: fallback for non-multicast capable drivers

christophefontaine:
af_packet and af_xdp drivers do not support enabling "all_multi". When we add a port, if the call to enable all_multi fails, try to enable "promisc" mode.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network interface configuration with enhanced error handling for multicast mode activation. If enabling multicast fails, the system now logs the error, automatically attempts an alternative (promiscuous) mode, and updates interface status accordingly. This fallback preserves connectivity and keeps interface state accurate, resulting in more reliable and stable network behavior across varied deployment scenarios.

### DIFF
--- a/modules/infra/control/port.c
+++ b/modules/infra/control/port.c
@@ -349,9 +349,15 @@ static int iface_port_reconfig(
 			return ret;
 
 		// always enable allmulti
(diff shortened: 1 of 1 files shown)


christophefontaine:
Closing as #702 is active.

PR #559, opened 2026-03-17
[RFC] infra: add iface mirror node

christophefontaine:
For all interfaces, add the basic infra to mirror packets.
As we may not want to copy _all_ packets, add a bpf filter:

grout# interface set port p0 mirror on filter "udp"

WIP: Packets are traced on grout stdoutput and dropped.

If the bpf filter can't be compiled (lack of JIT on aarch64), the BPF code is interpreted.
This is _partially_ fixed with https://dpdk.org/patch/162488 (suboptimal implementation compared to x86_64, but should be faster than interpreting BPF code !).

A new optional dynfield is allocated for pcapng_packetid.
As the packet may be captured on both ingress and egress interfaces,
this unique identifier allows a correlation after any modification
(for example, NAT) by the router.

TODO:
The next patch will write these packets to a pcapng file, from the control plane.


modules/infra/control/mirror.c
```diff
@@ -0,0 +1,61 @@
...
+	struct rte_bpf_prm *prm;
+	struct rte_bpf *bpf;
+	pcap_t *pcap;
+
+	*out = NULL;
+
+	if (expr == NULL || expr[0] == '\0')
+		return 0;
+
+	pcap = pcap_open_dead(DLT_EN10MB, 262144);
```
rjarry: I am not familiar with libpcap, could you add a comment explaining what are these arguments? Especially the `262144` magic number.

rjarry (changes requested):
Hey Chris,

I know this is still in progress, but I have a couple of remarks.

PR #576, opened 2026-03-26
ip,ip6: add rate limiting for ICMP errors, ARP and ICMP input

### LINKED ISSUE #14: implement icmp rate limiting

rjarry:
Under a flood of bad packets (TTL=1, no-route, etc.), every packet triggers expensive ICMP error construction: address lookups, header prepend with potential memmove, and checksum computation. Similarly, ARP and ICMP/ICMPv6 input processing can saturate the control plane thread when flooded. None of this is rate limited today.

Add a per-node token bucket to all ICMP error nodes, ARP input, and ICMP/ICMPv6 input nodes. The bucket refills once per second and the check runs before any per-packet work. When the bucket is empty, the whole vector is moved to a dedicated drop node via rte_node_next_stream_move(), avoiding any unnecessary processing.

Three separate rates are configurable via the graph config API: icmp-error-rate, arp-rate and icmp-rate. All default to 1000 packets/sec per node per worker. Setting a rate to 0 disables the limit. Rate changes take effect immediately without graph reload.

Closes: #14 

## Rate limiting for ICMP errors, ARP, and ICMP/ICMPv6 input

Behavior
- Per-node token-bucket limiter applied at node entry for ARP input, ICMP input (IPv4), ICMPv6 input, and IP/IPv6 ICMP error nodes.
- Limiter check runs before any per-packet work; when rate_limited(...) returns true the node moves the entire incoming vector with rte_node_next_stream_move(graph, node, RATE_LIMITED) and returns nb_objs, skipping address lookups, header prepends/memmoves, checksum computation, and enqueueing.
- Token refill is TSC-based and performed once per second (rte_rdtsc()/rte_get_tsc_hz()).
- A max_rate of 0 disables limiting.
- When limited, streams are routed to the shared drop node "error_rate_limited" (GR_DROP_REGISTER(error_rate_limited) in ip_error.c).

Implementation
- Added struct rate_limit_ctx { uint16_t tokens; uint64_t last_refill; } and static inline bool rate_limited(struct rate_limit_ctx *ctx, const uint16_t max_rate, const uint16_t nb_pkts) in modules/infra/datapath/datapath.h.
  - rate_limited(): asserts ctx, returns false if max_rate==0, reads rte_rdtsc(); if now - last_refill >= rte_get_tsc_hz() then tokens = max_rate and last_refill = now; if tokens == 0 returns true; otherwise decrements tokens by min(tokens, nb_pkts) and returns false.
- Each affected node stores a node-local context containing struct rate_limit_ctx (GR_NODE_CTX_TYPE(...)) and registers an .init hook that initializes ctx->limit.tokens = graph_conf.*_rate and ctx->limit.last_refill = rte_rdtsc().
- Nodes updated to perform the early check and add RATE_LIMITED next-node mapped to "error_rate_limited":
  - modules/ip/datapath/arp_input.c
  - modules/ip/datapath/icmp_input.c
  - modules/ip6/datapath/icmp6_input.c
  - modules/ip/datapath/ip_error.c (ttl_exceeded, no_route, frag_needed initializers)
  - modules/ip6/datapath/ip6_error.c (ttl_exceeded, no_route initializers)
- "error_rate_limited" drop node is registered via GR_DROP_REGISTER(error_rate_limited) and referenced by RATE_LIMITED mappings.

Control-plane and configuration
- modules/infra/api/gr_infra.h:
  - struct gr_graph_conf extended with uint16_t icmp_error_rate, arp_rate, icmp_rate (comments indicate defaults 1000, 0 disables).
  - Added gr_graph_conf_set_attr_t bitmask flags including GR_GRAPH_SET_ICMP_ERROR, GR_GRAPH_SET_ARP, GR_GRAPH_SET_ICMP; existing RX_BURST/VECTOR flags preserved.
  - New request type struct gr_graph_conf_set_req embeds struct gr_graph_conf plus set_attrs bitmask for selective updates.
- modules/infra/control/graph.c / graph.h:
  - graph_conf promoted to an external global (extern struct gr_graph_conf graph_conf) so datapath reads rates directly.
  - graph_conf defaults set: icmp_error_rate = 1000, arp_rate = 1000, icmp_rate = 1000.
  - graph_conf_set() changed to accept gr_graph_conf_set_req and apply only fields indicated by set_attrs; rx_burst/vector validated per-field; rate fields assigned directly when set_attrs bits present.
  - The early "no-change" reload decision excludes rate fields so updating rates does not trigger the graph reload path; rate changes take effect immediately without reloading the graph.

CLI
- modules/infra/cli/graph.c:
  - graph config set uses gr_graph_conf_set_req and accepts optional ICMP_ERROR_RATE, ARP_RATE, ICMP_RATE parameters; corresponding set_attrs bits set only when parameters parse successfully.
  - graph config show prints icmp_error_rate, arp_rate, icmp_rate.

Precise semantics from code
- Token budget resets to max_rate once per second (using rte_get_tsc_hz()).
- The limiter decrements by the vector size (nb_objs) in one operation using min(tokens, nb_objs).
- Rate limits are configured per node and per worker via graph_conf.*_rate fields; setting a rate to 0 disables that limiter.

### DIFF
--- a/api/gr_api.h
+++ b/api/gr_api.h
@@ -11,7 +11,7 @@
 #include <stdlib.h>
 
 // Must be bumped when making non-backward compatible changes in API headers
-#define GR_API_VERSION 3
+#define GR_API_VERSION 4
 
 // API request header.
 struct gr_api_request {
--- a/api/gr_api.h
+++ b/api/gr_api.h
@@ -11,7 +11,7 @@
 #include <stdlib.h>
 
 // Must be bumped when making non-backward compatible changes in API headers
-#define GR_API_VERSION 2
+#define GR_API_VERSION 3
 
 // API request header.
 struct gr_api_request {
--- a/api/gr_api.h
+++ b/api/gr_api.h
@@ -11,7 +11,7 @@
 #include <stdlib.h>
 
 // Must be bumped when making non-backward compatible changes in API headers
-#define GR_API_VERSION 7
+#define GR_API_VERSION 8
 
 // API request header.
 struct gr_api_request {
--- a/api/gr_api.h
+++ b/api/gr_api.h
@@ -11,7 +11,7 @@
 #include <stdlib.h>
 
 // Must be bumped when making non-backward compatible changes in API headers
-#define GR_API_VERSION 6
+#define GR_API_VERSION 7
 
 // API request header.
 struct gr_api_request {
--- a/modules/infra/cli/graph.c
+++ b/modules/infra/cli/graph.c
@@ -13,11 +13,16 @@
 #include <stdio.h>
 
 static cmd_status_t graph_conf_set(struct gr_api_client *c, const struct ec_pnode *p) {
-	struct gr_graph_conf req = {0};
+	struct gr_graph_conf_set_req req = {.set_attrs = 0};
 
-	if (arg_u16(p, "VECTOR", &req.vector_max) < 0 && errno != ENOENT)
+	if (arg_u16(p, "VECTOR", &req.vector_max) == 0)
+		req.set_attrs |= GR_GRAPH_SET_VECTOR;
+	else if (errno != ENOENT)
 		return CMD_ERROR;
-	if (arg_u16(p, "BURST", &req.rx_burst_max) < 0 && errno != ENOENT)
+
+	if (arg_u16(p, "BURST", &req.rx_burst_max) == 0)
+		req.set_attrs |= GR_GRAPH_SET_RX_BURST;
+	else if (errno != ENOENT)
 		return CMD_ERROR;
 
 	if (gr_api_client_send_recv(c, GR_GRAPH_CONF_SET, sizeof(req), &req, NULL) < 0)
--- a/modules/infra/cli/graph.c
+++ b/modules/infra/cli/graph.c
@@ -25,6 +25,21 @@ static cmd_status_t graph_conf_set(struct gr_api_client *c, const struct ec_pnod
 	else if (errno != ENOENT)
 		return CMD_ERROR;
 
+	if (arg_u16(p, "ICMP_ERROR_RATE", &req.icmp_error_rate) == 0)
+		req.set_attrs |= GR_GRAPH_SET_ICMP_ERROR;
+	else if (errno != ENOENT)
+		return CMD_ERROR;
+
+	if (arg_u16(p, "ARP_RATE", &req.arp_rate) == 0)
+		req.set_attrs |= GR_GRAPH_SET_ARP;
+	else if (errno != ENOENT)
+		return CMD_ERROR;
+
+	if (arg_u16(p, "ICMP_RATE", &req.icmp_rate) == 0)
+		req.set_attrs |= GR_GRAPH_SET_ICMP;
+	else if (errno != ENOENT)
+		return CMD_ERROR;
+
 	if (gr_api_client_send_recv(c, GR_GRAPH_CONF_SET, sizeof(req), &req, NULL) < 0)
 		return CMD_ERROR;
 
@@ -43,6 +58,9 @@ static cmd_status_t graph_conf_show(struct gr_api_client *c, const struct ec_pno
 	struct gr_object *o = gr_object_new(NULL);
 	gr_object_field(o, "vector_max", GR_DISP_INT, "%u", sizes->vector_max);
 	gr_object_field(o, "rx_burst_max", GR_DISP_INT, "%u", sizes->rx_burst_max);
+	gr_object_field(o, "icmp_error_rate", GR_DISP_INT, "%u", sizes->icmp_error_rate);
+	gr_object_field(o, "arp_rate", GR_DISP_INT, "%u", sizes->arp_rate);
+	gr_object_field(o, "icmp_rate", GR_DISP_INT, "%u", sizes->icmp_rate);
 	gr_object_free(o);
 
 	free(resp_ptr);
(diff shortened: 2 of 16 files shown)


api/gr_api.h
```diff
@@ -11,7 +11,7 @@
 #include <stdlib.h>
 
 // Must be bumped when making non-backward compatible changes in API headers
-#define GR_API_VERSION 2
+#define GR_API_VERSION 3
```
christophefontaine: As `gr_graph_conf_set_attr_t` and `gr_graph_conf` are modified, do we need to bump this GR_API_VERSION manually ?

modules/infra/cli/graph.c
```diff
@@ -81,14 +99,27 @@ static int ctx_init(struct ec_node *root) {
...
+			"ICMP errors/sec per node per worker (0 = no limit).",
+			ec_node_uint("ICMP_ERROR_RATE", 0, UINT16_MAX, 10)
+		),
+		with_help(
+			"ARP packets/sec per worker (0 = no limit).",
+			ec_node_uint("ARP_RATE", 0, UINT16_MAX, 10)
+		),
+		with_help(
+			"ICMP/ICMPv6 input packets/sec per worker (0 = no limit).",
+			ec_node_uint("ICMP_RATE", 0, UINT16_MAX, 10)
```
christophefontaine, line 122: What about setting the lower bound to RTE_GRAPH_BURST_SIZE, to have an explicit value for the user ?

With the current code: `	ctx->tokens -= RTE_MIN(ctx->tokens, nb_pkts);`, if the user sets the number of tokens to `1`, we may still send up to `RTE_GRAPH_BURST_SIZE` errors, or ICMP packets, etc.

api/gr_api.h
rjarry: Yes, we do. Otherwise the API check complains about non-compatible changes made without explicitly bumping the API version.

rjarry:
```
Checking for API changes between v0.15.0-35-g84e372a9 and v0.15.0-36-gc8cc65e7

1 function with some indirect sub-type change:

  [C] 'function gr_graph_conf* GR_GRAPH_CONF_GET_(gr_empty*)' at gr_infra.h:420:1 has some indirect sub-type changes:
    return type changed:
      in pointed to type 'struct gr_graph_conf' at gr_infra.h:412:1:
        type size changed from 4 to 10 (in bytes)
        3 data member insertions:
          'uint16_t icmp_error_rate', at offset 4 (in bytes) at gr_infra.h:415:1
          'uint16_t arp_rate', at offset 6 (in bytes) at gr_infra.h:416:1
          'uint16_t icmp_rate', at offset 8 (in bytes) at gr_infra.h:417:1

1 changed type unreachable from any public interface:

  [C] 'struct gr_graph_conf_set_req' changed:
    type size changed from 6 to 12 (in bytes)
    2 data member changes:
      anonymous data member at offset 0 (in bytes) changed from:
        struct gr_graph_conf{uint16_t gr_graph_conf::rx_burst_max; uint16_t gr_graph_conf::vector_max;}
      to:
        struct gr_graph_conf{uint16_t gr_graph_conf::rx_burst_max; uint16_t gr_graph_conf::vector_max; uint16_t gr_graph_conf::icmp_error_rate; uint16_t gr_graph_conf::arp_rate; uint16_t gr_graph_conf::icmp_rate;}
      and size changed from 4 to 10 (in bytes) (by +6 bytes)
      'gr_graph_conf_set_attr_t set_attrs' offset changed from 4 to 10 (in bytes) (by +6 bytes)

breaking API changes
GR_API_VERSION changed from 2 to 3
```

modules/infra/cli/graph.c
rjarry: But then how would you disable it?

Indeed, the limit is soft and we allow the full burst to pass even if there is less than nb_pkts left in the bucket. But since it is transient, I thought it was fine.

modules/infra/datapath/datapath.h
```diff
@@ -3,4 +3,32 @@
...
+	uint64_t now = rte_rdtsc();
+	assert(ctx != NULL);
+
+	if (max_rate == 0)
+		return false;
+
+	if (now - ctx->last_refill >= rte_get_tsc_hz()) {
+		ctx->tokens = max_rate;
+		ctx->last_refill = now;
+	} else if (ctx->tokens == 0) {
```
ghost: Just a minor detail, shouldn’t the refill be proportional?

rjarry: Proportional to what? I am not familiar with token bucket algorithms I have implemented the most basic solution.

ghost: It is not far from what you have added, simply:
- define the refill rate of your bucket --> reflects to number of msgs per second
- define the capacity of your bucket  --> in idle state this will allow some instantaneous burst behavior

your refill should be similar to:
```
add_tokens = (now - ctx->last_refill) * max_rate / rte_get_tsc_hz(); // proportional to elapsed ticks 
(some granularity guard can be added here to skip short elapsed time)
if (!add_tokens) return;
ctx->tokens += add_tokens;
ctx->tokens = RTE_MIN(ctx->tokens, ctx->bucket_capacity);
ctx->last_refill = now;
```
I just found the kernel implementation a great example to inspire of:
- https://docs.kernel.org/networking/ip-sysctl.html
- https://github.com/torvalds/linux/blob/dca922e019dd758b4c1b4bec8f1d509efddeaab4/net/ipv4/icmp.c#L229

```
net->ipv4.icmp_global_credit --> your tokens
net->ipv4.sysctl_icmp_msgs_per_sec --> refill rate
net->ipv4.icmp_global_stamp --> last refill
net->ipv4.sysctl_icmp_msgs_burst  --> bucket capacity
```

rjarry: got it, thanks!

PR #592, opened 2026-04-13
frr: add FRR 10.6.0 compatibility

zeeke:
FRR 10.6.0 introduced several API changes that break compilation:

- rib_add_multipath() gained a bool replace parameter
- dplane_ctx_mac_set_vtep_ip() and dplane_ctx_mac_get_vtep_ip() now use struct ipaddr instead of struct in_addr
- zebra_l2info_vxlan.vtep_ip changed from struct in_addr to struct ipaddr
- struct event_loop is now opaque, requiring the new accessor frr_event_loop_get_pthread_owner() instead of direct field access

Detect the FRR version at build time in meson and define CURRENT_FRR_VERSION macro. Guard
all changed call sites with #ifdef to support both 10.5 and 10.6.

## Compatibility with FRR 10.6.0

Build-time FRR version detection and conditional compilation enable building against both FRR 10.5.x and FRR 10.6.x API variants by exposing a packed numeric FRR version to the compiler and guarding affected call sites with #if CURRENT_FRR_VERSION >= MAKE_FRRVERSION(10,6,0).

### Build system changes
- frr/meson.build: Read the resolved FRR dependency version, split into major/minor/sub, compute a packed integer using MAKE_FRRVERSION layout (major*2^24 + minor*256 + sub), and pass it to the compiler as -DCURRENT_FRR_VERSION=<value> via c_args for the dplane_grout shared module.
- subprojects/packagefiles/frr/meson.build: Export the FRR dependency version by adding version: meson.project_version() to the frr_dep declare_dependency().

### API adaptations (version-guarded)
- frr/if_grout.c:
  - Include <lib/libfrr.h>.
  - VXLAN L2 info: when CURRENT_FRR_VERSION >= MAKE_FRRVERSION(10,6,0) populate vi.vtep_ip as struct ipaddr (set ipa_type = IPADDR_V4 and write ipaddr_v4.s_addr); otherwise assign legacy vi.vtep_ip.s_addr. vi is passed unchanged to dplane_ctx_set_ifp_vxlan_info().

- frr/rt_grout.c:
  - Include <lib/libfrr.h>.
  - Route installation: call rib_add_multipath(...) with the additional bool replace parameter (two trailing false arguments) when building against FRR >= 10.6.0; retain the prior call signature for older FRR.
  - MAC-FDB handling: when FRR >= 10.6.0, pass and read VTEP addresses using struct ipaddr (ipa_type = IPADDR_V4 and ipaddr_v4.s_addr) to/from dplane_ctx_mac_set_vtep_ip()/dplane_ctx_mac_get_vtep_ip(); for older FRR keep using struct in_addr and ->s_addr.

- frr/zebra_dplane_grout.c:
  - Include <lib/libfrr.h>.
  - Event-loop ownership check: use frr_event_loop_get_pthread_owner(dg_master) == pthread_self() when CURRENT_FRR_VERSION >= MAKE_FRRVERSION(10,6,0); otherwise continue using direct dg_master->owner comparison.

### Notes
- All API changes are covered with #if CURRENT_FRR_VERSION conditionals so the codebase remains buildable against both older and newer FRR headers without changing public/exported interfaces.

### DIFF
--- a/frr/if_grout.c
+++ b/frr/if_grout.c
@@ -10,6 +10,7 @@
 #include <gr_l2.h>
 #include <gr_srv6.h>
 
+#include <lib/libfrr.h>
 #include <linux/if.h>
 #include <net/if.h>
 #include <zebra/interface.h>
@@ -173,7 +174,12 @@ void grout_link_change(struct gr_iface *gr_if, bool new, bool startup) {
 			vi.vni_info.iftype = ZEBRA_VXLAN_IF_VNI;
 			vi.vni_info.vni.vni = gr_vxlan->vni;
 			vi.ifindex_link = ifindex_grout_to_frr(gr_vxlan->encap_vrf_id);
+#if CURRENT_FRR_VERSION >= MAKE_FRRVERSION(10, 6, 0)
+			vi.vtep_ip.ipa_type = IPADDR_V4;
+			vi.vtep_ip.ipaddr_v4.s_addr = gr_vxlan->local;
+#else
 			vi.vtep_ip.s_addr = gr_vxlan->local;
+#endif
 			dplane_ctx_set_ifp_vxlan_info(ctx, &vi);
 		}
 	} else {
--- a/frr/meson.build
+++ b/frr/meson.build
@@ -18,6 +18,15 @@ endif
 
 install_build_flag = frr_dep.get_variable('install_on_build', default_value: 'false') == 'true'
 frr_moduledir = frr_dep.get_variable('moduledir')
+
+frr_version = frr_dep.version()
+frr_ver = frr_version.split('.')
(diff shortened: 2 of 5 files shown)


maxime-leroy:
I didn't have time to do a full review yet. 

But first question, do we plan to support several FRR version at the same time? In this case we should run smoke test on different version 

Most fixes are for vtep, nothing else is break with FRR 10.6 ?

christophefontaine:
@maxime-leroy Yes, it is planned to run the frr some tests against 10.5 and 10.6.
We're also discussing with @rjarry how to build & ship multiple versions of the .so.

How many versions of FRR ( or when we should remove the support for specific version) we want to support is still TBD.

christophefontaine (approved):
smoke tests ran on my laptop, works well !
I'm working on getting multiple FRR versions tested in GitHub.

PR #593, opened 2026-04-15
frr: tests multiple FRR versions

christophefontaine:
Add a new meson option to select the FRR version to build and test. Current supported values are '10.5.3', '10.6.0' and 'master'.

As meson wrap files don't support variables, add a wrap file for each version, use the proper one from frr/meson.build.

Run smoke tests with FRR 10.6.0 as well as the master branch.

## Summary

Adds configurable selection and CI coverage for multiple FRRouting (FRR) versions via Meson.

### Build configuration
- Introduced Meson combo option `frr_version` with choices `default`, `10.6.0`, and `master` (value `default` corresponds to the existing 10.5.3 behavior).

### Subproject / wrap handling
- Added separate wrap files:
  - `subprojects/frr-10.6.0.wrap` — fetches FRR 10.6.0 release archive.
  - `subprojects/frr-master.wrap` — fetches FRR master via shallow git checkout.
- Updated `frr/meson.build` to select the subproject based on `frr_version`:
  - `subproject('frr')` when `frr_version == 'default'`
  - `subproject('frr-' + frr_version)` otherwise
- Rationale (reflected in code): separate wrap files required because Meson wrap files do not support variable substitution.

### Package metadata
- `subprojects/packagefiles/frr/meson.build` now computes the project `version` by extracting the version string from `configure.ac` (using sed) instead of a hardcoded `'10.5.3'`.

### CI
- Extended `.github/workflows/check.yml` matrix with two new jobs that pass `-Dfrr=enabled` and `-Dfrr_version=10.6.0` or `-Dfrr_version=master` to exercise FRR 10.6.0 and master in CI (ubuntu-24.04, gcc-14, buildtype=debugoptimized, sanitize=none).

### Tests
- Smoke tests configured to run for FRR 10.6.0 in addition to the existing default branch.

### DIFF
--- a/frr/meson.build
+++ b/frr/meson.build
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 # Copyright (c) 2025 Maxime Leroy, Free Mobile
 frr_opt = get_option('frr')
+frr_version = get_option('frr_version')
 
 if frr_opt.disabled()
   subdir_done()
@@ -9,7 +10,11 @@ endif
 frr_dep = dependency('frr', version: '>= 10.5', required: false)
 if not frr_dep.found()
   if get_option('frr').enabled()
-    sp = subproject('frr')
+    if frr_version == 'default'
+      sp = subproject('frr')
+    else
+      sp = subproject('frr-' + frr_version)
+    endif
     frr_dep = sp.get_variable('frr_dep')
   else
     subdir_done()
--- a/subprojects/frr-10.6.0.wrap
+++ b/subprojects/frr-10.6.0.wrap
@@ -1,0 +1,9 @@
+[wrap-file]
+source_url = https://github.com/FRRouting/frr/archive/refs/tags/frr-10.6.0.tar.gz
+source_filename = frr-10.6.0.tar.gz
+source_hash = f575c821f0b6a2c9034eda0b464e6053eadce892d29ccaf54d60e601cc9f59ef
+directory = frr-frr-10.6.0
+patch_directory = frr
+
+[provide]
+dependency_names = frr-10.6.0
--- a/subprojects/frr-master.wrap
+++ b/subprojects/frr-master.wrap
@@ -1,0 +1,9 @@
+[wrap-git]
+url = https://github.com/FRRouting/frr
+revision = master
+depth = 1
+directory = frr-frr-master
+patch_directory = frr
+
+[provide]
+dependency_names = frr-master
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -37,6 +37,18 @@ jobs:
             os: ubuntu-24.04
             extra_opts: -Dfrr=enabled
             rebase: false
+          - compiler: gcc-14
+            sanitize: none
+            buildtype: debugoptimized
+            os: ubuntu-24.04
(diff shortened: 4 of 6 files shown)


christophefontaine:
Found the difference and why it fails here: I use meson 1.7.2, while meson 1.3.2 is used in the CI ...
Working on it.

frr/meson.build
```diff
@@ -1,15 +1,20 @@
...
+  version: '>= 10.5.0',
+  default_options: ['frr_version=' + frr_version],
+  required: false)
+
 if not frr_dep.found()
   if get_option('frr').enabled()
-    sp = subproject('frr')
+    sp = subproject('frr-' + frr_version)
     frr_dep = sp.get_variable('frr_dep')
   else
```
rjarry, line 19: This will break detection of FRR headers installed into /usr/include. You should change this to:

```meson
frr_dep = dependency('frr', version: '>= 10.5', required: false)
if not frr_dep.found()
  if get_option('frr').enabled()
    frr_dep = dependency('frr', fallback: ['frr-' + get_option('frr_version'), 'frr_dep'])
  else
    subdir_done()
  endif
endif
```

subprojects/frr-10.5.3.wrap
```diff
@@ -6,4 +6,4 @@ directory = frr-frr-10.5.3
 patch_directory = frr
 
 [provide]
-dependency_names = frr
+dependency_names = frr-10.5.3
```
rjarry: no need to change this with the method I posted above (I think).

subprojects/frr-10.6.0.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-file]
+source_url = https://github.com/FRRouting/frr/archive/refs/tags/frr-10.6.0.tar.gz
+source_filename = frr-10.6.0.tar.gz
+source_hash = f575c821f0b6a2c9034eda0b464e6053eadce892d29ccaf54d60e601cc9f59ef
+directory = frr-frr-10.6.0
+patch_directory = frr
+
+[provide]
+dependency_names = frr-10.6.0
```
rjarry, line 9: Same here, it should be

```ini
dependency_names = frr
```

subprojects/frr-master.wrap
```diff
@@ -0,0 +1,9 @@
+[wrap-git]
+url = https://github.com/FRRouting/frr
+revision = master
+depth = 1
+directory = frr-frr-master
+patch_directory = frr
+
+[provide]
+dependency_names = frr-master
```
rjarry, line 9: ditto

frr/meson.build
christophefontaine: Good point, I'll fix that.

subprojects/frr-10.5.3.wrap
christophefontaine: Unfortunately, meson fails if multiple wrap files provide the same dependency, so we need to add the version in these wrap files.

subprojects/frr-10.6.0.wrap
christophefontaine: same

subprojects/frr-master.wrap
christophefontaine: idem

christophefontaine:
@coderabbitai review

rjarry (approved):

PR #594, opened 2026-04-16
Api check stuff

rjarry:
## API inline function compatibility checking

Extends the API compatibility tooling to detect changes in generated inline wrapper functions by placing them into a dedicated ELF section and comparing their disassembly between revisions.

### Changes

api/gr_api.h
- Adds a conditional macro definition:
  - If GR_API_INLINE is not defined, define it as `#define GR_API_INLINE static inline`.

devtools/check_api.sh
- Compiler command: appends `-Wno-missing-declarations` and `-Wno-missing-prototypes`.
- Injects a GR_API_INLINE definition used during the check: `#define GR_API_INLINE __attribute__((section(".api_inline")))` so generated inline wrapper functions are emitted into the `.api_inline` ELF section.
- After building previous (`a.bin`) and current (`b.bin`) test binaries, extracts symbol names from the `.api_inline` section (using `objdump -t -j .api_inline`) and writes them to `inline_funcs`.
- Introduces `inline_change` variable (initialized false) and logic to set it to true when inline functions are added, removed, or when per-function disassembled instruction streams differ between the two builds.
- Integrates inline changes into the ABI check:
  - The `abidiff` invocation is treated as failed when `inline_change` is true (`if ! abidiff ... || [ "$inline_change" = true ]`).
  - The script treats inline changes as breaking-change triggers when determining whether to emit the `[DC]` diagnostic.
- Moves GR_API_VERSION extraction and the unchanged/changed reporting outside the `abidiff` failure conditional so the version comparison is always printed.

devtools/gen_api_header.sh
- Fixes the date command in the generated header comment from `date +Y` to `date +%Y` to emit the four-digit year.

### DIFF
--- a/devtools/gen_api_header.sh
+++ b/devtools/gen_api_header.sh
@@ -3,7 +3,7 @@
 # Copyright (c) 2026 Robin Jarry
 
 echo "// SPDX-License-Identifier: BSD-3-Clause"
-echo "// Copyright (c) $(date +Y) Red Hat"
+echo "// Copyright (c) $(date +%Y) Red Hat"
 echo
 echo "#pragma once"
 echo
--- a/api/gr_api.h
+++ b/api/gr_api.h
@@ -156,6 +156,9 @@ const char *gr_api_message_name(uint32_t type);
 		code##_OBJ_SIZE = sizeof(obj)                                                      \
 	}
 #endif
+#ifndef GR_API_INLINE
+#define GR_API_INLINE static inline
+#endif
 
 struct gr_empty { };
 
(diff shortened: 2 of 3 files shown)


devtools/gen_api_header.sh
```diff
@@ -3,7 +3,7 @@
 # Copyright (c) 2026 Robin Jarry
 
 echo "// SPDX-License-Identifier: BSD-3-Clause"
-echo "// Copyright (c) $(date +Y) Red Hat"
+echo "// Copyright (c) $(date +%Y) Red Hat"
```
maxime-leroy, line 6: ack

christophefontaine (approved):

PR #595, opened 2026-04-16
cli events improvements

rjarry:
## Event-driven smoke test orchestration

- Remove script runner and move orchestration into Makefile: `smoke/run.sh` deleted; GNUmakefile adds SMOKE_MATCH/SMOKE_SKIP glob-based filtering of smoke scripts, exports PAUSE_ON_FAILURE, INTERACTIVE, GDB (all default false) into a constructed `smoke_env` that is printed by a new phony `smoke_env_print` target and prepended to each test invocation (`sudo $(smoke_env) $@ $(BUILDDIR)`). Temporary log cleanup now uses a `trap` on EXIT instead of explicit rm on success/failure.

- Replace polling-based checks with event-driven synchronization: `smoke/_init.sh` records `grcli events` to `$tmp/events.log`, and adds `mark_events()` (capturing current log offset) and `wait_event()` (tails from the mark and blocks until a grep match or timeout). Test helpers and tests (`smoke/_init_frr.sh`, `bgp*_test.sh`, `evpn_vxlan_frr_test.sh`, etc.) call `mark_events()` before making configuration changes and use `wait_event` to wait for specific event patterns instead of retry loops that polled `grcli -j … | jq` repeatedly.

## CLI event subsystem: name-based filtering and safer subscriptions

- Add `const char *name` to `struct cli_event_printer` so printers can be looked up by string.

- `events_show`:
  - Parse optional COUNT and TYPE arguments from `ec_pnode`.
  - If TYPE absent, subscribe once to GR_EVENT_ALL; if one or more TYPE values provided, resolve each TYPE to a registered `cli_event_printer` by `name` and issue GR_EVENT_SUBSCRIBE for that printer’s `ev_types`, returning EINVAL for unknown names.
  - Stop conditions: global atomic `stop` set by signal handlers (SIGINT/SIGTERM/SIGPIPE) or when printed events reach COUNT (COUNT defaults to unlimited).
  - Install signal handlers via `sigaction` before entering receive loop and restore them after.
  - Preserve errno across GR_EVENT_UNSUBSCRIBE.
  - Add completion callback `complete_event_names` to suggest valid TYPE values.

- Populate `.name` in many module printers: `iface`, `nexthop`, `addr4`, `addr6`, `route4`, `route6`, `fdb`, `flood` to enable name-based filtering.

## Event output formatting and cache hardening

- Replace numeric interface/VRF IDs with resolved names in event output across modules:
  - Address, route, nexthop, ip6 route/address, l2 fdb/flood, etc., now use `iface_name_from_id(NULL, id)` and print string values instead of `%u` numeric IDs.
- Harden `iface_name_from_id()` cache fill: only call `iface_from_id(c, ifid)` when client pointer `c` is non-NULL; when lookup fails, fall back to storing the raw ID string. In `iface_event_print()` copy `iface->name` into `name_cache[iface->id]` and clear cache on `GR_EVENT_IFACE_REMOVE`.

## Signal and error handling changes

- Use `sigaction` for SIGINT handlers and restore previous handlers with `sigaction` (applied in ICMP/ICMPv6 ping/traceroute implementations) and change shared stop flags to `static volatile sig_atomic_t` for signal-safety.

- Adjust send/receive socket helpers in `api/gr_api_client_impl.h`:
  - `send_all()` no longer retries on `EINTR`/`EAGAIN`; it returns the negative `send()` result immediately on any send error.
  - `recv_all()` calls `recv()` with flags `0` (removed `MSG_WAITALL`) and similarly stops retrying on `EINTR`/`EAGAIN`, returning the negative `recv()` result immediately on error.
  - Partial transfer semantics now rely on underlying `send`/`recv` behavior rather than internal retry loops.

## Control-flow and error-path simplifications

- In ICMP/ICMPv6 send routines, replace immediate `return CMD_ERROR` on send/recv failures with incrementing an `errors` counter and breaking out of loops to allow unified post-loop handling.

- Smoke test helper functions (e.g., `set_ip_address`, `set_ip_route`, SRv6 helpers) removed local retry loops and instead mark events prior to configuration and then wait via `wait_event` for the expected event(s); related per-function polling state (`max_tries`/`count`) removed where unused.


modules/infra/cli/iface.c
```diff
@@ -594,6 +594,8 @@ static void iface_event_print(uint32_t event, const void *obj) {
 	const struct gr_iface *iface = obj;
 	const char *action;
 
+	gr_strcpy(name_cache[iface->id], sizeof(name_cache[iface->id]), iface->name);
```
maxime-leroy: if (ifid >= ARRAY_DIM(name_cache))
		return;

Defensive programming like is done in iface_name_from_id ?

rjarry: I was assuming that the event contains a correct interface id since it is sent by grout. But you are right.

api/gr_api_client_impl.h
```diff
@@ -168,8 +168,6 @@ static ssize_t send_all(const struct gr_api_client *c, const void *buf, size_t l
 	while (remaining > 0) {
 		n = send(c->sock_fd, ptr, remaining, MSG_NOSIGNAL);
 		if (n < 0) {
-			if (errno == EINTR || errno == EAGAIN)
-				continue;
 			return n;
```
maxime-leroy, line 171: if (n < 0) { return n };

useless braces, can be removed here

maxime-leroy (approved):

PR #596, opened 2026-04-16
Fix control-plane socket handling for FRR daemons (BFD, OSPF, BGP-in-VRF)

maxime-leroy:
FRR daemons had several distinct failures when running against grout,                                                                                                                                                               
  all rooted in control-plane path issues:                                                                                                                                                                                            
                                                                                                                                                                                                                                      
  - **OSPF** convergence was flaky: NOARP unicast packets (DD/LSR/LSU/                                                                                                                                                                
    LSAck) sent to unresolved neighbors went out with a stale MAC and                                                                                                                                                                 
    were dropped by the peer.                                                                                                                                                                                                         
  - **BFD** sessions never received replies: bfdd uses separate send/                                                                                                                                                                 
    receive sockets bound to the port TAP, but replies were forwarded                                                                                                                                                                 
    out the VRF TUN loopback, which is a different kernel device.                                                                                                                                                                     
  - **BGP in a non-default VRF** could not receive traffic: grout created                                                                                                                                                             
    the Linux VRF master but enslaved only its TUN loopback, so sockets                                                                                                                                                               
    bound to the VRF master (or unbound sockets relying on                                                                                                                                                                            
    l3mdev_accept) were blind to traffic arriving on port TAPs.                                                                                                                                                                       
  - **IPv4 SO_BINDTODEVICE** on those TAPs then broke for a different                                                                                                                                                                 
    reason: once enslaved, the l3mdev slave rule forced lookups into the                                                                                                                                                              
    VRF table which only contained a default via the TUN loopback                                                                                                                                                                     
    (RTA_OIF mismatch, ENETUNREACH). IPv6 had a policy-routing rule for                                                                                                                                                               
    this case; IPv4 did not.                                                                                                                                                                                                          
                                                                                                                                                                                                                                      
  Fixes, one per commit:                                                                                                                                                                                                              
                                                                                                                                                                                                                                      
  - `ctlplane`: re-inject NOARP unicast through the L3 loopback so they
    follow the normal ARP/NDP resolution path.                                                                                                                                                                                        
  - `l4`: steer locally-handled replies to the ingress port TAP when the
    iface has a cp_fd, instead of always going through the VRF TUN.                                                                                                                                                                   
  - `infra`: enslave port/VLAN TAPs to the kernel VRF master in
    non-default VRFs.                                                                                                                                                                                                                 
  - `infra`: extend the per-oif policy-routing rule (shared table 999)
    to IPv4.                                                                                                                                                                                                                          
                  
  Coverage:                                                                                                                                                                                                                           
  - New `so_bindtodevice_test` exercises 8 bind scenarios (port TAP, VRF
    master, VLAN, non-default VRF, nobind, nobind+l3mdev) across UDP/TCP                                                                                                                                                              
    over IPv4/IPv6.                                                                                                                                                                                                                   
  - Previously-disabled `ospf` and `ospf6` tests are re-enabled, now                                                                                                                                                                  
    passing 20/20 on CI runners.                                                                                                                                                                                                      
  - Existing bgp/bfd FRR tests unaffected.  

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Overview
Fixes control-plane socket handling for FRR daemons (BFD, OSPF, BGP in non-default VRF) by correcting receive and send paths for packets tied to TAP port representors and installing per-interface policy routes for IPv4 and IPv6.

## Technical Changes

### modules/infra/control/ctlplane.c — NOARP receive re-injection and VRF master wiring
- Detect Ethernet frames whose destination MAC equals the interface MAC in iface_cp_poll(); strip the Ethernet header, mark the mbuf as L3 (IPv4/IPv6), set eth_input_mbuf_data.iface to the VRF L3 loopback iface and domain = ETH_DOMAIN_LOOPBACK, and reinject via post_to_stack(loopback_get_control_id(), mbuf). This forces such locally-destined packets to traverse normal ARP/NDP resolution before being transmitted.
- Add cp_set_vrf_master() and call it from cp_update() to set the kernel link master for control-plane interfaces (netlink_link_set_master) so port/VLAN TAPs are enslaved to a non-default VRF master when applicable.
- On interface events, call netlink_add_cp_route()/netlink_del_cp_route() to add/remove the per-oif policy rule + default route when the interface comes up or is removed; call netlink_flush_cp_route_table() during module init.

### modules/infra/control/netlink.{c,h} — per-OIF policy routing for IPv4 and IPv6
- Introduce a dedicated control-plane routing table (GR_CP_RT_TABLE = 999) and install per-interface policy rules that match the input device (FRA_OIFNAME) to direct lookups into that table.
- Provide netlink_add_cp_route()/netlink_del_cp_route() that add/remove both the per-family (AF_INET/AF_INET6) rule and a default route (RTA_OIF=set to the interface) in table 999.
- Add netlink_flush_cp_route_table() to dump and remove stale rules/routes for table 999.
- Refactor route add/delete to a family-parameterized helper and normalize IPv6 prefix handling (use /128 for non-IPv4 addresses).

### modules/l4/l4_loopback_output.c — steer locally-handled replies to ingress TAP when cp_fd present
- Add CP_OUTPUT path and for locally-terminated replies where the originating iface has cp_fd != 0: prepend IP header, prepend Ethernet header (src/dst MAC, EtherType), clear VLAN metadata, set a control_output callback to iface_cp_tx, and enqueue to CP_OUTPUT so replies are written to the ingress TAP fd instead of always using the VRF TUN loopback. If no cp_fd, fall back to existing VRF TUN path or NO_IFACE.

## Runtime/test changes
- Add so_bindtodevice_test.sh exercising 8 bind scenarios across UDP/TCP and IPv4/IPv6 (binds to base port, VRF master, VLAN sub-interfaces, combined paths) and toggling tcp/udp l3mdev_accept where relevant.
- Add bfd_frr_test.sh and bfd6_frr_test.sh to validate BFD sessions and control-packet counters.
- Re-enable ospf/ospf6 smoke tests and update namespace/link handling (use move_to_netns helper).
- Update smoke/_init_frr.sh to enable bfdd, include bfdd in daemon readiness checks, ensure tmpfs targets exist, and make namespace teardown a TERM-then-KILL sequence.

## Operational impact (concrete gaps closed)
- Ensures frames that would have been sent NOARP are reinjected to follow ARP/NDP so peers do not observe stale MACs (fixes OSPF NOARP unicast behavior).
- Ensures reply packets for sockets bound to TAP representors are steered back to the ingress TAP (when iface exposes cp_fd), restoring bidirectional SO_BINDTODEVICE semantics (fixes bfdd and other control-plane daemons).
- Ensures TAP/VLANs in non-default VRFs are enslaved to the VRF master and that IPv4 and IPv6 per-oif policy rules plus default routes exist in table 999 so kernel routing selects the TAP as the output interface (fixes BGP-in-VRF receive/send mismatches).
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


maxime-leroy:
@rjarry and @christophefontaine help is welcome here

maxime-leroy:
Currently, grout punts all locally-destined L3 packets to the TUN loopback (main). This means SO_BINDTODEVICE sockets on the TAP (p0) cannot receive replies, since they arrive on main instead of p0.

  To fix this, grout should also punt incoming packets to the TAP of the port they arrived on. This requires prepending an Ethernet header with dst MAC = interface own MAC (similar to what the kernel NOARP does in the opposite
  direction) before writing to the TAP fd.

  This would allow bidirectional SO_BINDTODEVICE on TAP port representors (currently documented as failing in so_bindtodevice_frr_test.sh).

  The TUN loopback remains necessary for traffic not tied to a specific port (VRF addresses, loopback routes, etc.).

  P1 (NOARP MAC re-inject) and P2 (IPv6 PBR) remain necessary for the send path.

rjarry (approved):

PR #598, opened 2026-04-17
fedora 44 fixes

rjarry:
Fedora comes with recent meson (1.10.2), cmocka (2.0.2) and clang-format (22) versions which introduce issues.

## Summary

Fixes build/test/formatting incompatibilities caused by Fedora tool upgrades: meson 1.10.2, cmocka 2.0.2, and clang-format 22.

### clang-format configuration
- Set PenaltyBreakOpenParenthesis = 0 in .clang-format to change wrapping preference around `(`.
- Added localized clang-format guards and normalized directive syntax from `// clang-format: off/on` to `// clang-format off/on` where required to prevent newer clang-format from reflowing constructs that must keep specific layout:
  - Wrapped compound-literal macros in cli/cli.h with `// clang-format off`/`on`.
  - Preserved or moved clang-format guards around include boundaries and multi-line expressions in cli/main.c, frr/zebra_dplane_grout.c, smoke/fib_inject.c.
  - Replaced some previous `clang-format off` blocks by expressing the same calls as single, well-formatted expressions (frr/rt_grout.c, modules/ip/datapath/icmp_local_send.c).

### meson
- Added subprojects/.wraplock to .gitignore to avoid committing meson-generated .wraplock files.

### cmocka 2.0 API migration and test changes
- Replaced deprecated generic cmocka macros with type-specific variants in tests and unit-test-only helpers to satisfy cmocka 2.0:
  - will_return_maybe(...) → will_return_int_maybe(...)/will_return_uint_maybe(...)/will_return_ptr_maybe(...)
  - check_expected(...) → check_expected_uint(...) (or check_expected_int where appropriate)
  - expect_value(...) → expect_uint_value(...)/expect_int_value(...)
- Files updated include modules/infra/control/worker_test.c, modules/ip/datapath/ip_input.c (tests), modules/ip6/datapath/ip6_input.c (tests), modules/ip6/datapath/ndp_na_input.c (tests), and modules/infra/control/graph.h (unit-test-only macro).

### compatibility shims
- Added conditional macro aliases in main/_cmocka.h to map cmocka 2.0 type-specific macro names to older generic names when the type-specific variants are not present, preserving compatibility with cmocka < 2.0.

## Notable code adjustments (non-API)
- Simplified/reshaped nexthop lookup NULL-check control flow in modules/ip/control/route.c and modules/ip6/control/route.c (restructured assignment + NULL-check without changing return semantics).
- In frr/rt_grout.c, removed a redundant `seg6_segs != NULL` test and replaced it with an assert before using `num_segs` (enforces non-null pointer at runtime).


main/_cmocka.h
```diff
@@ -8,6 +8,17 @@
 
 #include <cmocka.h>
 
+// cmocka < 2.0 compat shims
+#ifndef will_return_int_maybe
```
maxime-leroy, line 12: does we are sure that will_return_int_maybe not defined, prevent other to be defined ?

modules/infra/control/worker_test.c
```diff
@@ -241,29 +241,29 @@ static int teardown(void **) {
...
+	will_return_int_maybe(__wrap_rte_eth_dev_info_get, 0);
+	will_return_int_maybe(__wrap_rte_eth_dev_socket_id, 0);
+	will_return_int_maybe(__wrap_rte_eth_dev_start, 0);
+	will_return_int_maybe(__wrap_rte_eth_dev_stop, 0);
+	will_return_int_maybe(__wrap_rte_eth_macaddr_get, 0);
+	will_return_int_maybe(__wrap_rte_eth_rx_queue_setup, 0);
+	will_return_int_maybe(__wrap_rte_eth_tx_queue_setup, 0);
+	will_return_int_maybe(__wrap_rte_free, 0);
+	will_return_int_maybe(__wrap_rte_mempool_free, 0);
+	will_return_int_maybe(__wrap_rte_pktmbuf_pool_create, 1);
```
maxime-leroy: int ? ptr no ?

modules/infra/control/worker_test.c
```diff
@@ -241,29 +241,29 @@ static int teardown(void **) {
...
+	will_return_int_maybe(gr_pktmbuf_pool_get, 1);
```
maxime-leroy: int ? ptr no ?

frr/zebra_dplane_grout.c
```diff
@@ -4,9 +4,9 @@
 
 #include <lib/queue.h>
 
-// clang-format: off
```
maxime-leroy, line 7: // clang-format: off -> it was not a valid format ?

modules/ip6/control/route.c
```diff
@@ -368,8 +368,12 @@ static struct api_out route6_add(const void *request, struct api_ctx *) {
...
 		if (nh == NULL)
 			return api_out(ENOENT, 0, NULL);
-	} else if ((nh = nexthop_lookup_l3(GR_AF_IP6, req->vrf_id, GR_IFACE_ID_UNDEF, &req->nh))
-		   == NULL) {
+	} else if (
+		// clang-format off
+		(nh = nexthop_lookup_l3(GR_AF_IP6, req->vrf_id, GR_IFACE_ID_UNDEF, &req->nh))
+		== NULL
+		// clang-format on
+	) {
```
maxime-leroy: // clang-format off why is needed ? it was not needed before.

maxime-leroy: } else {
      nh = nexthop_lookup_l3(GR_AF_IP6, req->vrf_id, GR_IFACE_ID_UNDEF, &req->nh);                                                                                                                                                    
      if (nh == NULL) {
          // ensure route gateway is reachable                                                                                                                                                                                        
          if ((nh = rib6_lookup(req->vrf_id, GR_IFACE_ID_UNDEF, &req->nh)) == NULL)
              return api_out(EHOSTUNREACH, 0, NULL);                                                                                                                                                                                  
          ...     
      }                                                                                                                                                                                                                               
  }

frr/zebra_dplane_grout.c
rjarry: no, only `// clang-format off` without `:`

modules/ip6/control/route.c
rjarry: because clang-format 22 changed the way this kind of expression is formatted. So I am preventing ping/pong with older clang-format versions by disabling formatting here.

main/_cmocka.h
rjarry: they were all added in cmocka 2.0

maxime-leroy (changes requested):
Consider splitting: the penalty flip and the guard additions are
logically distinct. Removing guards is mechanical; adding them
freezes formatting and deserves its own commit.

1. lower PenaltyBreakOpenParenthesis to 0 + guard removals
2. guard spots where v20/v22 disagree (additions only)

Also, "versions disagree" is vague -- name the versions and show
the divergence in the message.

The else-if guards in route.c (v4/v6) and rt_grout.c could be
avoided by splitting the assignment out or extracting a named
predicate. Keep the guard only for the cli.h compound-literal
macros.

modules/ip6/control/route.c
maxime-leroy: nh = nexthop_lookup_l3(GR_AF_IP6, req->vrf_id, GR_IFACE_ID_UNDEF, &req->nh);
if (nh == NULL) {

this format should be supported in clang-format 22 and older version ?

maxime-leroy (approved):
ack

modules/infra/control/worker_test.c
```diff
@@ -241,29 +241,29 @@ static int teardown(void **) {
...
+	will_return_int_maybe(worker_graph_free, 0);
+	will_return_int_maybe(worker_graph_reload, 0);
+	will_return_int_maybe(worker_graph_reload_all, 0);
+	will_return_int_maybe(__wrap_numa_available, 0);
+	will_return_int_maybe(__wrap_numa_node_of_cpu, 0);
+	will_return_int_maybe(__wrap_pthread_cond_wait, 0);
+	will_return_int_maybe(__wrap_pthread_cond_signal, 0);
+	will_return_int_maybe(__wrap_pthread_create, 0);
+	will_return_int_maybe(__wrap_pthread_join, 0);
+	will_return_ptr_maybe(__wrap_rte_dev_name, "");
```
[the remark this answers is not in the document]
maxime-leroy: The mock_func macro always uses mock() (reads the uint slot), but will_return_ptr_maybe on lines 253/265/266 writes to the pointer slot in cmocka 2.0+. Probably works by accident because the union overlays both slots.

@rjarry could you please check that ?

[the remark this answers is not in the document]
rjarry: This will not be possible with cmocka 1.2. I don't think I can add any compat layer.

PR #603, opened 2026-04-20
Simplify FRR build and rework CI matrix

rjarry:
Simplify the FRR meson dependency resolution by using the builtin fallback parameter instead of manually calling subproject(). Rename the wrap files to match the version they provide so the fallback name resolves naturally.

Rework the CI compiler matrix to use compact cc/cv fields with setup-gcc and setup-clang actions. Add gcc-15, gcc-16, clang-20 and clang-22 to the build matrix. Always install FRR build dependencies in all CI jobs including aarch64 cross-compilation. Mark the FRR master build as non-voting.

## Meson build changes

frr/meson.build
- Use a single Meson dependency() with fallback instead of manual subproject()/found() checks:
  - frr_dep = dependency('frr', version: '>= 10.5', required: frr_opt, fallback: ['frr-' + frr_version, 'frr_dep'])
- Early exit when the frr option is disabled remains (if frr_opt.disabled() subdir_done()).
- Read variables from frr_dep (install_on_build, moduledir) and compute CURRENT_FRR_VERSION from frr_dep.version(); pass frr_dep as the shared_module dependency.
- Add compiler flag -Wno-constant-logical-operand to frr_c_args.

subprojects/packagefiles/frr/meson.build
- Capture the detected C compiler: cc = meson.get_compiler('c').
- Introduce extra_configure_env and append a host triplet to configure options during cross builds: --host=<cpu-family>-<system>-gnu.
- Prefix the configure invocation with extra_configure_env (sets CC from cc.cmd_array(), and PKG_CONFIG_LIBDIR when provided) so the configure script runs with the Meson-detected toolchain environment.

meson_options.txt
- Change frr_version option choices and default:
  - choices: ['10.5', '10.6', 'master']
  - value: '10.5'
  - description: 'FRR version, defaults to 10.5.'

subprojects/*.wrap
- Rename/provide dependency names to versioned identifiers:
  - subprojects/frr-10.5.wrap: dependency_names = frr-10.5
  - subprojects/frr-10.6.wrap: dependency_names = frr-10.6

## CI workflow changes

.github/workflows/check.yml
- Replace previous per-entry matrix fields with a compact matrix.conf list containing compiler, frr, and optional asan/rebase flags; jobs run on ubuntu-24.04.
- Set SANITIZE via case(matrix.conf.asan || false, 'address', 'none'); set CC to "ccache ${matrix.conf.compiler}" and always include -Dfrr_version=${{ matrix.conf.frr }} in MESON_EXTRA_OPTS.
- Install FRR APT key/repo and FRR development packages unconditionally in the main build job (libjson-c-dev, libelf-dev, libprotobuf-c-dev, protobuf-c-compiler, libreadline-dev, libcap-dev, librtr-dev, libyang2-dev, libyang2-tools).
- Cross aarch64 job runs in debian:stable container, sets MESON_EXTRA_OPTS=--cross-file=devtools/cross/aarch64.ini, and installs an expanded list of cross/arm64 toolchain and FRR-related development packages (autoconf, automake, bison, flex, patch, python3-dev, crossbuild-essential-arm64, plus many :arm64 packages).

.github/workflows/publish.yml
- Extract FRR revision/version from subprojects/frr-10.5.wrap (sed ...) and expose it via steps.frr-version.outputs.version for Docker image metadata.

## Packaging / Makefile changes

GNUmakefile
- frr-rpm target now reads frr_version and frr_hash from subprojects/frr-10.5.wrap, runs meson subprojects download frr-10.5, verifies the downloaded archive checksum, and uses the derived frr_version for rpmbuild inputs and archive paths.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -28,46 +28,34 @@ jobs:
           - compiler: gcc-13
             sanitize: none
             buildtype: debugoptimized
-            os: ubuntu-24.04
             extra_opts: -Dfrr=enabled
             rebase: true
           - compiler: gcc-14
             sanitize: address
             buildtype: debug
-            os: ubuntu-24.04
             extra_opts: -Dfrr=enabled
             rebase: false
           - compiler: gcc-14
             sanitize: none
             buildtype: debugoptimized
-            os: ubuntu-24.04
             extra_opts: -Dfrr=enabled -Dfrr_version=10.6
             rebase: false
           - compiler: gcc-14
             sanitize: none
             buildtype: debugoptimized
-            os: ubuntu-24.04
             extra_opts: -Dfrr=enabled -Dfrr_version=master
-            rebase: false
-          - compiler: clang-15
-            sanitize: none
-            buildtype: debugoptimized
-            os: ubuntu-22.04
-            extra_opts:
             rebase: false
           - compiler: clang-16
             sanitize: none
             buildtype: debugoptimized
-            os: ubuntu-24.04
             extra_opts:
             rebase: false
           - compiler: clang-18
             sanitize: none
             buildtype: debugoptimized
-            os: ubuntu-24.04
             extra_opts:
             rebase: false
-    runs-on: ${{ matrix.conf.os }}
+    runs-on: ubuntu-24.04
     env:
       SANITIZE: ${{ matrix.conf.sanitize }}
       BUILDTYPE: ${{ matrix.conf.buildtype }}
@@ -97,10 +85,6 @@ jobs:
               $(lsb_release -s -c) frr-stable" | sudo tee /etc/apt/sources.list.d/frr.list > /dev/null
             sudo apt-get update -qy
             sudo apt-get install -qy librtr-dev libyang2-dev libyang2-tools
-          fi
-          if printf '%s\n0.63.0\n' "$(meson --version)" | sort -V -C ; then
-            sudo apt install -y python3-pip
-            pip3 install --user -U meson~=0.63.0
           fi
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -41,7 +41,7 @@ jobs:
             sanitize: none
             buildtype: debugoptimized
             os: ubuntu-24.04
-            extra_opts: -Dfrr=enabled -Dfrr_version=10.6.0
+            extra_opts: -Dfrr=enabled -Dfrr_version=10.6
             rebase: false
           - compiler: gcc-14
             sanitize: none
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -26,39 +26,24 @@ jobs:
       matrix:
         conf:
           - compiler: gcc-13
-            sanitize: none
-            buildtype: debugoptimized
             extra_opts: -Dfrr=enabled
             rebase: true
           - compiler: gcc-14
-            sanitize: address
-            buildtype: debug
             extra_opts: -Dfrr=enabled
-            rebase: false
+            asan: true
+            debug: true
           - compiler: gcc-14
-            sanitize: none
-            buildtype: debugoptimized
             extra_opts: -Dfrr=enabled -Dfrr_version=10.6
-            rebase: false
           - compiler: gcc-14
-            sanitize: none
-            buildtype: debugoptimized
             extra_opts: -Dfrr=enabled -Dfrr_version=master
-            rebase: false
           - compiler: clang-16
-            sanitize: none
-            buildtype: debugoptimized
             extra_opts:
-            rebase: false
           - compiler: clang-18
-            sanitize: none
-            buildtype: debugoptimized
             extra_opts:
-            rebase: false
     runs-on: ubuntu-24.04
     env:
-      SANITIZE: ${{ matrix.conf.sanitize }}
-      BUILDTYPE: ${{ matrix.conf.buildtype }}
+      SANITIZE: ${{ case(matrix.conf.asan || false, 'address', 'none') }}
+      BUILDTYPE: ${{ case(matrix.conf.debug || false, 'debug', 'debugoptimized') }}
       DEBIAN_FRONTEND: noninteractive
       NEEDRESTART_MODE: l
       CC: ccache ${{ matrix.conf.compiler }}
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -26,20 +26,20 @@ jobs:
       matrix:
         conf:
           - compiler: gcc-13
-            extra_opts: -Dfrr=enabled
+            frr: '10.5'
             rebase: true
           - compiler: gcc-14
-            extra_opts: -Dfrr=enabled
+            frr: '10.5'
             asan: true
             debug: true
           - compiler: gcc-14
-            extra_opts: -Dfrr=enabled -Dfrr_version=10.6
+            frr: '10.6'
           - compiler: gcc-14
-            extra_opts: -Dfrr=enabled -Dfrr_version=master
+            frr: 'master'
           - compiler: clang-16
-            extra_opts:
+            frr: '10.5'
           - compiler: clang-18
-            extra_opts:
+            frr: '10.5'
     runs-on: ubuntu-24.04
     env:
       SANITIZE: ${{ case(matrix.conf.asan || false, 'address', 'none') }}
@@ -47,11 +47,13 @@ jobs:
       DEBIAN_FRONTEND: noninteractive
       NEEDRESTART_MODE: l
       CC: ccache ${{ matrix.conf.compiler }}
-      MESON_EXTRA_OPTS: -Ddpdk:cpu_instruction_set=generic ${{ matrix.conf.extra_opts }}
+      FRR: ${{ matrix.conf.frr }}
+      MESON_EXTRA_OPTS: -Ddpdk:cpu_instruction_set=generic
     steps:
       - name: install system dependencies
         run: |
           set -xe
+          # grout dependencies
           sudo apt-get update -qy
           sudo apt-get install -qy --no-install-recommends \
             make gcc gdb ccache ninja-build meson git scdoc \
@@ -60,17 +62,15 @@ jobs:
             socat tcpdump traceroute graphviz iproute2 iputils-ping ndisc6 jq \
             dnsmasq systemd-coredump abigail-tools \
             "linux-modules-extra-$(uname -r)"
-          if echo $MESON_EXTRA_OPTS | grep -q frr=enabled ; then
-            sudo apt-get install -qy --no-install-recommends \
-              libjson-c-dev libelf-dev libprotobuf-c-dev protobuf-c-compiler \
-              libreadline-dev libcap-dev
-            sudo mkdir -p /etc/apt/keyrings
-            sudo curl -s -o /etc/apt/keyrings/frrouting.gpg https://deb.frrouting.org/frr/keys.gpg
-            echo "deb [signed-by=/etc/apt/keyrings/frrouting.gpg] https://deb.frrouting.org/frr \
-              $(lsb_release -s -c) frr-stable" | sudo tee /etc/apt/sources.list.d/frr.list > /dev/null
-            sudo apt-get update -qy
-            sudo apt-get install -qy librtr-dev libyang2-dev libyang2-tools
-          fi
+          # FRR dependencies
+          sudo mkdir -p /etc/apt/keyrings
+          sudo curl -s -o /etc/apt/keyrings/frrouting.gpg https://deb.frrouting.org/frr/keys.gpg
+          echo "deb [signed-by=/etc/apt/keyrings/frrouting.gpg] https://deb.frrouting.org/frr \
+            $(lsb_release -s -c) frr-stable" | sudo tee /etc/apt/sources.list.d/frr.list > /dev/null
+          sudo apt-get update -qy
+          sudo apt-get install -qy --no-install-recommends \
+            libjson-c-dev libelf-dev libprotobuf-c-dev protobuf-c-compiler \
+            libreadline-dev libcap-dev librtr-dev libyang2-dev libyang2-tools
       - uses: actions/checkout@v4
         with:
           persist-credentials: false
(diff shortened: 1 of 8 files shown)


.github/workflows/check.yml
```diff
@@ -148,7 +148,7 @@ jobs:
     runs-on: ubuntu-24.04
     container: debian:stable
     env:
-      MESON_EXTRA_OPTS: --cross-file=devtools/cross/aarch64.ini
+      MESON_EXTRA_OPTS: --cross-file=devtools/cross/aarch64.ini -Dfrr=disabled
```
maxime-leroy: why this change in this commit ?

It missing at least some explanation in the commit log about frr=auto does before this commit, what it does after.

frr/meson.build
```diff
@@ -1,26 +1,23 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 # Copyright (c) 2025 Maxime Leroy, Free Mobile
-frr_opt = get_option('frr')
-frr_version = get_option('frr_version')
 
-if frr_opt.disabled()
+if not get_option('frr').allowed()
```
maxime-leroy: get_option('frr').disabled() ? why this change

meson_options.txt
```diff
@@ -12,7 +12,7 @@ option(
 )
 
 option(
-  'frr_version', type: 'combo', choices: ['default', '10.6.0', 'master'], value: 'default',
+  'frr_version', type: 'combo', choices: ['default', '10.6', 'master'], value: 'default',
```
maxime-leroy: what is the link with this commit ?

meson_options.txt
```diff
@@ -12,7 +12,7 @@ option(
 )
```
maxime-leroy, line 12: [subprojects/frr-10.6.0.wrap‎ ‎-> subprojects/frr-10.6.wrap

[provide]                                                                                                                                                                                                                                                  
 - dependency_names = frr-10.6.0 
 + dependency_names = frr-10.6

frr/meson.build
```diff
@@ -1,26 +1,23 @@
...
-    if frr_version == 'default'
-      sp = subproject('frr')
-    else
-      sp = subproject('frr-' + frr_version)
-    endif
-    frr_dep = sp.get_variable('frr_dep')
-  else
-    subdir_done()
-  endif
+if get_option('frr_version') == 'default'
```
maxime-leroy: Why keep the default magic value + frr.wrap indirection? Since the commit already renames frr-10.6.0.wrap to match its option choice, it would be consistent to rename frr.wrap to frr-10.5.wrap and change the default option value from 'default' to '10.5'. That removes the if frr_version == 'default' branch entirely and makes all choices homogeneous.

maxime-leroy (changes requested):

.github/workflows/check.yml
```diff
@@ -25,83 +25,76 @@ jobs:
...
+          - cc: clang
+            cv: 18
+            frr: default
+          - cc: clang
+            cv: 20
+            frr: default
+          - cc: clang
+            cv: 21
+            frr: default
+    runs-on: ubuntu-24.04
```
maxime-leroy, line 43: Could you please provide in the commit log a list of CI job started before this commits, after this one.

.github/workflows/check.yml
```diff
@@ -25,83 +25,76 @@ jobs:
...
     env:
-      SANITIZE: ${{ matrix.conf.sanitize }}
-      BUILDTYPE: ${{ matrix.conf.buildtype }}
-      DEBIAN_FRONTEND: noninteractive
-      NEEDRESTART_MODE: l
-      CC: ccache ${{ matrix.conf.compiler }}
-      MESON_EXTRA_OPTS: -Ddpdk:cpu_instruction_set=generic ${{ matrix.conf.extra_opts }}
+      SANITIZE: ${{ case(matrix.conf.sanitize || false, 'address', 'none') }}
```
maxime-leroy: BUILDTYPE is gone ?

.github/workflows/check.yml
```diff
@@ -25,83 +25,76 @@ jobs:
...
+            cv: 16
+            frr: default
+          - cc: clang
+            cv: 18
+            frr: default
+          - cc: clang
+            cv: 20
+            frr: default
+          - cc: clang
+            cv: 21
```
maxime-leroy: Clang 22 (commit log) or Clang 21 ?

.github/workflows/check.yml
```diff
@@ -25,83 +25,76 @@ jobs:
...
           sudo apt-get update -qy
           sudo apt-get install -qy --no-install-recommends \
-            make gcc gdb ccache ninja-build meson git scdoc \
+            make gdb ccache ninja-build meson git scdoc \
             libibverbs-dev libasan8 libcmocka-dev libedit-dev libarchive-dev \
             libevent-dev libmnl-dev libnuma-dev python3-pyelftools \
             socat tcpdump traceroute graphviz iproute2 iputils-ping ndisc6 jq \
             dnsmasq systemd-coredump abigail-tools \
             "linux-modules-extra-$(uname -r)"
-          if echo $MESON_EXTRA_OPTS | grep -q frr=enabled ; then
```
maxime-leroy, line 90: Why do you install FRR dependencies if not needed (aarch64.ini -Dfrr=disabled) ?

.github/workflows/check.yml
```diff
@@ -25,83 +25,76 @@ jobs:
...
+          - cc: gcc
+            cv: 14
+            frr: master
+            voting: false
+          - cc: gcc
+            cv: 15
+            frr: default
+          - cc: gcc
+            cv: 16
+            frr: default
```
maxime-leroy: clang-15, clang-16 are gone ?

could you please do at least 2 commits one  for matrix, one to del/add jobs.

.github/workflows/check.yml
```diff
@@ -25,83 +25,76 @@ jobs:
...
+          - cc: clang
+            cv: 21
+            frr: default
+    runs-on: ubuntu-24.04
     env:
-      SANITIZE: ${{ matrix.conf.sanitize }}
-      BUILDTYPE: ${{ matrix.conf.buildtype }}
-      DEBIAN_FRONTEND: noninteractive
-      NEEDRESTART_MODE: l
```
maxime-leroy, line 75: DEBIAN_FRONTEND and  NEEDRESTART_MODE are gone ?

.github/workflows/check.yml
rjarry: I'll fix the commit log. DPDK 25.11 does not build with clang 22

.github/workflows/check.yml
rjarry: Not needed. I can split that into a separate commit. But honestly, this is CI stuff. I didn't think it was worth it.

.github/workflows/check.yml
rjarry: OK

.github/workflows/check.yml
rjarry: This job is for x86. FRR is always enabled.

.github/workflows/check.yml
rjarry: Yes, there is no point in building with -O0. We don't need to run step gdb.

.github/workflows/check.yml
rjarry: Ack

frr/meson.build
rjarry: I had originally done it this way. But it requires modifying lots of other files to use the correct .wrap file. (GNUmakefile, etc.)

I can do it if you prefer.

meson_options.txt
rjarry: ack

.github/workflows/check.yml
```diff
@@ -149,8 +142,6 @@ jobs:
     container: debian:stable
     env:
       MESON_EXTRA_OPTS: --cross-file=devtools/cross/aarch64.ini -Dfrr=disabled
```
maxime-leroy: Does we know in github job that FRR is disabled on ARM ? (with the CI title job). Could you please fix it ?

frr/meson.build
rjarry: Not needed, I'll revert it.

.github/workflows/check.yml
rjarry: Because `-Dfrr=auto` now automatically builds the frr subproject if no headers are found on the system. And our local meson.build for FRR does not support cross-compilation.

.github/workflows/check.yml
rjarry: I can do better: patch the FRR meson.build local file to support cross compilation.

.github/workflows/check.yml
```diff
@@ -114,11 +107,11 @@ jobs:
         id: cache
         with:
           path: ${{ env.CCACHE_DIR }}
-          key: ccache-x86_64-${{ matrix.conf.compiler }}
+          key: ccache-x86_64-${{ matrix.conf.cc }}-${{ matrix.conf.cv }}
       - run: ccache -z
-      - run: git rebase -x .github/workflows/check.sh "HEAD~${{ github.event.pull_request.commits }}"
+      - run: git rebase -x .github/workflows/check.sh "HEAD~${{ github.event.pull_request.commits }}" || [ "${{ matrix.conf.voting }}" = false ]
```
maxime-leroy: The current || [ "$voting" = false ] trick hides the failure by exiting 0 — the check appears green in the PR status, so a reviewer scanning results sees nothing amiss when FRR master breaks. That defeats the purpose of having a canary build.

For "non-blocking but visible" you really need two things: (1) let the job fail normally (red), and (2) configure branch protection to exclude that specific check from required status checks. Usually this means splitting non-voting entries into a  separate job (build-and-tests-experimental or similar) so the check name is stable and can be carved out of branch protection.

rjarry: We do not have any branch protection for the moment. And unfortunately, github does not have a "non-voting" option for jobs. Do you think it is worth it? I can make this job voting maybe. And we can decide on a case by case basis if failures are worth fixing?

maxime-leroy: Without branch protection, "voting" doesn't actually block merge anyway - the button stays active regardless. Making it voting gives you visibility without any technical blocking cost: you can still merge through it if the failure is unrelated to  the PR.

.github/workflows/check.yml
```diff
@@ -27,38 +27,31 @@ jobs:
         conf:
           - compiler: gcc-13
             sanitize: none
-            buildtype: debugoptimized
             extra_opts: -Dfrr=enabled
             rebase: true
           - compiler: gcc-14
             sanitize: address
-            buildtype: debug
```
maxime-leroy, line 36: There is no point in building with -O0  -> I have already seen some code that compiled in O2/O3 but not in O0. Compiler optimized enough to remove code that should not compile.

frr/meson.build
```diff
@@ -26,6 +26,7 @@ frr_ver = frr_version.split('.')
 frr_ver_num = frr_ver[0].to_int() * 16777216 + frr_ver[1].to_int() * 256 + frr_ver[2].to_int()
 frr_c_args = [
   '-DCURRENT_FRR_VERSION=' + frr_ver_num.to_string(),
+  '-Wno-constant-logical-operand',
```
maxime-leroy, line 27: subprojects/packagefiles/frr/meson.build ->   '-Wno-constant-logical-operand' should be in this file.
With other compilations option of FRR:
```
frr_dep = declare_dependency(
  include_directories: include_directories('.', 'lib'),
  compile_args: [
    '-fms-extensions',
    '-DMULTIPATH_NUM=256',
    '-Wno-missing-field-initializers',
    '-Wno-unused-parameter',
+  '-Wno-constant-logical-operand',
  ],
  sources: [build],
  version: meson.project_version(),
  variables: {
    'moduledir': moduledir,
    'prefix': prefix,
    'install_on_build': 'true',
  }
)
```

.github/workflows/check.yml
```diff
@@ -26,49 +26,48 @@ jobs:
...
+      MESON_EXTRA_OPTS: -Ddpdk:cpu_instruction_set=generic -Dfrr_version=${{ matrix.conf.frr }}
     steps:
       - name: install system dependencies
         run: |
           set -xe
+          # grout dependencies
           sudo apt-get update -qy
           sudo apt-get install -qy --no-install-recommends \
-            make gcc gdb ccache ninja-build meson git scdoc \
+            make gdb ccache ninja-build meson git scdoc \
```
maxime-leroy: This change is unrelated to this commit.

subprojects/packagefiles/frr/meson.build
```diff
@@ -33,19 +35,31 @@ moduledir = prefix / get_option('libdir') / 'frr/modules'
...
 extra_configure_option = '--enable-user=' + user_name + ' ' + '--enable-group=' + group_name
+extra_configure_env = ''
 
 buildtype = get_option('buildtype')
 debug = buildtype.startswith('debug')
 if debug
   extra_configure_option += ' --enable-dev-build'
 endif
 
+if meson.is_cross_build()
```
maxime-leroy, line 46: -Dfrr=disabled  is removed for aarch64 job ?

rjarry: yes, we now cross compile the frr plugin for aarch64

frr/meson.build
rjarry: This wouldn't be consistent with what is exported in the upstream pkgconfig file:

https://github.com/FRRouting/frr/blob/c46e6f24bd114e9157c4c1776b6e7f5ba10eaef8/tools/frr.pc.in#L17-L19

maxime-leroy: Please do a PR on frr upstream project. 


Does FRR support clang compilation at least ?

maxime-leroy: ack for this.

rjarry: > Please do a PR on frr upstream project.

This seems to be a quirk of clang 16. It works fine on clang 18. I am not sure it is worth the hassle.

> Does FRR support clang compilation at least ?

Probably not.

.github/workflows/check.yml
rjarry: I restored one job with gcc-14 and debug.

frr/meson.build
maxime-leroy: ack

GNUmakefile
```diff
@@ -142,13 +142,13 @@ rpm:
 		mv -vf ~/rpmbuild/RPMS/$$arch/grout-frr-debuginfo-$$version.$$arch.rpm grout-frr-debuginfo.$$arch.rpm; \
 	fi
 
-frr_version = $(shell sed -nE 's/^source_filename = frr-(.+)\.tar\.gz$$/\1/p' subprojects/frr.wrap)
-frr_hash = $(shell sed -nE 's/^source_hash = //p' subprojects/frr.wrap)
+frr_version = $(shell sed -nE 's/^source_filename = frr-(.+)\.tar\.gz$$/\1/p' subprojects/frr-10.5.wrap)
+frr_hash = $(shell sed -nE 's/^source_hash = //p' subprojects/frr-10.5.wrap)
```
maxime-leroy: hardcoded  '10.5' value.

instead:
```
  frr_default_version ?= 10.5
  frr_wrap    = subprojects/frr-$(frr_default_version).wrap                                                                                                                
  frr_version = $(shell sed -nE 's/^source_filename = frr-(.+)\.tar\.gz$$/\1/p' $(frr_wrap))                                                                               
  frr_hash    = $(shell sed -nE 's/^source_hash = //p' $(frr_wrap))                                                                                                        
  frr_archive = subprojects/packagecache/frr-$(frr_version).tar.gz                                                                                                         
                                                                                                                                                                           
  .PHONY: frr-rpm                                                                                                                                                          
  frr-rpm:                                                                                                                                                                 
        meson subprojects download frr-$(frr_default_version)                                                                                                          
    ```

File: GNUmakefile [hunk 19]
```diff
@@ -142,13 +142,13 @@ rpm:
...
 frr_archive = subprojects/packagecache/frr-$(frr_version).tar.gz
 
 .PHONY: frr-rpm
 frr-rpm:
-	meson subprojects download frr
+	meson subprojects download frr-10.5
```
maxime-leroy: same here use a global variable in Makefile

I understand there is not easy way to get meson default value Here.

maxime-leroy (approved):
ack

PR #605, opened 2026-04-22
frr: sweep stale grout routes across zebra restart

maxime-leroy:
When zebra is killed and restarted while grout keeps running, grout's FIB retains every route the previous FRR instance had installed. With the current plugin those routes become orphans: the startup sync explicitly drops every self-originated route returned by the grout dump (rt_grout.c:337) so zebra's RIB never sees them, and no mechanism ever garbage-collects them from grout. Unreclaimed prefixes linger indefinitely and can blackhole traffic to stale nexthops.

Upstream FRR solves the same problem for the kernel dplane through the ZEBRA_FLAG_SELFROUTE flag, rib_sweep_route and the GR_RUN metaQ sub-queue: kernel-read routes are re-inserted as selfroute placeholders with a pre-startup uptime; protocol clients take them over via rib_compare_routes when they re-announce; anything unreclaimed is purged when rib_sweep_route fires after -K seconds. The grout plugin could not reuse that directly because grout_sync runs from frr_config_post, which is after zebra_main_router_started() has already stamped zrouter.startup_time and armed t_rib_sweep. Entries created in grout_sync therefore have uptime > startup_time and would never match the sweep predicate, and the sweep itself is already scheduled.

Transplant the mechanism into the plugin:

* Thread a bool startup flag through grout_route4_change / grout_route6_change / grout_route_change (mirroring the existing grout_nexthop_change signature), true on the initial dump path, false on asynchronous notifications. The blanket self-origin skip stays for the notification case (an echo of an install we sent) but becomes injection for the dump case: build the route_entry with ZEBRA_FLAG_SELFROUTE and force re->uptime = zrouter.startup_time so the sweep predicate SELFROUTE && uptime <= startup_time matches, and rib_compare_routes will replace the placeholder cleanly when a client reclaims the prefix.

* In grout_add_del_route, short-circuit DPLANE_OP_ROUTE_INSTALL and DPLANE_OP_ROUTE_UPDATE when the ctx carries ZEBRA_FLAG_SELFROUTE. Those are the routes we just learned from grout; pushing them back down would be a redundant round-trip. DPLANE_OP_ROUTE_DELETE is left alone so the sweep's uninstall reaches grout via the normal flow.

* At the very end of the sync chain (last VRF of grout_sync_routes), cancel zrouter.t_rib_sweep and re-arm it with delay equal to the operator's -K value. This aligns the sweep on sync_complete + -K instead of startup_time + -K, giving our backdated SELFROUTE entries the full reclaim window the operator asked for. The -K value is recovered by re-parsing /proc/self/cmdline with getopt_long on a minimal option table: zebra_di.gr_cleanup_time is static in main.c and not exported, re-invoking frr_getopt would re-run frr_opt() with its side effects (exit on -h/-v, double-allocation of modules, errors++ path), and zrouter.t_rib_sweep loses the -K value once it has fired, so none of those routes are reliable. /proc/self/cmdline always holds the original argv. When -K was passed without a value the plugin falls back to ZEBRA_GR_DEFAULT_RIB_SWEEP_TIME exactly as zebra itself does; non-GR startups keep delay = 0 (fire immediately) to match kernel-dplane wipe semantics.

With these pieces in place, zebra's existing SELFROUTE machinery works uniformly over grout-injected entries:

  * Reclaim: clients reconnecting over zapi take over matching SELFROUTE entries via rib_compare_routes; the fresh entry loses SELFROUTE and has a current uptime.
  * GR_RUN (metaQ sub-q 10): per-client early cleanup triggered by ZEBRA_CLIENT_ROUTE_UPDATE_COMPLETE works automatically because our entries carry the real protocol type (via origin2zebra) and a backdated uptime (older than client->restart_time).
  * rib_sweep_route catch-all: reprogrammed to fire after the full grout sync chain completes, purges any SELFROUTE entry nobody reclaimed, and the subsequent DPLANE_OP_ROUTE_DELETE ctxs flow through grout_add_del_route to remove the routes from grout's FIB (the install guard above is install/update only).

## Problem
When zebra is killed and restarted while grout remains running, grout's FIB retains routes installed by the previous FRR instance. The plugin’s startup sync currently drops self-originated routes returned by the grout dump so zebra’s RIB never sees them; without a reclamation path these prefixes persist in grout and can blackhole traffic.

## Technical changes

Route handling during startup sync
- Thread a startup boolean through grout_route4_change(), grout_route6_change(), and grout_route_change() (true for initial dump, false for async notifications).
- For dump-path entries, set ZEBRA_FLAG_SELFROUTE on the created route_entry and backdate re->uptime to zrouter.startup_time so they match the sweep predicate (SELFROUTE && uptime <= startup_time) and are reclaimable when protocol clients reannounce.
- Preserve existing self-origin skip behavior for notification-path entries.

Dplane operations
- In grout_add_del_route(), short-circuit DPLANE_OP_ROUTE_INSTALL and DPLANE_OP_ROUTE_UPDATE when the dplane context carries ZEBRA_FLAG_SELFROUTE (return success without redundant install/update round-trips).
- Leave DPLANE_OP_ROUTE_DELETE unchanged so rib_sweep_route can uninstall unreclaimed SELFROUTE entries.
- Add a targeted delete short-circuit for the SRTE end-of-replay sync marker to avoid needless uninstall round-trips.

Nexthop handling
- When startup is true, lookup corresponding nexthop-hash entries and backdate their uptime to zrouter.startup_time; log a warning if the nexthop entry cannot be found.

Sweep rescheduling and sync marker
- Append an IPv6 self-route sync marker (::/128) tagged with GROUT_SYNC_MARKER_TAG during replay to detect end-of-replay visibility.
- After the last VRF in grout_sync_routes completes, cancel any prior zrouter.t_rib_sweep and re-arm it with a delay equal to the operator’s -K value (re-parsed from /proc/self/cmdline once via getopt_long and cached; fallback to ZEBRA_GR_DEFAULT_RIB_SWEEP_TIME if -K provided without a value; non-GR startups keep delay = 0).
- Defer arming rib_sweep_route until the marker is observed (or max retry reached), so backdated SELFROUTE entries receive the intended reclaim window; delete the marker once observed.

API/signature changes
- grout_route4_change(bool new, struct gr_ip4_route *gr_r4) → grout_route4_change(bool new, struct gr_ip4_route *gr_r4, bool startup)
- grout_route6_change(bool new, struct gr_ip6_route *gr_r6) → grout_route6_change(bool new, struct gr_ip6_route *gr_r6, bool startup)
- grout_route_change(...) now accepts a bool startup parameter.
- Add public constant GROUT_SYNC_MARKER_TAG.

Tests and infra
- New smoke tests:
  - smoke/frr_restart_graceful_frr_test.sh: validates graceful restart with -K5, ensuring orphan routes are swept while configured routes persist and are reclassified in zebra.
  - smoke/frr_restart_sweep_no_k_frr_test.sh: validates sweep behavior when zebra is started without -K.
  - smoke/grout_restart_resync_frr_test.sh: validates grout self-resynchronization after grout restart while FRR remains running.
- smoke/_init_frr.sh: keep flog variable non-local so tests can inspect log file after start.

Behavioral outcomes
- Reconnecting protocol clients reclaim matching SELFROUTE entries via rib_compare_routes() (SELFROUTE removed, uptime updated to current).
- Per-client early cleanup functions operate correctly because entries carry protocol type and backdated uptime.
- rib_sweep_route, after the rescheduled delay, purges unreclaimed SELFROUTE entries and causes DPLANE_OP_ROUTE_DELETE flows that remove routes from grout’s FIB.

Status
- Work in progress: unit tests still in development; author notes upstream FRR PR FRRouting/frr#21550 addresses similar kernel-dplane startup cases and recommends auditing NHG cleanup across zebra restarts as follow-up.

### DIFF
--- a/frr/rt_grout.c
+++ b/frr/rt_grout.c
@@ -324,10 +324,11 @@ static void grout_route_change(
 	uint16_t family,
 	void *dest_addr,
 	uint8_t dest_prefixlen,
+	uint16_t gr_route_vrf_id,
 	struct gr_nexthop *gr_nh,
 	bool startup
 ) {
-	uint32_t vrf_id = vrf_grout_to_frr(gr_nh->vrf_id);
+	uint32_t vrf_id = vrf_grout_to_frr(gr_route_vrf_id);
 	bool selfroute = is_selfroute(origin);
 	int proto = ZEBRA_ROUTE_KERNEL;
 	uint32_t nh_id = gr_nh->nh_id;
@@ -440,12 +441,13 @@ static void grout_route_change(
 
 void grout_route4_change(bool new, struct gr_ip4_route *gr_r4, bool startup) {
 	gr_log_debug(
-		"%s %pI4/%u origin %s nh_id %u vrf %u",
+		"%s %pI4/%u origin %s nh_id %u route_vrf %u nh_vrf %u",
 		new ? "add" : "del",
 		&gr_r4->dest.ip,
 		gr_r4->dest.prefixlen,
 		gr_nh_origin_name(gr_r4->origin),
 		gr_r4->nh.nh_id,
+		gr_r4->vrf_id,
 		gr_r4->nh.vrf_id
 	);
 	grout_route_change(
@@ -454,6 +456,7 @@ void grout_route4_change(bool new, struct gr_ip4_route *gr_r4, bool startup) {
 		AF_INET,
 		(void *)&gr_r4->dest.ip,
 		gr_r4->dest.prefixlen,
+		gr_r4->vrf_id,
 		&gr_r4->nh,
 		startup
 	);
@@ -461,12 +464,13 @@ void grout_route4_change(bool new, struct gr_ip4_route *gr_r4, bool startup) {
 
 void grout_route6_change(bool new, struct gr_ip6_route *gr_r6, bool startup) {
 	gr_log_debug(
-		"%s %pI6/%u origin %s nh_id %u vrf %u",
+		"%s %pI6/%u origin %s nh_id %u route_vrf %u nh_vrf %u",
 		new ? "add" : "del",
 		&gr_r6->dest.ip,
 		gr_r6->dest.prefixlen,
 		gr_nh_origin_name(gr_r6->origin),
 		gr_r6->nh.nh_id,
+		gr_r6->vrf_id,
 		gr_r6->nh.vrf_id
 	);
 	grout_route_change(
@@ -475,6 +479,7 @@ void grout_route6_change(bool new, struct gr_ip6_route *gr_r6, bool startup) {
 		AF_INET6,
 		(void *)&gr_r6->dest.ip,
 		gr_r6->dest.prefixlen,
+		gr_r6->vrf_id,
 		&gr_r6->nh,
 		startup
 	);
--- a/frr/rt_grout.c
+++ b/frr/rt_grout.c
@@ -279,6 +279,7 @@ static int grout_gr_nexthop_to_frr_nexthop(
 
 		ctx.table = vrf_grout_to_frr(sr6->out_vrf_id);
 		nexthop_add_srv6_seg6local(nh, action, &ctx);
+		nh->type = nh->ifindex ? NEXTHOP_TYPE_IPV6_IFINDEX : NEXTHOP_TYPE_IPV6;
 		*nh_family = AF_INET6;
 		break;
 	}
@@ -298,12 +299,14 @@ static int grout_gr_nexthop_to_frr_nexthop(
 		}
 
 		nexthop_add_srv6_seg6(nh, (void *)sr6->seglist, sr6->n_seglist, encap_behavior);
+		nh->type = nh->ifindex ? NEXTHOP_TYPE_IPV6_IFINDEX : NEXTHOP_TYPE_IPV6;
 		*nh_family = AF_INET6;
 		break;
 	}
 	case GR_NH_T_GROUP:
 		nh->ifindex = ifindex_grout_to_frr(gr_nh->iface_id);
 		nh->vrf_id = vrf_grout_to_frr(gr_nh->vrf_id);
+		nh->type = NEXTHOP_TYPE_IFINDEX;
 		*nh_family = AF_UNSPEC;
 		nh->weight = 1;
 		break;
--- a/frr/rt_grout.c
+++ b/frr/rt_grout.c
@@ -321,9 +321,11 @@ static void grout_route_change(
 	uint16_t family,
 	void *dest_addr,
 	uint8_t dest_prefixlen,
-	struct gr_nexthop *gr_nh
+	struct gr_nexthop *gr_nh,
+	bool startup
 ) {
 	uint32_t vrf_id = vrf_grout_to_frr(gr_nh->vrf_id);
+	bool selfroute = is_selfroute(origin);
 	int proto = ZEBRA_ROUTE_KERNEL;
 	uint32_t nh_id = gr_nh->nh_id;
 	// Grout has no per‑VRF routing tables; table_id always equals vrf_id
@@ -334,8 +336,12 @@ static void grout_route_change(
 	size_t sz;
 	afi_t afi;
 
-	if (new && is_selfroute(origin)) {
-		gr_log_debug("self-originated %s route, skip", gr_nh_origin_name(origin));
+	// Echo of our own install: skip. Mirrors the !startup && selfroute
+	// guard in netlink_route_change (zebra/rt_netlink.c).
+	if (new && !startup && selfroute) {
+		gr_log_debug(
+			"self-originated %s route notification, skip", gr_nh_origin_name(origin)
+		);
 		return;
 	}
 
@@ -383,6 +389,11 @@ static void grout_route_change(
 
 	proto = origin2zebra(origin, family, false);
 
+	// Route inserted by Zebra: mark SELFROUTE so rib_compare_routes can
+	// swap with a client's re-push and rib_sweep_route can purge.
+	if (selfroute)
+		flags |= ZEBRA_FLAG_SELFROUTE;
+
 	if (new) {
 		struct route_entry *re;
 		struct nexthop_group *ng = NULL;
@@ -424,7 +435,7 @@ static void grout_route_change(
 	}
 }
 
-void grout_route4_change(bool new, struct gr_ip4_route *gr_r4) {
+void grout_route4_change(bool new, struct gr_ip4_route *gr_r4, bool startup) {
 	gr_log_debug(
 		"%s %pI4/%u origin %s nh_id %u vrf %u",
 		new ? "add" : "del",
@@ -440,11 +451,12 @@ void grout_route4_change(bool new, struct gr_ip4_route *gr_r4) {
 		AF_INET,
 		(void *)&gr_r4->dest.ip,
 		gr_r4->dest.prefixlen,
-		&gr_r4->nh
+		&gr_r4->nh,
+		startup
 	);
 }
 
-void grout_route6_change(bool new, struct gr_ip6_route *gr_r6) {
+void grout_route6_change(bool new, struct gr_ip6_route *gr_r6, bool startup) {
 	gr_log_debug(
 		"%s %pI6/%u origin %s nh_id %u vrf %u",
 		new ? "add" : "del",
@@ -460,7 +472,8 @@ void grout_route6_change(bool new, struct gr_ip6_route *gr_r6) {
 		AF_INET6,
 		(void *)&gr_r6->dest.ip,
 		gr_r6->dest.prefixlen,
-		&gr_r6->nh
+		&gr_r6->nh,
+		startup
 	);
 }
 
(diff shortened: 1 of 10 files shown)


frr/zebra_dplane_grout.c
```diff
@@ -129,6 +133,54 @@ static void grout_sync_fdb(struct event *) {
...
+	int saved_optind = optind;
+	int saved_opterr = opterr;
+	optind = 0;
+	opterr = 0;
+	int ret = -1;
+	int c;
+	while ((c = getopt_long(argc, argv, "-:K::", lo, NULL)) != -1) {
+		if (c == 'K' && optarg)
+			ret = atoi(optarg);
+	}
```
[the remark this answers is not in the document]
maxime-leroy: Intentional - this mirrors lib/libfrr.c:133 which uses the same "K::"
shortopts and the same optional_argument long option at lib/libfrr.c:119.
An operator invoking "zebra -K 30" (with a space) does not get 30 honored
by zebra itself either: optarg is NULL, and "30" is left as a positional
argv element. Matching that behavior is the correct thing here.

Switching to "K:" (required_argument) as the AI-agent prompt suggests
would break the bare "-K" case that zebra supports. Replacing the
getopt_long loop with the proposed manual argv walk would diverge from
zebra's own parsing: the plugin would honor "-K 30" while zebra would
not, producing inconsistent graceful-restart semantics between the two.

rjarry (approved):

maxime-leroy:
I am still working on a unit test, please don't merge it.

maxime-leroy:
Note: FRRouting/frr#21550 covers the same startup-after-crash class on the kernel-dplane side. It also includes a proto-owned NHG refcount fix. NHG cleanup across zebra restarts should be audited for grout as well (follow-up).

maxime-leroy:
WIP STILL

maxime-leroy:
This PR must be carefully review it before any merge please.

@christophefontaine @rjarry @vjardin

vjardin:
what's happen if in between, the marker is injected, there is a crash, then restart with another marker injected (while the former one is not removed) ?

maxime-leroy:
> what's happen if in between, the marker is injected, there is a crash, then restart with another marker injected (while the former one is not removed) ?
> 
>

Grout crash:

The old marker does not survive a grout restart because the reconnect path wipes zebra’s RIB through "grout_ns_reset() -> vrf_terminate()". The marker is removed along with all the other routes before the next sync injects a new one.

Zebra crash:

The marker is not installed into the FIB, only kept in zebra’s RIB for polling visibility. After a zebra crash, that RIB state is gone, so there is no stale marker left when zebra restarts.

vjardin:
LGTM

@maxime-leroy : it seems OK for me. I feel a bit lost by the over verbose commit logs. 

For sure, such feature turns the usage of FRR/zebra with Grout to a nice product ready system wich independent services that can be upgraded isolated. It should help during runtime and development too. It was a real missing piece.

Cosmetic: I'd prefer a shorter commit log that summarizes the commits's intends. Or if we want to keep such long commit logs which has some values, to have 2 sections into the commit log: a TL;DR summary at the beginning with the intends and then the descriptions you did.

christophefontaine (approved):

smoke/_init_frr.sh
```diff
@@ -179,9 +179,12 @@ set_srv6_route() {
...
 
-    # ----- build CLI & Grout forms ----------------------------------------
-    local seg_frr   ; IFS=/ ; seg_frr="${sids[*]}"       # SID/SID/…
-    local seg_space ; IFS=' ' ; seg_space="${sids[*]}"   # SID SID …
+    # ----- build CLI form -------------------------------------------------
+    # Run the join in a subshell so the IFS change does not leak to the
+    # rest of the function (where wait_event/mark_events read $tmp paths
+    # that would split on /).
+    local seg_frr
+    seg_frr=$(IFS=/; echo "${sids[*]}")   # SID/SID/...
```
rjarry: Remove all this code. And add a simple one liner below:

smoke/_init_frr.sh
rjarry:
```
${frr_ip} route ${prefix} ${nhop} segments ${sids// //}
```

frr/zebra_dplane_grout.c
```diff
@@ -18,14 +18,26 @@
...
 
 #define TOSTRING(x) #x
 
 static const char *gr_sock_path = GR_DEFAULT_SOCK_PATH;
 
+// Marker prefix and polling cadence. The marker is a dummy ::/128 SHARP
+// entry the plugin uses as a metaQ drain barrier: once observable in
+// the RIB, FIFO ordering of META_QUEUE_EARLY_ROUTE guarantees every
+// earlier ere has drained.
+#define GROUT_SYNC_MARKER_PFX "::/128"
```
rjarry: Why use a string representation of this? You could do:

```c
static const struct prefix grout_sync_marker_prefix = {
	.family = AF_INET6,
	.prefixlen = 128,
};
```

And avoid any parsing.

frr/zebra_dplane_grout.c
```diff
@@ -42,11 +54,23 @@ struct grout_ctx_t {
...
 };
 
 static struct grout_ctx_t grout_ctx = {0};
 static const char *plugin_name = "zebra_dplane_grout";
 
+// Parsed once from GROUT_SYNC_MARKER_PFX at zd_grout_plugin_init.
+// Reused by poll/cleanup/inject so str2prefix is only called (and
+// validated) in one place.
+static struct prefix grout_sync_marker_prefix;
+
```
rjarry: Moved above.

frr/zebra_dplane_grout.c
```diff
@@ -729,6 +924,10 @@ static int zd_grout_plugin_init(struct event_loop *) {
 
 	gr_log_debug("%s register status %d", plugin_name, ret);
 
+	// Parse the marker prefix once. Hardcoded constant: failure is a
+	// build-time mistake in GROUT_SYNC_MARKER_PFX.
+	assert(str2prefix(GROUT_SYNC_MARKER_PFX, &grout_sync_marker_prefix) == 1);
+
```
rjarry: Not needed, see above.

smoke/_init_frr.sh
rjarry: This could be changed to a more explicit variable name: `frr_log`

smoke/frr_restart_graceful_frr_test.sh
```diff
@@ -0,0 +1,133 @@
...
+#                     route survives.
+
+. $(dirname $0)/_init_frr.sh
+
+prefix_orphan=10.99.99.0/24
+prefix_kept=10.88.88.0/24
+nh_ip=172.16.0.2
+
+create_interface p0
+set_ip_address p0 172.16.0.1/24
```
rjarry, line 60: No need for `set_ip_address --persist p0 172.16.0.1/24` here?

smoke/_init_frr.sh
```diff
@@ -38,7 +38,37 @@ create_interface() {
...
+	local persist="$1"
+	local wait_pattern="$2"
+	local config="$3"
+
+	if [ "$persist" = 1 ]; then
+		cat >>"$builddir/frr_install/etc/frr/frr.conf" <<EOF
+$config
+EOF
+		return 0
+	fi
```
rjarry: Do not return early. And move the whole block *after* the `vtysh` + wait command has succeeded. That way, you don't need to call `set_ip_route` twice in tests. Just once with `--persist`.

smoke/grout_restart_resync_frr_test.sh
```diff
@@ -0,0 +1,89 @@
...
+# Respawn grout with the same command _init.sh used initially.
+$local_grout_cmd &
+grout_pid=$!
+
+SECONDS=0
+while ! socat FILE:/dev/null UNIX-CONNECT:$GROUT_SOCK_PATH 2>/dev/null; do
+	[ "$SECONDS" -gt 30 ] && fail "respawned grout did not open its socket"
+	kill -0 "$grout_pid" || fail "respawned grout died"
+	sleep 0.2
+done
```
rjarry: This would warrant an explicit function in _init.sh that takes care of the timeout and respawns the `grcli events` background command.

rjarry: This would warrant an explicit function in _init.sh that takes care of the timeout and respawns the `grcli events` background command.

frr/rt_grout.c
```diff
@@ -278,6 +278,7 @@ static int grout_gr_nexthop_to_frr_nexthop(
 
 		ctx.table = vrf_grout_to_frr(sr6->out_vrf_id);
 		nexthop_add_srv6_seg6local(nh, action, &ctx);
+		nh->type = nh->ifindex ? NEXTHOP_TYPE_IPV6_IFINDEX : NEXTHOP_TYPE_IPV6;
```
rjarry, line 282: Can you add one (or many) `Fixes:` tag for this?

Candidates: f481a7b6b4dbe db5852f3a7f8d 4eb73547b35f6

rjarry (changes requested):

rjarry (approved):

PR #607, opened 2026-04-27
icmp rate limiter

FRIDM636:
Hello,

I've been studying the grout codebase and I'm looking forward to contributing to it.
As a first step, I have implemented a basic ICMP rate limiter based on time-spacing.

Before extending it, I would like your feedback on:

- Use token bucket instead of or alongside time-spacing, to support burst tolerance
- Per-type granularity via type mask, allowing independent limits per ICMP message type
- Any specifics to ICMPv6 rate limiting support

Thanks.

## ICMP Rate Limiter Implementation

Implements a basic global ICMP rate limiter based on minimum time spacing between Echo Request responses.

### Core Implementation

**modules/ip/control/icmp_rl.h** and **modules/ip/control/icmp_rl.c**
- Introduces a time-spacing rate limiter with a static `time_space` structure tracking `last_ts` (microsecond timestamp) and `interval_ms` (millisecond interval).
- `icmp_rl_init(interval_ms)` initializes the limiter with caller-provided interval and captures current time via `gr_clock_us()`.
- `icmp_rl_allow()` compares elapsed microseconds against the configured interval; rejects requests occurring before the interval elapses (returns `false`), updates timestamp on acceptance (returns `true`).

### Control Plane Integration

**modules/ip/api/gr_ip4.h**
- Adds `GR_IP4_ICMP_RATE_LIMIT` request type with `struct gr_ip4_icmp_rl_req` containing a `uint32_t rate_limit` field.

**modules/ip/control/icmp.c**
- Registers `icmp_rate_limit()` API handler that receives the interval value and calls `icmp_rl_init()`.

### Data Path Integration

**modules/ip/datapath/icmp_input.c**
- Adds `RATE_LIMITED` edge to the ICMP input node state machine.
- Gates Echo Request handling with `icmp_rl_allow()` check.
- Routes rate-limited packets to dedicated `icmp_rate_limited` drop path instead of echo reply output.

### User Interface

**modules/infra/cli/icmp.c**
- Adds `icmp rate-limit (INTERVAL)` CLI command accepting interval in milliseconds (1 to UINT32_MAX).
- Parses argument and sends API request; returns error on parsing or API failure.

**modules/ip/control/meson.build**
- Adds `icmp_rl.c` to build source list.

### DIFF
--- a/modules/infra/cli/icmp.c
+++ b/modules/infra/cli/icmp.c
@@ -6,6 +6,7 @@
 #include "cli_l3.h"
 
 #include <gr_api.h>
+#include <gr_ip4.h>
 #include <gr_net_types.h>
 
 #include <ecoli.h>
@@ -37,6 +38,18 @@ static cmd_status_t ping(struct gr_api_client *c, const struct ec_pnode *p) {
 
 	errno = ENOPROTOOPT;
 	return CMD_ERROR;
+}
+
+static cmd_status_t icmp_rate_limit(struct gr_api_client *c, const struct ec_pnode *p) {
+	struct gr_ip4_icmp_rl_req req;
+
+	if (arg_u32(p, "INTERVAL", &req.rate_limit))
+		return CMD_ERROR;
+
+	if (gr_api_client_send_recv(c, GR_IP4_ICMP_RATE_LIMIT, sizeof(req), &req, NULL) < 0)
+		return CMD_ERROR;
+
+	return CMD_SUCCESS;
 }
 
 static cmd_status_t traceroute(struct gr_api_client *c, const struct ec_pnode *p) {
@@ -92,6 +105,19 @@ static int ctx_init(struct ec_node *root) {
 		with_help(
 			"Icmp ident field (default: random).",
 			ec_node_uint("IDENT", 1, UINT16_MAX, 10)
+		)
+	);
+	if (ret < 0)
+		return ret;
+
+	ret = CLI_COMMAND(
+		CLI_CONTEXT(root, CTX_ARG("icmp", "Icmp rate limit config")),
+		"rate-limit (INTERVAL)",
+		icmp_rate_limit,
+		"Set the icmp rate limiter",
+		with_help(
+			"The time space interval in milliseconds",
+			ec_node_uint("INTERVAL", 1, UINT32_MAX, 10)
 		)
 	);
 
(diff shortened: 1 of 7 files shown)


rjarry:
Hi @FRIDM636,

Thanks for the patch but this feature was already submitted in #576 

I forgot to link #14 to it

christophefontaine:
Hi, thanks a lot for your contribution !
It seems we have 2 PRs trying to achieve the same function, so how is this different from #576 ?

FRIDM636:
Hi @christophefontaine 
I think @rjarry has already done the right work. As I mentioned I added a simple rate limiting based on minimum time space between packets. I intended to go for the token bucket with a proportional refill rate instead of the full used refill.

I guess I should close the current PR.

**I would like to see if there is a specific subject I can work on.**

rjarry:
> **I would like to see if there is a specific subject I can work on.**

There are many things which would require some work. They are more or less fast/easy to implement. Some of them require extensive design work. Is there anything you need for a use case or are you doing this for exploration purposes?

Here are a few random ideas:

- Support for route metrics (multiple nexthops attached to the same route prefix but only one is installed in the FIB).
- Consistent hash algorithm for ECMP.
- Basic ACL/Firewall (will require some design work)
- Use BGP flowspec as control plane for ACL/Firewall
- Basic QoS (classify on RX, shaping on TX with hardware offload)

FRIDM636:
> There are many things which would require some work. They are more or less fast/easy to implement. Some of them require extensive design work. Is there anything you need for a use case or are you doing this for exploration purposes?

I have no specific use case, just looking to make open-source contributions. Grout caught my eye for two reasons: it aligns with my work (DPI, networking, DPDK integration...) and will definitely help harden my skills, and I genuinely appreciate your commitment to clean code and software architecture. Such high-quality work!

maxime-leroy:
@FRIDM636 it seems you have some experience on FRR, any help on this topic will be appreciate. 


You can join your channel on stack: https://dpdkproject.slack.com/archives/C07NAFWE1MG

FRIDM636:
@maxime-leroy It's been a while, I will do my best.

FRIDM636:
> Here are a few random ideas:
> 
>     Support for route metrics (multiple nexthops attached to the same route prefix but only one is installed in the FIB).
>     Consistent hash algorithm for ECMP.
>     Basic ACL/Firewall (will require some design work)
>     Use BGP flowspec as control plane for ACL/Firewall
>     Basic QoS (classify on RX, shaping on TX with hardware offload)

@rjarry interesting list, I would like to go with the **basic acl/firewall**, may you create an issue (feature) for that?
I will start digging on the subject.

rjarry:
@FRIDM636 here it is https://github.com/DPDK/grout/issues/609

PR #608, opened 2026-04-30
vxlan: add support for IPv6 underlay

rjarry:
Allow VXLAN tunnels to use IPv6 addresses for encapsulation. This extends the existing VXLAN implementation which only supported IPv4 local/remote VTEP addresses.

The smoke tests cover both basic VXLAN connectivity with IPv6 underlay and a full BGP EVPN scenario pairing FRR+grout against a standalone FRR+Linux peer with type-2/type-3 route exchange. Tests that require features not available in the installed FRR version are skipped gracefully.

## Overview
Adds IPv6-underlay VXLAN support by introducing an address-family-aware L3 address abstraction propagated through control-plane, CLI, FRR plugin, and datapath; implements IPv6-specific encapsulation/decapsulation and UDP checksum handling; increments GR API version to 2; and adds smoke tests for IPv6 underlay and EVPN over IPv6 with graceful skipping when the environment (FRR < 10.6.0) lacks required features.

## Core infrastructure
- Introduced struct l3_addr (addr_family_t + union for IPv4/IPv6) and static inline l3_addr_eq() which first matches af then compares IPv4 fields directly or memcmp's IPv6 bytes; for non-V4/V6 af the implementation falls through to true.
- Bumped GR_API_VERSION 1 → 2 to reflect non-backward-compatible changes to VXLAN-related public API types.
- Added FRR ↔ grout conversion helpers: ipaddr_to_l3_addr() and l3_addr_to_ipaddr() with fallbacks (GR_AF_UNSPEC / IPADDR_NONE) for unsupported cases.

## FRR plugin
- FRR-version gating: when CURRENT_FRR_VERSION < 10.6.0, code rejects/avoids IPv6 VTEP injection (logs error and aborts link/FDB operations); when >= 10.6.0, VTEPs are converted between ipaddr and l3_addr and handled normally.
- Replaced IPv4-specific population of vi.vtep_ip with l3_addr-based conversion on supported FRR versions; updated debug/log formatting to use l3_addr conversions.

## Datapath changes
- Encapsulation (vxlan_output_process):
  - Branch on d->vtep.af: IPv4 path uses fib4_lookup, IPv4 template, sets IPv4 lengths/checksums; IPv6 path uses fib6_lookup, IPv6 template, sets IPv6 payload length and computes UDP checksum in software (IPv6 requires non-zero UDP checksum).
  - Added NO_AF drop for unsupported families.
- Decapsulation (vxlan_input):
  - Determine source VTEP from mbuf packet type (IPv4 vs IPv6), populate struct l3_addr src_vtep, tag mbufs with L3 packet type, and route unsupported L3 types to a BAD_AF drop.
- VXLAN interface templates:
  - Replaced single IPv4 template with union of ipv4/ipv6 templates; control fills template fields (hop limit, proto, UDP dst port, VNI/flags) per local.af.
- Mbuf metadata and trace:
  - iface_mbuf_data.vtep and trace_vxlan_data.vtep changed from ip4_addr_t to struct l3_addr.

## Control-plane and CLI
- CLI:
  - Added arg_ip_guess(...) which tries IPv4 then IPv6 parsing and returns addr_family_t + raw address.
  - Replaced IPv4-only argument parsing/regexes (arg_ip4 / IPV4_RE / IP4_F) with address-family-agnostic parsing/formatting (arg_ip_guess / IP_ANY_RE / ADDR_F/ADDR_W).
  - FDB and flood CLI now condition visibility on fdb->vtep.af != GR_AF_UNSPEC.
- Control-plane:
  - fdb_learn signature changed to accept const struct l3_addr *vtep; MAC-moved detection uses l3_addr_eq() and updates use the dereferenced l3_addr.
  - Local nexthop selection moved from rib4_lookup → nexthop_lookup_l3 with NH_LOCAL_ADDR_FLAGS validation; missing nexthop returns EADDRNOTAVAIL.

## API and exported types
- Public API types updated from ip4_addr_t → struct l3_addr where VTEP can appear:
  - modules/l2/api/gr_l2.h: gr_iface_info_vxlan.local, gr_fdb_entry.vtep, gr_flood_vtep.addr.
  - modules/l2/control/l2.h: fdb_learn(...) parameter changed to const struct l3_addr *vtep; VXLAN iface template field converted to a union holding ipv4/ipv6 templates; flood_vteps storage type changed to struct l3_addr *.
  - Trace and other public-facing structs updated to use struct l3_addr.

## Tests and smoke harness
- Added tests:
  - smoke/vxlan6_test.sh: basic VXLAN tunnel and L3 connectivity with IPv6 underlay between GR and a Linux namespace.
  - smoke/evpn_vxlan6_frr_test.sh: full EVPN (type-2/type-3) over IPv6 underlay; script probes zebra version and exits 125 when FRR < 10.6.0.
- Makefile/smoke harness:
  - Smoke test selection no longer queries Meson options via jq; FRR-related smoke scripts are conditionally excluded based on the presence of built FRR dplane plugin.
  - Removed smoke_env wrapper machinery; each smoke script is invoked under sudo, exit code captured in rc; rc==125 is treated as SKIPPED (printed and coerced to rc=0), other non-zero rc printed as FAILED; recipe exits with the captured rc.

## Notable implementation details
- IPv6 path performs UDP checksum computation in software to satisfy RFC 2460 (IPv6 requires UDP checksum).
- Code explicitly avoids injecting IPv6 VTEP state into FRR versions that cannot represent it (FRR < 10.6.0) and logs/aborts those operations.


api/gr_net_types.h
```diff
@@ -172,3 +173,28 @@ static inline int ip6_net_parse(const char *s, struct ip6_net *net, bool zero_ma
...
+
+struct l3_addr {
+	addr_family_t af;
+	union {
+		struct {
+		} addr;
+		ip4_addr_t ipv4;
+		struct rte_ipv6_addr ipv6;
+	};
+};
```
christophefontaine, line 185: Is it time to simplify the CLI, so we don't need to make a difference between IP/IP6 commands ?
If yes, should we extend the concept, with an API agnostic of the ip version ?

christophefontaine (approved):

api/gr_net_types.h
rjarry: We already have generic CLI commands for route/address management.

```
address add 1.2.3.4/24 iface foo
address add fd22:666::4/48 iface bar
```

The dispatch is done by the CLI to send either `GR_IP4_ADDR_ADD` or `GR_IP6_ADDR_ADD` depending on the parsing result.

Are you referring to merging `GR_IP4_*` and `GR_IP6_*` APIs? We would still need to dispatch them internally in grout anyway since the routing tables or address lists are separate. I am not sure we can gain much from this.

PR #611, opened 2026-05-05
ctlplane: fix confusing vlan header in traces

rjarry:
When receiving an OSPF packet, the transmitted trace on the control plane interface sometimes shows a VLAN header which wasn't present on reception:

```
TRACE: [rx p0] 02:e3:ad:74:50:9b > 01:00:5e:00:00:05 / IP 172.16.0.2 > 224.0.0.5 ttl=1 proto=OSPF(89), (pkt_len=78)
TRACE: [cp tx p0] 52:54:00:00:00:01 > 01:00:5e:00:00:05 / VLAN id=4268 / IP 172.16.0.2 > 224.0.0.5 ttl=1 proto=OSPF(89), (pkt_len=78)
```

There is no VLAN in the packet and the packet lengths are equal. Nothing was inserted. The trace function just displays a garbage iface_mbuf_data->vlan_id value which wasn't reset.

Reset vlan_id before sending the packets to iface_cp_tx() like it is done in l4_loopback_output.

## Bug Fix

VLAN metadata was not being cleared before control-plane transmit, causing stale `vlan_id` values to appear in traces for packets that did not have VLAN headers.

## Changes

**modules/dhcp/datapath/dhcp_input.c**
- Clear `iface_mbuf_data(mbuf)->vlan_id` after prepending Ethernet and IP headers, before handing the packet to the control-plane transmit path.

**modules/l4/ospf_redirect.c**
- Clear `iface_mbuf_data(mbuf)->vlan_id` after writing the Ethernet header, before handing the packet to the control-plane transmit path.

Both changes ensure vlan_id is reset in code paths that prepend Ethernet headers, preventing stale VLAN metadata from being traced or sent via TAP frames.

### DIFF
--- a/modules/dhcp/datapath/dhcp_input.c
+++ b/modules/dhcp/datapath/dhcp_input.c
@@ -102,6 +102,10 @@ dhcp_input_process(struct rte_graph *graph, struct rte_node *node, void **objs, 
 			iface_get_eth_addr(iface, &eth->dst_addr);
 			eth->ether_type = RTE_BE16(RTE_ETHER_TYPE_IPV4);
 
+			// No VLAN header is written to the TAP. Clear the residual
+			// mbuf metadata so traces reflect the actual frame bytes.
+			iface_mbuf_data(mbuf)->vlan_id = 0;
+
 			control_output_set_cb(mbuf, iface_cp_tx, 0);
 		}
 
--- a/modules/l4/ospf_redirect.c
+++ b/modules/l4/ospf_redirect.c
@@ -6,6 +6,7 @@
 #include "ip4_datapath.h"
 #include "ip6_datapath.h"
 #include "port.h"
+#include "rxtx.h"
 
 #include <gr_infra.h>
 
@@ -110,6 +111,10 @@ static uint16_t ospf_redirect_process(
 		eth->ether_type = ether_type;
 		eth->src_addr = src;
 		eth->dst_addr = dst;
+
+		// No VLAN header is written to the TAP. Clear the residual
+		// mbuf metadata so traces reflect the actual frame bytes.
+		iface_mbuf_data(mbuf)->vlan_id = 0;
 next:
 		if (gr_mbuf_is_traced(mbuf)) {
 			gr_mbuf_trace_add(mbuf, node, 0);


david-marchand:
I am confused.
Is there no way to get a vlan packet sent to the loopback?

rjarry:
> I am confused. Is there no way to get a vlan packet sent to the loopback?

Not at the moment. And I don't think we will ever want to. The only purpose of these loopback interfaces is to allow daemons to connect to the outside world through grout. Receiving VLAN frames in linux would prevent this unless we add some sort of linux VLAN configuration to pop the tag.

modules/l4/l4_loopback_output.c
```diff
@@ -82,10 +82,6 @@ static uint16_t l4_loopback_output_process(
 			eth->src_addr = mac;
 			eth->ether_type = ether_type;
 
-			// No VLAN header is written to the TAP. Clear the residual
-			// mbuf metadata so traces reflect the actual frame bytes.
-			iface_mbuf_data(mbuf)->vlan_id = 0;
```
christophefontaine, line 87: I'd rather keep that `vlan_id` reset here, and reset the field in `dhcp_input_process` and `l2_redirect_process`.
We'd want to keep `struct iface_mbuf_data` with valid data, wouldn't we ?

rjarry: This isn't used by iface_cp_tx() except for tracing. So I figured we should do it in a single spot instead of duplicating code.

david-marchand: Hum, I agree with @christophefontaine, in general, we should leave only valid data in the metadata before passing a mbuf to a next node.

rjarry: fair enough

david-marchand:
There is still something off.

I would not touch anything in the l2_redirect node itself.

To me, any code prepending with a ethernet header should be clearing the vlan_id field.
So I think in your case, the bug is in

```
modules/l4/ospf_redirect.c-             edge = L2_REDIRECT;
modules/l4/ospf_redirect.c-
modules/l4/ospf_redirect.c:             eth = gr_mbuf_prepend(mbuf, eth);
modules/l4/ospf_redirect.c-             if (unlikely(eth == NULL)) {
modules/l4/ospf_redirect.c-                     edge = NO_HEADROOM;
modules/l4/ospf_redirect.c-                     goto next;
modules/l4/ospf_redirect.c-             }
modules/l4/ospf_redirect.c-
modules/l4/ospf_redirect.c-             eth->ether_type = ether_type;
modules/l4/ospf_redirect.c-             eth->src_addr = src;
modules/l4/ospf_redirect.c-             eth->dst_addr = dst;
modules/l4/ospf_redirect.c-next:
modules/l4/ospf_redirect.c-             if (gr_mbuf_is_traced(mbuf)) {
modules/l4/ospf_redirect.c-                     gr_mbuf_trace_add(mbuf, node, 0);
modules/l4/ospf_redirect.c-             }
modules/l4/ospf_redirect.c-             rte_node_enqueue_x1(graph, node, edge, mbuf);
```

david-marchand:
Looking at the rest of the tree, DHCP and OSPF seem to be the two locations needing some fix.

david-marchand:
Not sure it matters, but I think this commit
Fixes: 393b14bf6978 ("dhcp: forward to control plane interface when disabled")
Fixes: 5b1141226a55 ("infra: redirect ospf packets to the cp interface")

david-marchand (approved):

rjarry:
I'll add these for posterity. :+1:

PR #612, opened 2026-05-05
ci: allow merge on frr-master test failure

christophefontaine:
Switch the test with frr-master branch as an "non-voting" job.

Added `allow_fail: true` to the `gcc-14`/`frr: 'master'` matrix configuration and set `continue-on-error: ${{ matrix.conf.allow_fail || false }}` on the `build-and-tests` job. This allows the frr-master test case to fail without blocking PR merge, while other matrix configurations remain blocking.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -39,11 +39,13 @@ jobs:
             frr: '10.6'
           - compiler: gcc-14
             frr: 'master'
+            allow_fail: true
           - compiler: clang-16
             frr: '10.5'
           - compiler: clang-18
             frr: '10.5'
     runs-on: ubuntu-24.04${{ matrix.conf.arch == 'arm' && '-arm' || '' }}
+    continue-on-error: ${{ matrix.conf.allow_fail || false }}
     env:
       SANITIZE: ${{ case(matrix.conf.asan || false, 'address', 'none') }}
       BUILDTYPE: ${{ case(matrix.conf.debug || false, 'debug', 'debugoptimized') }}


.github/workflows/check.yml
```diff
@@ -36,10 +36,12 @@ jobs:
             frr: '10.6'
           - compiler: gcc-14
             frr: 'master'
+            allow_fail: true
           - compiler: clang-16
             frr: '10.5'
           - compiler: clang-18
             frr: '10.5'
+    continue-on-error: ${{ matrix.conf.allow_fail || false }}
```
rjarry, line 48: I don't think `continue-on-error` will affect the outcome when the job runs on the `main` branch. It just avoids the matrix to "fail early".

maxime-leroy: If the check is always green, nobody will look at the logs and an FRR master regression will go unnoticed for weeks -- so the test loses its signal value.

rjarry: We should not ignore the failures. But `frr-master` should *not* be run on the `main` branch. It should only run on pull requests.

christophefontaine: No need to rush on that, this PR was a test based on my understanding of the doc.

The `continue-on-error` should allow the merge, while keeping the check-mark in red if placed in the job definition.
If it is placed in the `step`, the mark should be green.

christophefontaine: If I look at 2 pipelines run:
- https://github.com/DPDK/grout/actions/runs/25366869330
- https://github.com/DPDK/grout/actions/runs/25366765893
The first one reports "success", while the other (from another pr, without the `continue-on-error` set) reports a failure.
In both pipelines, the `frr-master` job fails.

rjarry: My bad, I had misunderstood what this setting was for.

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idcontinue-on-error

> Prevents a workflow run from failing when a job fails. Set to `true` to allow a workflow run to pass when this job fails.

They even provide an example.

rjarry (approved):

PR #614, opened 2026-05-05
smoke: limit the duration of individual tests to 5 minutes

rjarry:
Kill the tests if they are not finished after 5 minutes. Also print the duration of tests when they finish.

Instead of storing test logs in temp files, put them into the build directory using unique names. In GitHub, upload all logs when the tests end (whether they failed or not) to enable post-mortem analysis.

### GNUmakefile
- Run each smoke test under a per-test timeout: tests are invoked via `timeout 5m` and executed under `sudo time` to capture elapsed time.
- Write deterministic artifacts into `$(BUILDDIR)/` using unique filenames: `$@.log` for test output and `$@.time` for elapsed time, replacing temporary files.
- Parse `time` output to record elapsed seconds and include elapsed time in all result messages.
- Treat exit codes explicitly: `124` reported as TIMEOUT (with elapsed time); other non-zero codes reported as FAILED (with elapsed time); zero reported as success (with elapsed time). Previous SKIPPED handling (`rc == 125`) is removed from the new flow.

### .github/workflows/check.yml
- Add an unconditional upload-artifact step (`if: always()`) to upload `build/smoke/` as an artifact named `smoke-logs-{matrix.arch}-{matrix.compiler}-dpdk-{matrix.dpdk_version}-frr-{matrix.frr_version}` with `retention-days: 10`. Step is positioned before existing `ccache -sv` and cache save/delete steps to ensure logs are archived regardless of test outcome.

### .github/workflows/codex_iliad.yml
- Remove the entire "PR Agent" workflow block (job, steps, environment and model/OpenAI endpoint configuration).

### .github/workflows/package.yml
- Modify rpm job to use an `arch` matrix (x86_64, aarch64) and dynamic `runs-on` to support multi-architecture builds.
- Scope ccache and artifact names to `matrix.arch` (replace prior static/mis-scoped values) and adjust container/build steps and artifact naming accordingly.

### main/id_pool_test.c
- Consolidate bench_summary logging into a single `print_message` call that emits both get/put statistics and the fail-rate in one line (removes separate `id_pool_put` log).
- Increase fail-rate threshold from 0.0001 to 0.001; keep the existing assertion on `bench->ret`.


maxime-leroy (approved):

maxime-leroy (approved):
acj

PR #615, opened 2026-05-11
cli: fix bash completion and vxlan display bugs

rjarry:
Fix use-after-free in bash completion mode where `bash_complete_node()` was taking ownership of the caller's `cmdlist` node, causing a double free when `main()` cleans up. Clone the node before passing it to `EC_NODE_SEQ`.

Also fix a truncation issue in bash completion where `gr_strcpy()` was cutting one character past the requested length, and fix the VXLAN local VTEP address display to handle IPv6 underlay.

## CLI bash completion fixes

Use-after-free in completion node handling:
- bash_complete_node() now calls ec_node_clone(cmdlist) and passes the clone into EC_NODE_SEQ/ec_node_sh_lex_expand; bash_complete() frees the cloned node with ec_node_free(cmdlist). This prevents ownership transfer of the caller's original cmdlist node and eliminates the double-free.

Buffer truncation in completion:
- bash_complete() copies exactly i bytes from COMP_LINE into buf using memcpy(buf, comp_line, i) and then sets buf[i] = '\0' instead of using gr_strcpy(buf, i, comp_line). This stops copying one character past the requested length and preserves correct next-word completion behavior.

Control-flow on error paths:
- Validation failures for COMP_LINE/COMP_POINT and COMP_POINT parsing now return immediately with failure (return ret) rather than falling through to the shared cleanup label; the shared cleanup remains responsible for freeing the cloned cmdlist and any allocated completion state.

## VXLAN display fix

IPv6-capable local VTEP address formatting:
- vxlan_show() and vxlan_list_info() switch from IPv4-specific formatting to address-family-aware formatting: they use ADDR_F with ADDR_W(vxlan->local.af) and pass &vxlan->local.addr (instead of IP4_F and vxlan->local). This allows correct rendering of the local VTEP address when the underlay is IPv6.

[Review Change Stack](https://app.coderabbit.ai/change-stack/DPDK/grout/pull/615?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cli/complete.c
+++ b/cli/complete.c
@@ -92,7 +92,8 @@ int bash_complete(struct ec_node *cmdlist) {
 		return ret;
 	}
 
-	gr_strcpy(buf, i, comp_line);
+	memcpy(buf, comp_line, i);
+	buf[i] = '\0';
 
 	if ((cmdlist = bash_complete_node(cmdlist)) == NULL) {
 		errorf("bash_complete_node: %s", strerror(errno));
--- a/cli/complete.c
+++ b/cli/complete.c
@@ -42,7 +42,7 @@ static struct ec_node *bash_complete_node(struct ec_node *cmdlist) {
 			EC_NO_ID,
 			ec_node("any", "prog_name"),
 			ec_node_option(EC_NO_ID, grcli_options_node()),
-			cmdlist
+			ec_node_clone(cmdlist)
 		)
 	);
 }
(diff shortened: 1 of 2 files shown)


cli/complete.c
```diff
@@ -92,7 +92,8 @@ int bash_complete(struct ec_node *cmdlist) {
 		goto end;
 	}
 
-	gr_strcpy(buf, i, comp_line);
```
maxime-leroy, line 95: gr_strcpy(buf, i+1, comp_line)  ?

maxime-leroy (approved):

PR #624, opened 2026-05-27
 Adaptive RSS auto-scaling 

maxime-leroy:
Why rx_usleep doesn't scale
  
  Grout's existing rx-side usleep(us) micro-sleep saves CPU at modest loads but breaks
  at line rate:

  - max_sleep_us is auto-clamped by ring depth / line rate, so at 25-100G the
  cap drops to a few microseconds → effectively zero. The worker busy-polls
  anyway, just with extra syscall overhead.
  - All queues remain RSS-fed during sleep, so a burst arriving on any queue can
  overflow its ring before the worker wakes.
  - The "idle CPU" gain only materializes at very low rates where usleep is
  long enough to matter, which is exactly the regime where the energy savings
  are negligible (the box was already idle).

  What rss-autoscale does

  Opt-in controller (--rss-autoscale=N, N = cluster grouping) that:
  
  - Narrows the HW RSS distribution per port to the smallest N that absorbs
  current load (via rte_eth_dev_rss_reta_update, with a portable emulation
  on DPAA2 that maps reta[i] = i % N to dpni_set_rx_hash_dist).
  - Parks unused workers via futex_wait → real 0% CPU (vs busy-poll noise).
  - Detection is edge-based in the datapath: each rx_burst that returns the
  HW burst-cap counts as "full"; 4 consecutive full → saturated. 1000
  consecutive empty (n_pkts == 0) → idle. Mid-load (1..cap-1) is steady,
  no transition. Counters are local, only edges trigger an atomic_store +
  control_queue push to wake the controller (~10 µs reaction).
  - Zero datapath cost when disabled (gr_config.rss_autoscale == 0):
  branch-predicted no-op at top of rxq_health_update.
  - Two policy knobs (grcli rss-autoscale set NAME cap|floor N) for thermal
  daemons and ops: cap = upper bound, floor = lower bound. Cap wins on
  conflict.
  

  Known limitations   

  - Elephant flow: a single 5-tuple flow hashes to one queue, so RSS
  scale-up cannot redistribute it. The controller will scale up uselessly
  until cap/max, leaving 11 unparked workers. 
  - Reordering on scale events: DPAA2 dispatches via hash % dist_size,
  so changing N reshuffles all flows. Scale events should be rare. No
  mitigation yet.
  - grcli bug: set NAME cap N and set NAME floor N patterns fail
  parsing in libecoli (to debug and fix)

## RSS Autoscale Controller

Implements an opt-in RSS autoscale controller (`--rss-autoscale=N`) that dynamically narrows per-port hardware RSS RETA to absorb current load and parks unused workers. Solves the limitation of rx-side usleep() which fails to scale at line-rate (max_sleep_us clamped by ring depth/line rate) and keeps all queues RSS-fed during idle sleep, risking ring overflow.

### Port Scaling Abstraction (`port_scale.{c,h}`)

New module for dynamic RETA reprogramming. Computes allowed queue counts by intersecting driver constraints with DPAA2's discrete set (via hardcoded lookup: 1,2,3,4,6,7,8,12,14,16,24,28,32,...) or falling back to [1..max]. Provides capability helpers (caps_get/free, caps_next/prev) and apply function that constructs uniform RETA (`reta[i] = i % n` across groups) and calls `rte_eth_dev_rss_reta_update()`. Query function reads back active queue count from current RETA or returns 0 if never programmed.

### RXQ Health Tracking (`rss_autoscale.{h,c}`)

Per-RXQ counters (consec_full, consec_empty) with atomic saturated/idle flags. Datapath `rxq_health_update()` hook (called per RX burst) increments counters: full bursts (HW burst-cap received) and empty bursts. State transitions trigger notifications: 4 consecutive full → saturated; 1000 consecutive empty → idle; transitions back to normal clear flags and wake controller. Notification pushes to control queue (~10 µs). Zero cost when disabled (rss_autoscale == 0).

### Worker Parking (`worker.{c,h}`)

Adds `atomic_int paused` field and futex-based park/unpark. `worker_park()` stores 1 and datapath futex_waits with 1s timeout at housekeeping. `worker_unpark()` stores 0 and issues FUTEX_WAKE. Datapath leaves RCU offline during wait to avoid blocking grace periods; wakes every second to check shutdown/reconfig events. Resets timing on unpark so park duration is not charged as busy cycles.

### Butterfly-Reverse RXQ Allocation (`worker.c`)

Changes distribution logic from simple forward walk to butterfly-reverse with separate global forward/reverse cursors. Even-indexed ports walk `cpus[]` forward from fwd_cursor; odd-indexed ports walk backward from rev_cursor. After each port, corresponding cursor advances by step derived from port's RXQ count. Improves endpoint placement: port 0 q0 → cpus[0], port 1 q0 → cpus[n-1]. On hyperthreaded systems (e.g., x86), endpoint workers land on different physical cores with exclusive sibling-thread and cache access. Enables multiple workers idle when dist_size=1 during low load.

### Control Loop (`rss_autoscale.c`)

Three-pass decision triggered on RXQ health transitions: (1) per-port load recommendation from atomic health flags, (2) global cluster-budget arbitration to constrain total cluster usage (filters allowed_n to keep multiples of cluster grouping size), (3) hardware apply. Exposes policy bounds: `rss_autoscale_set_cap()` / `rss_autoscale_clear_cap()` for upper bound (shedding) and `rss_autoscale_set_floor()` / `rss_autoscale_clear_floor()` for lower bound (force load). Cap overrides floor on conflict. Immediate apply on cap/floor changes with worker park/unpark sync. Subscribes to interface-add events to pin new ports to configured autoscale value.

### Datapath Integration

`port_rx.c` calls `rx_burst_done()` on every RX burst (replaces direct histogram increment) to update both histogram and health tracking. `main_loop.c` worker futex_wait loop integrated at housekeeping.

### CLI/API Infrastructure (`modules/infra/api/rss_autoscale.c`, `modules/infra/cli/rss_autoscale.c`)

New API handlers: `GR_RSS_AUTOSCALE_LIST` (streams per-port state: active queues, recommended load, cap/floor/min/max), `GR_RSS_AUTOSCALE_CAP_SET/CLEAR`, `GR_RSS_AUTOSCALE_FLOOR_SET/CLEAR`. CLI commands: `infra rss-autoscale show`, `set NAME cap N`, `set NAME floor N`, `clear NAME cap`, `clear NAME floor`.

### Configuration

New `--rss-autoscale=N` option in main.c: N=0 (disabled/legacy), N=1 (per-core), N=2 (grouping-by-2), N>2 (grouping-by-N). Stored in `gr_config.rss_autoscale`. Tick=20ms, up_hold=20ms, dn_hold=1000ms. Boots with N=1; widens as load appears.

[Review Change Stack](https://app.coderabbit.ai/change-stack/DPDK/grout/pull/624?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/main/config.h
+++ b/main/config.h
@@ -21,6 +21,20 @@ struct gr_config {
 	bool poll_mode;
 	bool log_syslog;
 	bool log_packets;
+	// Adaptive RSS auto-scaling: dynamically narrows the per-port RSS
+	// distribution + parks idle workers. The value is the cluster
+	// grouping size used to keep scaling steps aligned with cache-
+	// sharing groups so a step never splits one across active/parked:
+	//   0   = disabled (legacy: all workers always poll all configured
+	//         queues, no parking, no RETA scale).
+	//   1   = enabled, per-core scaling (no cluster constraint).
+	//   2   = enabled, scale by groups of 2. Typical use: x86 with
+	//         hyperthreading (sibling threads share L1/L2), or any SoC
+	//         where 2 cores share an L2 cache.
+	//   N>2 = enabled, scale by groups of N. For larger cache-sharing
+	//         topologies.
+	// Set via --rss-autoscale=N at startup.
+	uint16_t rss_autoscale;
 	vec char **eal_extra_args;
 	cpu_set_t control_cpus; // control plane threads allowed CPUs
 	cpu_set_t datapath_cpus; // datapath threads allowed CPUs
--- a/main/main.c
+++ b/main/main.c
@@ -60,6 +60,15 @@ static void usage(void) {
 	puts("                                 (default [::]:9111).");
 	puts("  -S, --syslog                   Redirect logs to syslog.");
 	puts("  -V, --version                  Print version and exit.");
+	puts("  -a, --rss-autoscale=N          Enable adaptive RSS auto-scaling and");
+	puts("                                 worker parking. Workers without traffic");
+	puts("                                 drop to 0% CPU. N = cluster grouping size:");
+	puts("                                   0 = disabled (default)");
+	puts("                                   1 = per-core scaling");
+	puts("                                   2 = scale by groups of 2 (x86 with");
+	puts("                                       hyperthreading, or SoCs sharing");
+	puts("                                       L2 between 2 cores).");
+	puts("                                   N = scale by groups of N.");
 	puts("  -h, --help                     Display this help message and exit.");
 	puts("  -m, --socket-mode PERMISSIONS  API socket file permissions (Default: 0660).");
 	puts("  -o, --socket-owner USER:GROUP  API socket file ownership");
@@ -176,8 +185,9 @@ static int parse_sock_owner(char *user_group_str) {
 static int parse_args(int argc, char **argv) {
 	int c;
 
-#define FLAGS ":M:Vhm:o:pSs:tu:vx"
+#define FLAGS ":M:Va:hm:o:pSs:tu:vx"
 	static struct option long_options[] = {
+		{"rss-autoscale", required_argument, NULL, 'a'},
 		{"help", no_argument, NULL, 'h'},
 		{"max-mtu", required_argument, NULL, 'u'},
 		{"metrics", required_argument, NULL, 'M'},
@@ -221,6 +231,13 @@ static int parse_args(int argc, char **argv) {
 			if (parse_sock_owner(optarg) < 0)
 				return errno_set(EINVAL);
 			break;
+		case 'a': {
+			unsigned int gm;
+			if (parse_uint(&gm, optarg, 10, 0, 16) < 0)
+				return perr("--rss-autoscale: %s", strerror(errno));
+			gr_config.rss_autoscale = (uint16_t)gm;
+			break;
+		}
 		case 'p':
 			gr_config.poll_mode = true;
 			break;
--- a/modules/infra/api/gr_infra.h
+++ b/modules/infra/api/gr_infra.h
@@ -102,6 +102,8 @@ struct gr_iface {
 #define GR_PORT_SET_N_TXQS GR_BIT64(33)
 #define GR_PORT_SET_Q_SIZE GR_BIT64(34)
 #define GR_PORT_SET_MAC GR_BIT64(35)
+#define GR_PORT_SET_RSS_CAP GR_BIT64(36)
+#define GR_PORT_SET_RSS_FLOOR GR_BIT64(37)
 
 // Base info structure for GR_IFACE_TYPE_PORT interfaces.
 struct __gr_iface_info_port_base {
@@ -110,6 +112,9 @@ struct __gr_iface_info_port_base {
 	uint16_t rxq_size;
 	uint16_t txq_size;
 	struct rte_ether_addr mac;
+	// rss-autoscale operator policy: max / min active RX queues (0 = unset).
+	uint16_t rss_cap;
+	uint16_t rss_floor;
 };
 
 // Complete port info structure including device arguments and driver name.
@@ -252,6 +257,7 @@ enum gr_infra_requests : uint32_t {
 	GR_IFACE_MAC_DEL,
 	GR_IFACE_MAC_LIST,
 	GR_IFACE_MAC_SET,
+	GR_RSS_AUTOSCALE_LIST,
 };
 
 enum gr_infra_events : uint32_t {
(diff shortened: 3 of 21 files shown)


MortenBroerup:
The power management algorithm per lcore in the SmartShare appliance is relatively simple:
After a certain duration of idle or low workloads, the system uses "napping" (our internal term for that power state) with relatively short durations.
And after a certain duration of idle workloads, the system uses "sleeping", which means longer durations.

As input for the algorithm for each lcore, we look at the maximum number of objects processed by any "node" every time the "graph" is traversed.
0 objects is Idle, 1 object is Low, more than 1 object is Normal.
If any "node" reports that it has more objects ready for processing, we don't nap or sleep at all, but restart traversing the graph immediately.

We are primarily targeting low power consumption during low traffic hours, and low scheduling latency during peak traffic hours.

For long durations. we use nanosleep(), where the kernel puts the lcore in a low power state.
For short durations, we use rte_power_pause(), which provides modest power savings with nearly no jitter or wake-up delay.

We don't use any of the fancy power management hooks in DPDK. It's all done by our application.

maxime-leroy:
> The power management algorithm per lcore in the SmartShare appliance is relatively simple: After a certain duration of idle or low workloads, the system uses "napping" (our internal term for that power state) with relatively short durations. And after a certain duration of idle workloads, the system uses "sleeping", which means longer durations.
> 
> As input for the algorithm for each lcore, we look at the maximum number of objects processed by any "node" every time the "graph" is traversed. 0 objects is Idle, 1 object is Low, more than 1 object is Normal. If any "node" reports that it has more objects ready for processing, we don't nap or sleep at all, but restart traversing the graph immediately.
> 
> We are primarily targeting low power consumption during low traffic hours, and low scheduling latency during peak traffic hours.
> 
> For long durations. we use nanosleep(), where the kernel puts the lcore in a low power state. For short durations, we use rte_power_pause(), which provides modest power savings with nearly no jitter or wake-up delay.
> 
> We don't use any of the fancy power management hooks in DPDK. It's all done by our application.

grout already does the equivalent of your long-idle tier: the rx loop ramps a per-worker usleep() up to a max_sleep_us clamped by the ring drain time. Your idle signal is better than ours though (max objects per node per graph walk vs our binary "any packets last window"), and I may adopt that.

 But a per-lcore sleep solves a different problem than this PR. It lowers a core's duty cycle while keeping every worker RSS-fed: it can't reduce the number of busy cores
  or move a flow off one to free it. rss-autoscale narrows the HW RSS distribution to concentrate traffic on fewer queues, then parks the drained workers, which is what
  reaches a real 0% / deep cpuidle state for a sustained period.
  
 On DPAA2 the short-idle tier also has no backing: rte_power_pause() is TPAUSE/WAITPKG (x86 only, no-op on Cortex-A72), and a ~3-4 us usleep is a non-starter at line rate (a 512-entry ring at 100G/64B fills in ~3.4 us). That clamp is what motivated this PR.

MortenBroerup:
Any existing or new considerations regarding conntrack?
All packets related to a connection should be handled by the same lcore.
I guess the forward direction is handled by the same core, and the reverse direction might be handled by a different lcore.
Preferably, the same lcore should handle both forward and reverse direction packets, e.g. for keeping track of TCP sequence/acknowledgement numbers. If the forward and reverse flows of a connection are handled by separate lcores, every write to the conntrack entry will cause an expensive cache miss, as the ownership of the cache line holding the conntrack entry will be alternating between those two lcores.

rjarry:
@MortenBroerup I am not sure how to guarantee that reverse flows are handled by the same worker. For this we would need to switch to a pipeline graph model where the dispatching is done in software.

MortenBroerup:
> As input for the algorithm for each lcore, we look at the maximum number of objects processed by any "node" every time the "graph" is traversed. 0 objects is Idle, 1 object is Low, more than 1 object is Normal. If any "node" reports that it has more objects ready for processing, we don't nap or sleep at all, but restart traversing the graph immediately.

This might need some more elaboration.
We don't use the rte_graph library; we use something similar, but home grown.
Unlike the rte_graph library, our "nodes" do not process all their pending objects. They only process up to a certain threshold. If more are pending after processing, the workload signal from the "node" is More.
We introduced the Low workload signal to let the power manager remain in low-power state, and not be woken up by some keep-alive packet or other low bandwidth traffic.

MortenBroerup:
> @MortenBroerup I am not sure how to guarantee that reverse flows are handled by the same worker. For this we would need to switch to a pipeline graph model where the dispatching is done in software.

I'm not yet familiar with Grout and the rte_graph library, so maybe it's impossible.
There are some tricks involving Symmetric RSS.
And for Dynamic NAT, there's a trick about cleverly selecting the NAT'ed TCP/UDP source port so the reverse tuple hashes to the same as the forward tuple, (Instead of selecting a random TCP/UDP source port when NAT'ing.)

PR #628, opened 2026-06-05
frr: signal FINISHED_READING after sync on FRR >= 10.7

rjarry:
FRR 10.7 introduced a new ZEBRA_DPLANE_FINISHED_READING startup stage. zebra_main_router_started() which calls zserv_start() to accept zclient connections from routing daemons is now deferred until that signal arrives. The kernel dplane sends it after reading interfaces, addresses and routes from netlink. The grout dplane never sends it, so zebra never starts listening and no routing daemon can connect. This breaks OSPF, OSPF6, BGP and all protocols that rely on zclient connections.

Send ZEBRA_DPLANE_FINISHED_READING after the grout sync chain completes and all state has been injected into zebra. When there is nothing to sync at initial boot, send it from the grout_sync_fdb early exit path. In the normal path, send it from grout_sync_arm_sweep before cancelling the pre-arm placeholder so that zebra_main_router_started's own sweep arm is a no-op.

The smoke kill_frr_daemons helper is also reworked to use inotifywait for more reliable respawn detection.

## frr/zebra_dplane_grout.c

- Route end-of-initial-sync completion through a new grout_main_router_started() callback that:
  - cleans up the injected marker,
  - re-stamps zrouter.startup_time,
  - cancels the pre-armed placeholder timer via event_cancel(&zrouter.t_rib_sweep),
  - then calls zebra_main_router_started() (so zebra performs its normal sweep arm + zserv start with the placeholder removed).
- grout_sync_fdb(): when no VRFs are marked for initial dataplane sync, call grout_main_router_started() immediately on the early-exit path.
- Marker-driven completion path: set the marker callback to grout_main_router_started() so the marker-observation flow performs marker cleanup, restamps startup_time, drops the pre-arm placeholder, and invokes zebra_main_router_started().

## frr/if_grout.c

- grout_link_change(): for CURRENT_FRR_VERSION >= 10.7.0, set the dplane context interface name to VRF_DEFAULT_NAME when the iface is the default VRF (zif_type == ZEBRA_IF_VRF && gr_if->id == GR_VRF_DEFAULT_ID); otherwise use gr_if->name. Preserve legacy behavior on older FRR versions.

## smoke/_init_frr.sh

- kill_frr_daemons(): replace pidfile-polling loop with an inotifywait-driven workflow that:
  - snapshots per-daemon pidfile contents, removes the pidfiles and builds pkill --pidfile arguments,
  - force-kills the recorded PIDs,
  - waits up to 10s for inotifywait modify events under the FRR run directory and removes observed entries from the snapshot to detect respawn via pidfile updates,
  - verifies respawn/termination via pkill -0 and fails if targets remain.

## .github/workflows/check.yml and README.md

- Add inotify-tools to CI install step and to README development dependency lists to support the new inotifywait-based smoke-test logic.

### DIFF
--- a/frr/if_grout.c
+++ b/frr/if_grout.c
@@ -127,10 +127,14 @@ void grout_link_change(struct gr_iface *gr_if, bool new, bool startup) {
 	dplane_ctx_set_ifp_link_nsid(ctx, GROUT_NS);
 	dplane_ctx_set_ifp_zif_type(ctx, zif_type);
 	dplane_ctx_set_ifindex(ctx, ifindex_grout_to_frr(gr_if->id));
-	dplane_ctx_set_ifname(ctx, gr_if->name);
 #if CURRENT_FRR_VERSION >= MAKE_FRRVERSION(10, 7, 0)
+	if (zif_type == ZEBRA_IF_VRF && gr_if->id == GR_VRF_DEFAULT_ID)
+		dplane_ctx_set_ifname(ctx, VRF_DEFAULT_NAME);
+	else
+		dplane_ctx_set_ifname(ctx, gr_if->name);
 	dplane_ctx_set_startup(ctx, startup);
 #else
+	dplane_ctx_set_ifname(ctx, gr_if->name);
 	dplane_ctx_set_ifp_startup(ctx, startup);
 #endif
 	dplane_ctx_set_ifp_family(ctx, AF_UNSPEC);
--- a/smoke/_init_frr.sh
+++ b/smoke/_init_frr.sh
@@ -253,32 +253,24 @@ set_srv6_route() {
 # Fails after 10s if any daemon is not respawned with a different pid.
 kill_frr_daemons() {
 	local frr_run="$builddir/frr_install/var/run/frr"
-	local daemon new
-	declare -A old_pids
+	local daemon f
+	declare -A pids
 
 	for daemon in "$@"; do
-		old_pids[$daemon]=$(cat "$frr_run/$daemon.pid")
+		f="$frr_run/$daemon.pid"
+		pids[$f]=$(cat "$f")
 	done
 
-	for daemon in "$@"; do
-		kill -9 "${old_pids[$daemon]}"
-	done
+	{ sleep 0.2 && kill -9 "${pids[@]}"; } &
 
-	SECONDS=0
-	while :; do
-		local all_respawned=1
-		for daemon in "$@"; do
-			new=$(cat "$frr_run/$daemon.pid" 2>/dev/null || true)
-			if [ -z "$new" ] || [ "$new" = "${old_pids[$daemon]}" ]; then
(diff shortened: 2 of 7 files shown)


frr/if_grout.c
```diff
@@ -127,10 +127,14 @@ void grout_link_change(struct gr_iface *gr_if, bool new, bool startup) {
 	dplane_ctx_set_ifp_link_nsid(ctx, GROUT_NS);
 	dplane_ctx_set_ifp_zif_type(ctx, zif_type);
 	dplane_ctx_set_ifindex(ctx, ifindex_grout_to_frr(gr_if->id));
-	dplane_ctx_set_ifname(ctx, gr_if->name);
 #if CURRENT_FRR_VERSION >= MAKE_FRRVERSION(10, 7, 0)
+	if (zif_type == ZEBRA_IF_VRF && gr_if->id == GR_VRF_DEFAULT_ID)
+		dplane_ctx_set_ifname(ctx, VRF_DEFAULT_NAME);
+	else
+		dplane_ctx_set_ifname(ctx, gr_if->name);
```
maxime-leroy, line 134: This commit looks good for me.

smoke/_init_frr.sh
```diff
@@ -253,32 +253,26 @@ set_srv6_route() {
...
-				all_respawned=0
-				break
-			fi
-		done
-		[ "$all_respawned" -eq 1 ] && break
-		[ "$SECONDS" -ge 10 ] && \
-			fail "watchfrr did not respawn $* within 10s"
-		sleep 0.1
-	done
+	timeout 10s inotifywait -mqr -e modify --format %f "$frr_run" |
```
maxime-leroy: What the goal of this commit ?

rjarry: the active polling clobbers the test log and makes it harder to read and find errors.

subprojects/packagefiles/frr/zebra-fix-startup-regression-for-external-dplane-plu.patch
```diff
@@ -0,0 +1,96 @@
...
++	 * listener lets mgmtd push config to routing daemons before they
++	 * connect to zebra and learn about VRFs/interfaces.
++	 * The call is idempotent (started_p guard in zserv_start).
++	 */
++	zserv_start(NULL);
++
+ 	return 0;
+ }
+ 
+--
```
maxime-leroy: This patch will never be merged in upstream, so NACK for this commit. We need to find a fix without patching FRR.

rjarry: What if the bug is in FRR? :D I searched for 3 days straight and didn't find a way to fix this in the grout dplane plugin alone.

maxime-leroy (approved):

Issue #3, opened 2024-03-21
write architecture document


rjarry:

MortenBroerup:
As a new Grout contributor, the lack of developer documentation makes the learning curve quite steep.

What is missing is mainly a description of the high level architecture and design (interaction between control plane and worker threads, and means for thread safety), and a description of how the source code is organized.
(It's easy to understand what the various functions do, just by reading the source code.)

E.g. use the L2 module as an example.
Architecture:
When adding a static entry to the FDB in the CLI, how does it make its way into the FDB, i.e. which messages/structures are passed around, and which functions do they traverse, and how is thread safety ensured?
Source code:
There are 4 subdirectories under modules/l2/.
Code under "datapath" is exclusively used for the Grout data plane Worker threads. DPDK can be used here.
Code under "cli" is exclusively used for building the Grout CLI application. DPDK cannot be used here.
Code under "api" is exclusively data structure definitions, shared by both the Grout application and external applications (such as the Grout CLI application)? DPDK cannot be used here.
What is under "control"? It seems DPDK can be used here. How to determine which functions are for Worker threads only, for control plane (Main thread) only, or both?

Scarce information is better than no information.
Format is not important. It could be a presentation or a YouTube video.

PR #629, opened 2026-06-09
Adaptive interrupt

maxime-leroy:
Opt-in --napi mode: an idle worker stops busy-polling and blocks on its rx queue interrupts via rte_eth_dev_rx_intr_* / rte_epoll_wait, resuming polling when a packet wakes it. A second commit pins the worker's uclamp_min to max so schedutil keeps the core at full clock while runnable, dropping only when it actually sleeps.

### Configuration / CLI
- Added `-n/--napi` (`gr_config.napi` in `main/config.h`).
- When `--napi` is enabled, option parsing forces `gr_config.poll_mode = true` (so idle uses the NAPI/interrupt-blocking path).
- Documented `--napi` in `main/main.c` help text.

### Port setup: enable RX queue interrupts only in NAPI mode
- In `port_configure()`, RX queue interrupt configuration (`conf.intr_conf.rxq`) is enabled only when `gr_config.napi` is set; link-status interrupt logic is unchanged.
- After `rte_eth_dev_start()` succeeds (in both `port_mtu_set()` and `iface_port_reconfig()` re-start paths), `worker_rearm_all()` is called so dataplane workers re-arm against the started ports.

### Worker wakeups while blocked in `rte_epoll_wait()`
- `struct worker` gains an `eventfd` (`wakeup_fd`) used to wake dataplane workers that are blocked in `rte_epoll_wait()`.
- `worker_create()` creates `wakeup_fd` (`EFD_NONBLOCK | EFD_CLOEXEC`); both failure cleanup and `worker_destroy()` close it.
- Wakeups are routed through the updated wakeup helper and write to `wakeup_fd` so epoll-blocked workers resume:
  - `worker_wakeup()` writes `1`
  - `worker_rearm_all()` issues a special “re-arm” kick (encoded via the new `WORKER_KICK_REARM` constant in `worker.h`)

### Datapath idle behavior (`modules/infra/datapath/main_loop.c`)
- Added an opt-in NAPI-style idle path gated by `gr_config.napi`:
  - Tracks consecutive “empty” housekeeping intervals; after `NAPI_EMPTY_WINDOWS` it enters `napi_wait()`.
  - `napi_wait()`:
    - Arms RX queue interrupts for started/owned RX queues via `rte_eth_dev_rx_intr_enable()`
    - Registers corresponding RX-queue event sources on the lcore’s per-thread epoll instance
    - Takes QSBR thread offline while blocking in `rte_epoll_wait()` (uses a short “settle” timeout first, then waits indefinitely)
    - Drains `wakeup_fd` on return (so control-plane/rearm kicks break/adjust the wait)
    - Disarms the RX interrupts it armed
- In NAPI mode, blocked time inside the idle wait is accounted by folding the block interval into the subsequent cycle delta as `sleep_cycles` / `n_sleeps`; otherwise the existing `usleep()`-based accounting remains.
- Adds per-thread NAPI state used to avoid redundant epoll registration and to clean up cached registrations on reconfiguration/shutdown.

### CPU utilization floor while runnable
- Adds `worker_perf_floor()` (called from the main loop when `gr_config.napi` is set) to raise `uclamp_min` via `sched_setattr` when available, so `schedutil` holds higher frequency while the worker is runnable.
- `meson.build` detects build-time availability of `sched_setattr` and defines `HAVE_SCHED_SETATTR`.

### Control-plane wake behavior
- `control_input.c` now includes `worker.h` and uses an atomic `gr_worker_active` check after queueing control messages:
  - if there are no active workers, it calls `worker_wakeup_any()` to ensure the pending ring element is processed.
- Introduces `control_input_pending()` helper and declares it in `control_input.h`.

### Smoke test enablement
- `smoke/_init.sh` appends `-n` to `grout_extra_options` when `${napi:-true}` is true (default enabled).

### DPDK tap PMD support needed for RX interrupt blocking
- Updated `dpdk-25.11.wrap` to include two new tap-net patches:
  - `net-tap-support-Rx-queue-interrupt-enable-disable.patch`: adds missing Rx interrupt enable/disable ops so the generic RX interrupt API can succeed (no-op implementations returning `0`).
  - `net-tap-drain-queue-fd-in-Rx-interrupt-mode.patch`: records/locks RX interrupt mode (`intr_mode` / `intr_mode_set`), adjusts `tun_alloc` and `pmd_rx_burst` early-exit behavior so epoll wakeups don’t incorrectly stop bursts while the fd remains readable.

### DIFF
--- a/docs/grout.8.scdoc
+++ b/docs/grout.8.scdoc
@@ -51,6 +51,14 @@ Grout is a software router based on DPDK _rte_graph_.
 *-p*, *--poll-mode*
 	Disable automatic micro-sleep.
 
+*-a*, *--adaptive-irq*
+	Make idle workers block on their RX queue interrupts instead of
+	busy-polling, so the CPU can idle during quiet periods. An incoming
+	packet raises the interrupt and the worker resumes polling. It
+	requires PMD support for RX queue interrupts; without it a worker
+	falls back to the micro-sleep ramp, unless *--poll-mode* is also
+	given.
+
 *-S*, *--syslog*
 	Redirect logs to syslog.
 
--- a/main/config.h
+++ b/main/config.h
@@ -19,6 +19,7 @@ struct gr_config {
 	unsigned max_mtu;
 	bool test_mode;
 	bool poll_mode;
+	bool adaptive_irq; // --adaptive-irq: adaptive interrupt rx
 	bool log_syslog;
 	bool log_packets;
 	bool override_default_route;
--- a/main/main.c
+++ b/main/main.c
@@ -61,6 +61,7 @@ static void usage(void) {
 	puts("                                 (default [::]:9111).");
 	puts("  -S, --syslog                   Redirect logs to syslog.");
 	puts("  -V, --version                  Print version and exit.");
+	puts("  -a, --adaptive-irq             Idle workers block on rxq interrupts.");
 	puts("  -h, --help                     Display this help message and exit.");
 	puts("  -m, --socket-mode PERMISSIONS  API socket file permissions (Default: 0660).");
 	puts("  -o, --socket-owner USER:GROUP  API socket file ownership");
@@ -185,8 +186,9 @@ static bool parse_bool_env(const char *name) {
 static int parse_args(int argc, char **argv) {
 	int c;
 
-#define FLAGS ":M:Vhm:o:pSs:tu:vx"
+#define FLAGS ":aM:Vhm:o:pSs:tu:vx"
 	static struct option long_options[] = {
+		{"adaptive-irq", no_argument, NULL, 'a'},
 		{"help", no_argument, NULL, 'h'},
 		{"max-mtu", required_argument, NULL, 'u'},
 		{"metrics", required_argument, NULL, 'M'},
@@ -218,6 +220,9 @@ static int parse_args(int argc, char **argv) {
 
 	while ((c = getopt_long(argc, argv, FLAGS, long_options, NULL)) != -1) {
 		switch (c) {
+		case 'a':
+			gr_config.adaptive_irq = true;
+			break;
 		case 'h':
 			usage();
 			exit(EXIT_SUCCESS);
--- a/modules/infra/control/port.c
+++ b/modules/infra/control/port.c
@@ -272,7 +272,7 @@ static int port_mtu_set(struct iface *iface, uint16_t mtu) {
 	if ((ret = rte_eth_dev_start(p->port_id)) < 0)
 		return errno_log(-ret, "rte_eth_dev_start");
 	p->started = true;
-	worker_wakeup_all(); // let workers (re)arm now that the port is started
+	worker_rearm_all(); // let workers (re)arm now that the port is started
 	iface->mtu = mtu;
 
 	vec_foreach (struct iface *s, iface->subinterfaces)
@@ -365,7 +365,7 @@ static int iface_port_reconfig(
 		return errno_log(-ret, "rte_eth_dev_start");
 
 	p->started = true;
-	worker_wakeup_all(); // let workers (re)arm now that the port is started
+	worker_rearm_all(); // let workers (re)arm now that the port is started
 
 	return 0;
 }
--- a/modules/infra/control/port.c
+++ b/modules/infra/control/port.c
@@ -152,6 +152,9 @@ int port_configure(struct iface_info_port *p, uint16_t n_txq_min) {
 	if (info.dev_flags != NULL && *info.dev_flags & RTE_ETH_DEV_INTR_LSC) {
 		conf.intr_conf.lsc = 1;
 	}
+	// Request RX queue interrupts so idle workers can block on them.
+	if (gr_config.adaptive_irq)
+		conf.intr_conf.rxq = 1;
 
 	if ((ret = rte_eth_dev_configure(p->port_id, p->n_rxq, p->n_txq, &conf)) < 0)
 		return errno_log(-ret, "rte_eth_dev_configure");
@@ -269,6 +272,7 @@ static int port_mtu_set(struct iface *iface, uint16_t mtu) {
 	if ((ret = rte_eth_dev_start(p->port_id)) < 0)
 		return errno_log(-ret, "rte_eth_dev_start");
 	p->started = true;
+	worker_wakeup_all(); // let workers (re)arm now that the port is started
 	iface->mtu = mtu;
 
 	vec_foreach (struct iface *s, iface->subinterfaces)
@@ -361,6 +365,7 @@ static int iface_port_reconfig(
 		return errno_log(-ret, "rte_eth_dev_start");
 
 	p->started = true;
+	worker_wakeup_all(); // let workers (re)arm now that the port is started
 
 	return 0;
 }
--- a/modules/infra/control/worker.c
+++ b/modules/infra/control/worker.c
@@ -32,6 +32,8 @@ LOG_TYPE("worker");
 
 struct workers workers = STAILQ_HEAD_INITIALIZER(workers);
 
+atomic_uint gr_worker_active;
+
 int worker_create(unsigned cpu_id) {
 	bool attr_inited = false, thread_created = false, inserted = false;
 	struct worker *worker = rte_zmalloc(__func__, sizeof(*worker), 0);
@@ -132,9 +134,7 @@ void worker_wait_wakeup(struct worker *w) {
 	pthread_mutex_unlock(&w->wakeup.lock);
 }
 
-void worker_wakeup(struct worker *w) {
-	uint64_t one = 1;
-
+static void worker_kick(struct worker *w, uint64_t val) {
 	pthread_mutex_lock(&w->wakeup.lock);
 	w->wakeup.set = true;
 	pthread_cond_signal(&w->wakeup.cond);
@@ -144,15 +144,41 @@ void worker_wakeup(struct worker *w) {
 	// worker idle on rte_epoll_wait is woken through the eventfd instead.
 	// EAGAIN means a kick is already pending and undrained, which is enough.
 	if (gr_config.adaptive_irq
-	    && write(w->adaptive_irq.wakeup_fd, &one, sizeof(one)) != sizeof(one)
+	    && write(w->adaptive_irq.wakeup_fd, &val, sizeof(val)) != sizeof(val)
 	    && errno != EAGAIN)
 		LOG(ERR, "worker %u wakeup_fd write: %s", w->cpu_id, strerror(errno));
+}
+
+void worker_wakeup(struct worker *w) {
+	worker_kick(w, 1);
 }
 
 void worker_wakeup_all(void) {
 	struct worker *worker;
 	STAILQ_FOREACH (worker, &workers, next)
 		worker_wakeup(worker);
+}
+
+// Wake one rxq-owning worker to drain the control input ring, if none is active.
+void worker_wakeup_any(void) {
+	struct worker *worker;
+
+	if (worker_active_count() != 0)
+		return;
+
+	STAILQ_FOREACH (worker, &workers, next) {
+		if (vec_len(worker->rxqs) > 0) {
+			worker_wakeup(worker);
+			return;
+		}
+	}
+}
+
+// Re-arm every worker's rxq interrupts after a port stop/start dropped them.
+void worker_rearm_all(void) {
+	struct worker *worker;
+	STAILQ_FOREACH (worker, &workers, next)
+		worker_kick(worker, WORKER_KICK_REARM);
 }
 
 unsigned worker_count(void) {
--- a/modules/infra/control/worker.c
+++ b/modules/infra/control/worker.c
@@ -24,6 +24,7 @@
 #include <pthread.h>
 #include <sched.h>
 #include <stdatomic.h>
+#include <sys/eventfd.h>
 #include <sys/queue.h>
 #include <unistd.h>
 
@@ -43,8 +44,15 @@ int worker_create(unsigned cpu_id) {
 
 	worker->cpu_id = cpu_id;
 	worker->lcore_id = LCORE_ID_ANY;
+	worker->adaptive_irq.wakeup_fd = -1;
 	pthread_mutex_init(&worker->wakeup.lock, NULL);
 	pthread_cond_init(&worker->wakeup.cond, NULL);
+
+	worker->adaptive_irq.wakeup_fd = eventfd(0, EFD_NONBLOCK | EFD_CLOEXEC);
+	if (worker->adaptive_irq.wakeup_fd < 0) {
+		ret = errno;
+		goto end;
+	}
 
 	CPU_ZERO(&cpuset);
 	CPU_SET(cpu_id, &cpuset);
@@ -83,6 +91,8 @@ end:
 				pthread_cancel(worker->thread);
 			pthread_cond_destroy(&worker->wakeup.cond);
 			pthread_mutex_destroy(&worker->wakeup.lock);
+			if (worker->adaptive_irq.wakeup_fd >= 0)
+				close(worker->adaptive_irq.wakeup_fd);
 			rte_free(worker);
 		}
 		LOG(ERR, "worker %u start failed: %s", cpu_id, strerror(ret));
@@ -104,6 +114,7 @@ int worker_destroy(unsigned cpu_id) {
 	pthread_join(worker->thread, NULL);
 	pthread_cond_destroy(&worker->wakeup.cond);
 	pthread_mutex_destroy(&worker->wakeup.lock);
+	close(worker->adaptive_irq.wakeup_fd);
 	worker_graph_free(worker);
 	vec_free(worker->rxqs);
 	vec_free(worker->txqs);
@@ -122,10 +133,26 @@ void worker_wait_wakeup(struct worker *w) {
 }
 
 void worker_wakeup(struct worker *w) {
+	uint64_t one = 1;
+
 	pthread_mutex_lock(&w->wakeup.lock);
 	w->wakeup.set = true;
 	pthread_cond_signal(&w->wakeup.cond);
 	pthread_mutex_unlock(&w->wakeup.lock);
+
+	// The condvar above only reaches a worker parked on graph == NULL. An adaptive-irq
+	// worker idle on rte_epoll_wait is woken through the eventfd instead.
+	// EAGAIN means a kick is already pending and undrained, which is enough.
+	if (gr_config.adaptive_irq
+	    && write(w->adaptive_irq.wakeup_fd, &one, sizeof(one)) != sizeof(one)
+	    && errno != EAGAIN)
+		LOG(ERR, "worker %u wakeup_fd write: %s", w->cpu_id, strerror(errno));
+}
+
+void worker_wakeup_all(void) {
+	struct worker *worker;
+	STAILQ_FOREACH (worker, &workers, next)
+		worker_wakeup(worker);
 }
 
 unsigned worker_count(void) {
--- a/modules/infra/control/worker.c
+++ b/modules/infra/control/worker.c
@@ -32,6 +32,7 @@ LOG_TYPE("worker");
 struct workers workers = STAILQ_HEAD_INITIALIZER(workers);
 
 int worker_create(unsigned cpu_id) {
+	bool attr_inited = false, thread_created = false, inserted = false;
 	struct worker *worker = rte_zmalloc(__func__, sizeof(*worker), 0);
 	pthread_attr_t attr;
 	cpu_set_t cpuset;
@@ -48,13 +49,16 @@ int worker_create(unsigned cpu_id) {
 	CPU_ZERO(&cpuset);
 	CPU_SET(cpu_id, &cpuset);
 	pthread_attr_init(&attr);
+	attr_inited = true;
 	if (!!(ret = pthread_attr_setaffinity_np(&attr, sizeof(cpuset), &cpuset)))
 		goto end;
 
 	if (!!(ret = pthread_create(&worker->thread, &attr, gr_datapath_loop, worker)))
 		goto end;
+	thread_created = true;
 
 	STAILQ_INSERT_TAIL(&workers, worker, next);
+	inserted = true;
 
 	// wait until thread has initialized lcore_id
 	for (unsigned i = 0; !atomic_load(&worker->started); i++) {
@@ -66,14 +70,17 @@ int worker_create(unsigned cpu_id) {
 	}
 
 end:
-	pthread_attr_destroy(&attr);
+	if (attr_inited)
+		pthread_attr_destroy(&attr);
 
 	if (ret == 0) {
 		LOG(INFO, "worker %u started", cpu_id);
 	} else {
 		if (worker != NULL) {
-			STAILQ_REMOVE(&workers, worker, worker, next);
-			pthread_cancel(worker->thread);
+			if (inserted)
+				STAILQ_REMOVE(&workers, worker, worker, next);
+			if (thread_created)
+				pthread_cancel(worker->thread);
 			pthread_cond_destroy(&worker->wakeup.cond);
 			pthread_mutex_destroy(&worker->wakeup.lock);
 			rte_free(worker);
--- a/modules/infra/datapath/control_input.c
+++ b/modules/infra/datapath/control_input.c
@@ -1,12 +1,14 @@
 // SPDX-License-Identifier: BSD-3-Clause
 // Copyright (c) 2024 Christophe Fontaine
 
+#include "config.h"
 #include "control_input.h"
 #include "graph.h"
 #include "log.h"
 #include "mbuf.h"
 #include "mempool.h"
 #include "trace.h"
+#include "worker.h"
 
 LOG_TYPE("graph");
 
@@ -47,7 +49,15 @@ int post_to_stack(control_input_t type, void *data) {
 	if (ret < 0)
 		return errno_set(-ret);
 
+	// adaptive-irq only: kick an idle worker blocked on rte_epoll_wait to drain it.
+	if (gr_config.adaptive_irq)
+		worker_wakeup_any();
+
 	return 0;
+}
+
+bool control_input_pending(void) {
+	return rte_ring_count(control_input_ring) > 0;
 }
 
 static uint16_t control_input_process(
--- a/modules/infra/datapath/main_loop.c
+++ b/modules/infra/datapath/main_loop.c
@@ -21,9 +21,11 @@
 #include <rte_malloc.h>
 
 #include <pthread.h>
+#include <sched.h>
 #include <stdatomic.h>
 #include <sys/epoll.h>
 #include <sys/prctl.h>
+#include <sys/syscall.h>
 #include <unistd.h>
 
 LOG_TYPE("graph");
@@ -188,6 +190,9 @@ static struct rte_rcu_qsbr *rcu;
 
 #define ADAPTIVE_IRQ_EMPTY_WINDOWS 2
 #define ADAPTIVE_IRQ_MAX_EVENTS 32
+// short settle waits let schedutil drop the clock before the worker blocks for good
+#define ADAPTIVE_IRQ_SETTLE_MS 100
+#define ADAPTIVE_IRQ_SETTLE_TRIES 3
 
 // --adaptive-irq: an idle worker arms its rxqs via rte_eth_dev_rx_intr_* and blocks on
 // their eventfds through this thread's epoll until one fires. Arm -EAGAIN means
@@ -199,7 +204,7 @@ adaptive_irq_wait(struct worker *w, vec struct queue_map *rxqs, vec struct queue
 	struct rte_epoll_event events[ADAPTIVE_IRQ_MAX_EVENTS];
 	vec struct queue_map *armed = NULL;
 	struct queue_map *qm, *e;
-	int ret, status = 0;
+	int n, ret, status = 0;
 	bool rearm = false;
 
 	// add the wakeup_fd to the same per-thread epfd rte_epoll_wait uses; done here,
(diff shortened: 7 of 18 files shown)


modules/infra/datapath/main_loop.c
```diff
@@ -182,6 +185,115 @@ static int stats_reload(const struct rte_graph *graph, struct stats_context *ctx
...
+	vec_foreach_ref (e, armed)
+		rte_eth_dev_rx_intr_disable(e->port_id, e->queue_id);
+	vec_free(armed);
+}
+
+// napi: the worker blocks on the rxq IRQ when idle, so schedutil sees low
+// utilization and downclocks the core even at line rate. Pin uclamp_min to the
+// max capacity: the governor runs the core at full speed while the worker is
+// runnable and lets it drop only when it actually sleeps on the interrupt.
+// glibc exposes neither struct sched_attr nor a sched_setattr() wrapper.
```
vjardin: https://lists.gnu.org/archive/html/info-gnu/2025-01/msg00014.html. ?

vjardin (changes requested):
sched_setattr() shall be used from glibc.

MortenBroerup:
Disclaimer: I haven't looked at the patch in detail, so my feedback is high-level only.

Have you measured the RX interrupt wakeup latency? In other words: Are user-space interrupts fast enough for general use, or only for low-traffic hours?
AFAIK, sleep(), usleep() and nanosleep() all end up in the nanosleep() syscall.
With the default timerslack of 50 µs, nanosleep() wakeup latency exceeds 50 µs.
With timerslack 1 - `prctl(PR_SET_TIMERSLACK, 1, 0, 0, 0)` - I have seen nanosleep() wakeup latency around 2.5 µs on hardware, and 15-20 µs on a VMware virtual appliance.

Another detail:
We considered using RX interrupts in the SmartShare too, and I like the concept.
You should consider that other events may want to trigger graph wakeup too...
E.g. a host-originating packet (e.g. a CLI output via SSH).
Or a timer wheel or other pollling-based dataplane timer triggering some high-frequency event, e.g. for calling rte_sched_port_dequeue().

maxime-leroy:
Thanks @MortenBroerup  , both points are spot on. Let me split them.

## Wakeup latency / timerslack

The timerslack concern doesn't apply to the napi path: an idle worker doesn't
sleep on a timer, it blocks on `rte_epoll_wait()` on the rx queue interrupt
eventfd (VFIO/UIO), so the wakeup is interrupt-driven and timerslack never
enters the picture. timerslack only widens the *timeout* part of a sleep/poll,
not the delivery of a real fd event. (For completeness, the existing
sleep-based poll mode already lowers timerslack to 1us via
`PR_SET_TIMERSLACK` at worker start, so it isn't stuck at the 50us default
either.)

So the dominant latency term isn't the scheduler, it's the PMD's interrupt
coalescing. The IRQ fires on the first of "N frames queued" or "holdoff timer
expires". On DPAA2 the default is threshold = 7 frames, holdoff = 100us. Under
any real load the frame threshold trips first and the holdoff never matters;
the holdoff only adds latency in the trickle case (1-2 frames then silence),
i.e. exactly the near-idle regime this mode targets. Worst-case single-packet
wakeup at trickle is therefore bounded by the holdoff (~100us on DPAA2,
tunable via `DPAA2_PORTAL_INTR_TIMEOUT`).

To answer the underlying question directly: this is opt-in (`--napi`) and aimed
at cutting power/clocks during quiet periods, not at minimum-latency max-pps
forwarding. It's "general use during low/moderate load", and you'd leave it off
on a latency-critical fast path. I haven't done a rigorous latency sweep across
PMDs yet; if that would help I can gather numbers.

## Other wakeup sources

Agreed this is the important design point. Two cases:

Control-plane events already work: the worker's own wakeup eventfd is in the
same per-thread epoll set, so a reconfig or shutdown breaks the block
immediately instead of waiting for a packet. That's the generic kick path, so
anything host-originated that needs to poke a sleeping worker can reuse it.

Datapath timers: today grout's datapath is purely packet-driven, there's no
timer wheel, no `rte_sched`, nothing time-driven that a sleeping worker would
starve, so blocking until the next packet is safe as-is. The day a periodic
fast-path event shows up (QoS dequeue / shaping via `rte_sched_port_dequeue()`,
TX pacing, datapath aging), an indefinite block would indeed starve it, and the
fix is the standard one: add a timerfd to the same epoll set (or cap the
epoll_wait timeout to the next deadline) so the worker wakes on a packet *or* a
deadline. Same shape as the wakeup eventfd that's already wired in, so the
extension point is there when it's needed.

main/main.c
```diff
@@ -65,6 +65,11 @@ static void usage(void) {
 	puts("  -m, --socket-mode PERMISSIONS  API socket file permissions (Default: 0660).");
 	puts("  -o, --socket-owner USER:GROUP  API socket file ownership");
 	puts("  -p, --poll-mode                Disable automatic micro-sleep.");
+	puts("  -n, --napi                     Adaptive interrupt (NAPI) rx: idle");
+	puts("                                 workers block on rxq interrupts");
+	puts("                                 instead of polling. Falls back to the");
+	puts("                                 micro-sleep ramp without HW support,");
+	puts("                                 unless --poll-mode is also given.");
```
rjarry: Can we make this text shorter and put the long description in `docs/grout.8.scdoc`.

modules/infra/datapath/main_loop.c
```diff
@@ -182,6 +185,105 @@ static int stats_reload(const struct rte_graph *graph, struct stats_context *ctx
...
+	int ret, status = 0;
+
+	// add the wakeup_fd to the same per-thread epfd rte_epoll_wait uses; done here,
+	// not at worker start, so it lands on the epfd the block actually uses.
+	if (!wakeup_done) {
+		wakeup_ev.epdata.event = EPOLLIN;
+		ret = rte_epoll_ctl(RTE_EPOLL_PER_THREAD, EPOLL_CTL_ADD, w->wakeup_fd, &wakeup_ev);
+		if (ret == 0 || errno == EEXIST)
+			wakeup_done = true;
+	}
```
rjarry: Instead of `static __thread` variables, move `wakeup_ev` into `struct worker`, next to `wakeup_fd`.

modules/infra/datapath/main_loop.c
```diff
@@ -193,7 +295,10 @@ void *gr_datapath_loop(void *priv) {
 	uint32_t sleep, max_sleep_us;
 	struct worker *w = priv;
 	struct rte_graph *graph;
-	unsigned cur, loop;
+	unsigned cur, loop, napi_empty = 0;
+	vec struct queue_map *napi_registered = NULL;
+	vec struct queue_map *napi_rxqs = NULL;
+	bool napi_off = false;
```
rjarry: Reverse xmas tree please.

modules/infra/datapath/main_loop.c
```diff
@@ -292,6 +306,53 @@ napi_wait(struct worker *w, vec struct queue_map *rxqs, vec struct queue_map **r
...
+}
+#endif
+#ifndef SCHED_FLAG_KEEP_ALL
+#define SCHED_FLAG_KEEP_POLICY 0x08
+#define SCHED_FLAG_KEEP_PARAMS 0x10
+#define SCHED_FLAG_KEEP_ALL (SCHED_FLAG_KEEP_POLICY | SCHED_FLAG_KEEP_PARAMS)
+#endif
+#ifndef SCHED_FLAG_UTIL_CLAMP_MIN
+#define SCHED_FLAG_UTIL_CLAMP_MIN 0x20
+#endif
```
rjarry: This patch looks to address an arch-specific issue (the kernel not down clocking a CPU when no threads are running). Maybe we should not try to band-aid this in grout.

rjarry: @david-marchand any opinion?

rjarry (changes requested):
Overall, I don't have any objections. Just some minor concerns regarding where to store variables. And maybe we should table the uclamp_min workaround for later.

I know NAPI is the Linux denomination for this hybrid IRQ/POLL mode of operation. But could we find another name which is more representative of what it does? How about calling it flexible or adaptive IRQ mode?

Also, I would like to have a CLI command to enable/disable the feature at runtime without restarting. This seems possible with the current architecture.

```
grcli graph config set adaptive-irq on
grcli graph config set adaptive-irq off
```

And we could even add another flag for the poll-mode:

```
grcli graph config set poll on
grcli graph config set poll off
```

What do you think?

MortenBroerup:
> Overall, I don't have any objections. Just some minor concerns regarding where to store variables. And maybe we should table the uclamp_min workaround for later.
> 
> I know NAPI is the Linux denomination for this hybrid IRQ/POLL mode of operation. But could we find another name which is more representative of what it does? How about calling it flexible or adaptive IRQ mode?

In the Linux kernel, NAPI is short for New API.
A more descriptive name for Grout would be better. :-)

Naming is hard. Here are some suggestions...
Maybe "adaptive interrupt/poll mode".
Or, if we consider "poll mode" the baseline, "adaptive interrupt mode" might suffice.

> 
> Also, I would like to have a CLI command to enable/disable the feature at runtime without restarting. This seems possible with the current architecture.

**No CLI for this!** I don't see any need for it IRL.
Also, it might it more difficult to implement alternative power management algorithms.

rjarry:
> **No CLI for this!** I don't see any need for it IRL. Also, it might it more difficult to implement alternative power management algorithms.

Why would allowing to change the poll mode at runtime make implementing different power management algorithms difficult? We can keep the startup argument. But I would really like to be able to do everything at runtime.

modules/infra/datapath/main_loop.c
maxime-leroy: It is not arch-specific, it is schedutil, the default governor on modern Linux. schedutil only recomputes frequency at cpufreq_update_util() points (enqueue/dequeue/tick). A napi worker that blocks makes its core tickless-idle (the isolcpus + nohz_full setup every DPDK datapath uses), so schedutil is never re-invoked and the OPP stays frozen at max. Same on x86 with schedutil + isolated cores. And the poll -> blocked-on-IRQ transition only exists inside grout: the governor cannot see it, so no static operator config can express "max while polling, drop while idle".

MortenBroerup:
> > **No CLI for this!** I don't see any need for it IRL. Also, it might it more difficult to implement alternative power management algorithms.
> 
> Why would allowing to change the poll mode at runtime make implementing different power management algorithms difficult? We can keep the startup argument.

Power management algorithms may be mutually exclusive, and possibly only build time configurable.

> But I would really like to be able to do everything at runtime.

Just because it is possible having a CLI command doesn't mean it is sensible!
It is nice for test and development purposes, yes.
But in a production environment, such commands must be hidden away, so they doesn't confuse network engineers working on setting up or reconfiguring a network.

We could introduce a CLI "debug" top level command to hide such commands under.
We could also introduce operator access levels (or access roles), requiring a special command (and maybe even a password) to enable the debug command tree.

rjarry:
`grcli` is way too low level for such requirements. You should see `grcli` as a grout equivalent of `ip` and `sysctl`. It was always meant to be stateless and for power users/developers.

What you're asking for is a high level NETCONF/RESTCONF interface with proper ACL where the low level internals are hidden.

MortenBroerup:
> `grcli` is way too low level for such requirements. You should see `grcli` as a grout equivalent of `ip` and `sysctl`. It was always meant to be stateless and for power users/developers.

OK. Nice to know.

> What you're asking for is a high level NETCONF/RESTCONF interface with proper ACL where the low level internals are hidden.

I'm trying to look ahead, so we don't introduce too much clutter that will be a barrier for long term goals.
Also, having a CLI that resembles what people (network engineers) are used to, makes it easier for them (network engineers) to operate the system. (In my previous job at a network silicon vendor, OEMs/ODMs kept asking for a "Cisco-like CLI".)

If we forget about the ACL concept, we could still think more carefully about organizing CLI commands.
E.g. why is "dhcp" a root level command, and not a subcommand to "address"?

PS: Still learning about Grout. Some [architecture documentation](https://github.com/DPDK/grout/issues/3) would be helpful. ;-)

rjarry:
I think the "cisco like CLI" requirement could be fulfilled by only configuring grout through FRR's vtysh which is deeply inspired of cisco's CLI style.

There are obviously things which cannot be configured through vtysh, such as creating interfaces and such. But we could think of long term plans to make this possible.

What do you think?

PS: I haven't forgotten about that architecture documentation. I'll work on it soon<sup>TM</sup>.

MortenBroerup:
I'm not familiar with vtysh, so it's hard for me to say.
vtysh is GPL, which might be a problem.

It also depends on the architecture of vtysh. If it's openly extensible, so it can integrate with Grout's API for modules in Grout not related to FRR. E.g. the L2 Bridge module and future QoS modules.

Also consider that Grout is designed to be the main controller, configuring Linux accordingly. It is not using Linux as the main controller and mirroring Linux' setup.
I think we want to keep Grout that way, and not make FRR the main controller (where Grout is mirroring FRR).

maxime-leroy:
> Overall, I don't have any objections. Just some minor concerns regarding where to store variables. And maybe we should table the uclamp_min workaround for later.
> 
> I know NAPI is the Linux denomination for this hybrid IRQ/POLL mode of operation. But could we find another name which is more representative of what it does? How about calling it flexible or adaptive IRQ mode?
> 
> Also, I would like to have a CLI command to enable/disable the feature at runtime without restarting. This seems possible with the current architecture.
> 
> ```
> grcli graph config set adaptive-irq on
> grcli graph config set adaptive-irq off
> ```
> 
> And we could even add another flag for the poll-mode:
> 
> ```
> grcli graph config set poll on
> grcli graph config set poll off
> ```
> 
> What do you think?
 
Enabling adaptive-irq at runtime isn't that simple: rxq interrupts are requested at rte_eth_dev_configure() time (intr_conf.rxq), so flipping it live would mean restarting all the ports to reconfigure them. The only way to avoid that port restart would be to always configure the rxq interrupts up front and just gate whether idle workers block, but we don't want interrupt mode enabled unconditionally on every deployment: that path isn't free or well-proven yet (it's PMD-dependent, e.g. the DPAA2 quirks we ran into). That's exactly why it's an opt-in boot flag.

A runtime poll on/off would be cheap on its own (just a flag the datapath reads plus a worker kick), but adaptive-irq on/off drags in the always-on interrupt config above. So I'd rather finalize this PR with the boot flag and revisit the runtime toggles (both poll and adaptive-irq) as a separate change once the interrupt path has more mileage.

rjarry:
> Enabling adaptive-irq at runtime isn't that simple: rxq interrupts are requested at rte_eth_dev_configure() time (intr_conf.rxq), so flipping it live would mean restarting all the ports to reconfigure them. The only way to avoid that port restart would be to always configure the rxq interrupts up front and just gate whether idle workers block, but we don't want interrupt mode enabled unconditionally on every deployment: that path isn't free or well-proven yet (it's PMD-dependent, e.g. the DPAA2 quirks we ran into). That's exactly why it's an opt-in boot flag.
> 
> A runtime poll on/off would be cheap on its own (just a flag the datapath reads plus a worker kick), but adaptive-irq on/off drags in the always-on interrupt config above. So I'd rather finalize this PR with the boot flag and revisit the runtime toggles (both poll and adaptive-irq) as a separate change once the interrupt path has more mileage.

Ok, fair enough. Can we at least rename the CLI flag from `-n --napi` to `-a --adaptive-irq`?

maxime-leroy:
> > Enabling adaptive-irq at runtime isn't that simple: rxq interrupts are requested at rte_eth_dev_configure() time (intr_conf.rxq), so flipping it live would mean restarting all the ports to reconfigure them. The only way to avoid that port restart would be to always configure the rxq interrupts up front and just gate whether idle workers block, but we don't want interrupt mode enabled unconditionally on every deployment: that path isn't free or well-proven yet (it's PMD-dependent, e.g. the DPAA2 quirks we ran into). That's exactly why it's an opt-in boot flag.
> > A runtime poll on/off would be cheap on its own (just a flag the datapath reads plus a worker kick), but adaptive-irq on/off drags in the always-on interrupt config above. So I'd rather finalize this PR with the boot flag and revisit the runtime toggles (both poll and adaptive-irq) as a separate change once the interrupt path has more mileage.
> 
> Ok, fair enough. Can we at least rename the CLI flag from `-n --napi` to `-a --adaptive-irq`?

already done

docs/grout.8.scdoc
```diff
@@ -51,6 +51,14 @@ Grout is a software router based on DPDK _rte_graph_.
 *-p*, *--poll-mode*
 	Disable automatic micro-sleep.
 
+*-a*, *--adaptive-irq*
+	Make idle workers block on their RX queue interrupts instead of
+	busy-polling, so the CPU can idle during quiet periods. An incoming
+	packet raises the interrupt and the worker resumes polling. This
+	implies *--poll-mode*. It requires PMD support for RX queue
```
rjarry: Remove `This implies *--poll-mode*.` It contradicts the next sentence.

main/config.h
```diff
@@ -19,6 +19,7 @@ struct gr_config {
 	unsigned max_mtu;
 	bool test_mode;
 	bool poll_mode;
+	bool adaptive_irq; // --adaptive-irq: adaptive interrupt rx
```
rjarry, line 22: the comment is probably redundant.

main/main.c
```diff
@@ -185,11 +186,12 @@ static bool parse_bool_env(const char *name) {
 static int parse_args(int argc, char **argv) {
 	int c;
 
-#define FLAGS ":M:Vhm:o:pSs:tu:vx"
+#define FLAGS ":aM:Vhm:o:pSs:tu:vx"
 	static struct option long_options[] = {
 		{"help", no_argument, NULL, 'h'},
 		{"max-mtu", required_argument, NULL, 'u'},
 		{"metrics", required_argument, NULL, 'M'},
+		{"adaptive-irq", no_argument, NULL, 'a'},
```
rjarry: Can you keep these in alpha order?

```diff
-#define FLAGS ":M:Vhm:o:pSs:tu:vx"
+#define FLAGS ":M:Vahm:o:pSs:tu:vx"
 	static struct option long_options[] = {
+		{"adaptive-irq", no_argument, NULL, 'a'},
 		{"help", no_argument, NULL, 'h'},
 		{"max-mtu", required_argument, NULL, 'u'},
 		{"metrics", required_argument, NULL, 'M'},
```

main/main.c
```diff
@@ -65,6 +65,7 @@ static void usage(void) {
 	puts("  -m, --socket-mode PERMISSIONS  API socket file permissions (Default: 0660).");
 	puts("  -o, --socket-owner USER:GROUP  API socket file ownership");
 	puts("  -p, --poll-mode                Disable automatic micro-sleep.");
+	puts("  -a, --adaptive-irq             Idle workers block on rxq interrupts.");
 	puts("  -s, --socket PATH              Path the control plane API socket.");
 	puts("                                 Default: GROUT_SOCK_PATH from env or");
```
rjarry, line 70: Also alpha order:

```diff
 	puts("  -S, --syslog                   Redirect logs to syslog.");
 	puts("  -V, --version                  Print version and exit.");
+	puts("  -a, --adaptive-irq             Idle workers block on rxq interrupts.");
 	puts("  -h, --help                     Display this help message and exit.");
 	puts("  -m, --socket-mode PERMISSIONS  API socket file permissions (Default: 0660).");
 	puts("  -o, --socket-owner USER:GROUP  API socket file ownership");
 	puts("  -p, --poll-mode                Disable automatic micro-sleep.");
```

main/main.c
```diff
@@ -233,6 +235,9 @@ static int parse_args(int argc, char **argv) {
 		case 'p':
 			gr_config.poll_mode = true;
 			break;
+		case 'a':
+			gr_config.adaptive_irq = true;
+			break;
```
rjarry: Move as first case (alpha order).

modules/infra/control/port.c
```diff
@@ -152,6 +152,9 @@ int port_configure(struct iface_info_port *p, uint16_t n_txq_min) {
 	if (info.dev_flags != NULL && *info.dev_flags & RTE_ETH_DEV_INTR_LSC) {
 		conf.intr_conf.lsc = 1;
 	}
+	// --adaptive-irq: request RX queue interrupts so idle workers can block on them.
```
rjarry: Can you remove the redundant prefix?

```c
// Request RX queue interrupts so idle workers can block on them.
```

modules/infra/control/worker.c
```diff
@@ -122,10 +133,24 @@ void worker_wait_wakeup(struct worker *w) {
...
 	pthread_mutex_lock(&w->wakeup.lock);
 	w->wakeup.set = true;
 	pthread_cond_signal(&w->wakeup.cond);
 	pthread_mutex_unlock(&w->wakeup.lock);
+
+	// The condvar above only reaches a worker parked on graph == NULL. A adaptive-irq
+	// worker idle on rte_epoll_wait is woken through the eventfd instead.
+	// EAGAIN means a kick is already pending and undrained, which is enough.
+	if (write(w->adaptive_irq.wakeup_fd, &one, sizeof(one)) != sizeof(one) && errno != EAGAIN)
+		LOG(ERR, "worker %u wakeup_fd write: %s", w->cpu_id, strerror(errno));
```
rjarry: Minor nit: you do not manage the `EINTR` error. If the main thread gets interrupted by a signal (e.g. `SIGSTOP`) before it could write to the event fd, it will not retry.

You could change this to (not tested):

```c
	int ret;
	do {
		ret = write(w->adaptive_irq.wakeup_fd, &one, sizeof(one));
	} while (ret != sizeof(one) && errno != EAGAIN && errno != EINTR);
	if (ret < 0)
		LOG(ERR, "worker %u wakeup_fd write: %s", w->cpu_id, strerror(errno));
```

modules/infra/datapath/main_loop.c
```diff
@@ -182,18 +185,124 @@ static int stats_reload(const struct rte_graph *graph, struct stats_context *ctx
...
+	// add the wakeup_fd to the same per-thread epfd rte_epoll_wait uses; done here,
+	// not at worker start, so it lands on the epfd the block actually uses.
+	if (!w->adaptive_irq.wakeup_registered) {
+		w->adaptive_irq.wakeup_ev.epdata.event = EPOLLIN;
+		ret = rte_epoll_ctl(
+			RTE_EPOLL_PER_THREAD,
+			EPOLL_CTL_ADD,
+			w->adaptive_irq.wakeup_fd,
+			&w->adaptive_irq.wakeup_ev
+		);
```
rjarry, line 219: Isn't there anything to cleanup/close/free when the worker exits?

modules/infra/datapath/main_loop.c
```diff
@@ -256,7 +369,18 @@ void *gr_datapath_loop(void *priv) {
...
 	loop = 0;
 	sleep = 0;
+	adaptive_irq_empty = 0;
+	adaptive_irq_off = false;
 	timestamp = rte_rdtsc();
+	// snapshot rxqs once a live config is picked up (next_config is set only after
+	// the control plane finishes mutating w->rxqs) so adaptive_irq_wait never races it.
+	if (gr_config.adaptive_irq) {
+		vec_free(adaptive_irq_rxqs);
+		adaptive_irq_rxqs = NULL;
```
rjarry: nit: vec_free() already sets its argument to NULL.

modules/infra/datapath/main_loop.c
```diff
@@ -276,15 +400,39 @@ void *gr_datapath_loop(void *priv) {
...
-				usleep(sleep);
-				ctx.w_stats->sleep_cycles += rte_rdtsc() - timestamp_tmp;
-				ctx.w_stats->n_sleeps += 1;
+			if (gr_config.adaptive_irq && !adaptive_irq_off) {
+				if (ctx.last_count == 0
+				    && ++adaptive_irq_empty >= ADAPTIVE_IRQ_EMPTY_WINDOWS) {
+					uint64_t now;
+					adaptive_irq_empty = 0;
+					if (adaptive_irq_wait(
+						    w, adaptive_irq_rxqs, &adaptive_irq_registered
```
rjarry: For readability, could you use shorter *local* variable names? (e.g. `adaptive_irq_` -> `airq_`)

modules/infra/control/worker.c
```diff
@@ -143,16 +143,40 @@ void worker_wakeup(struct worker *w) {
 	// The condvar above only reaches a worker parked on graph == NULL. A adaptive-irq
 	// worker idle on rte_epoll_wait is woken through the eventfd instead.
 	// EAGAIN means a kick is already pending and undrained, which is enough.
-	if (write(w->adaptive_irq.wakeup_fd, &one, sizeof(one)) != sizeof(one) && errno != EAGAIN)
+	if (gr_config.adaptive_irq
```
rjarry, line 146: this should go in previous commit?

modules/infra/datapath/control_input.c
```diff
@@ -47,9 +51,19 @@ int post_to_stack(control_input_t type, void *data) {
 	if (ret < 0)
 		return errno_set(-ret);
 
+	// no worker is running the graph: kick one so the injected packet is drained.
+	// seq_cst orders the enqueue before the active-count read (eventcount).
+	if (gr_config.adaptive_irq
+	    && atomic_load_explicit(&gr_worker_active, memory_order_seq_cst) == 0)
+		worker_wakeup_any();
```
rjarry: can you move the boilerplate into `worker_wakeup_any()?

modules/infra/control/worker.c
```diff
@@ -143,16 +143,40 @@ void worker_wakeup(struct worker *w) {
...
 void worker_wakeup_all(void) {
 	struct worker *worker;
 	STAILQ_FOREACH (worker, &workers, next)
 		worker_wakeup(worker);
 }
 
+// Wake one worker that owns rxqs (can drain the control input ring) when idle.
+void worker_wakeup_any(void) {
+	struct worker *worker;
+	STAILQ_FOREACH (worker, &workers, next) {
```
rjarry: Instead of checking whether adaptive IRQ is enabled in all call sites, move the check here.

modules/infra/datapath/main_loop.c
```diff
@@ -260,27 +262,33 @@ adaptive_irq_wait(struct worker *w, vec struct queue_map *rxqs, vec struct queue
...
+			rearm = true;
+	if (rearm) {
 		vec_free(*registered);
 		*registered = NULL;
-		while (read(w->adaptive_irq.wakeup_fd, &kick, sizeof(kick)) > 0)
-			;
 	}
 
+wake:
+	atomic_fetch_add_explicit(&gr_worker_active, 1, memory_order_seq_cst);
```
rjarry: I am not fond of exposing atomics in headers and sharing them between files. Could we wrap this in functions? I does not even need to be static inline.

rjarry (changes requested):

docs/grout.8.scdoc
rjarry: Btw, the commit message also has this leftover from previous implementation which effectively forced poll mode when --napi was specified.

modules/infra/control/worker.c
```diff
@@ -114,11 +127,49 @@ void worker_wait_wakeup(struct worker *w) {
...
 	pthread_mutex_lock(&w->wakeup.lock);
 	w->wakeup.set = true;
 	pthread_cond_signal(&w->wakeup.cond);
 	pthread_mutex_unlock(&w->wakeup.lock);
+
+	// The condvar above only reaches a worker parked on graph == NULL. A napi
+	// worker idle on rte_epoll_wait is woken through the eventfd instead.
+	// EAGAIN means a kick is already pending and undrained, which is enough.
+	if (write(w->wakeup_fd, &val, sizeof(val)) != sizeof(val) && errno != EAGAIN)
+		LOG(ERR, "worker %u wakeup_fd write: %s", w->cpu_id, strerror(errno));
```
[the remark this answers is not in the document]
maxime-leroy: The wakeup_fd is created with EFD_NONBLOCK, so this write() can't return EINTR here:
  it can only succeed or return EAGAIN.

  man 2 eventfd only ever describes the write blocking on counter overflow, and "fails
  with the error EAGAIN if the file descriptor has been made nonblocking" — EINTR is
  not mentioned anywhere on that page. And man 2 write documents EINTR as "interrupted
  by a signal before any data was written", i.e. only while the call is actually
  blocked; a non-blocking eventfd write never blocks.

  Same story in the kernel: eventfd_write() only enters the interruptible wait
  (do_wait_intr_irq() → signal_pending() → -ERESTARTSYS, which the syscall return path
  turns into EINTR) in the !O_NONBLOCK branch. With EFD_NONBLOCK that branch is
  skipped and it returns -EAGAIN (or success) without ever sleeping.

  So the existing != sizeof(val) && errno != EAGAIN check already covers every
  reachable outcome. I'd keep it as is, unless we ever drop EFD_NONBLOCK on this fd.

modules/infra/control/worker.c
maxime-leroy: The wakeup_fd is created with EFD_NONBLOCK, so this write() can't return EINTR here:
  it can only succeed or return EAGAIN.

  man 2 eventfd only ever describes the write blocking on counter overflow, and "fails
  with the error EAGAIN if the file descriptor has been made nonblocking" — EINTR is
  not mentioned anywhere on that page. And man 2 write documents EINTR as "interrupted
  by a signal before any data was written", i.e. only while the call is actually
  blocked; a non-blocking eventfd write never blocks.

  Same story in the kernel: eventfd_write() only enters the interruptible wait
  (do_wait_intr_irq() → signal_pending() → -ERESTARTSYS, which the syscall return path
  turns into EINTR) in the !O_NONBLOCK branch. With EFD_NONBLOCK that branch is
  skipped and it returns -EAGAIN (or success) without ever sleeping.

  So the existing != sizeof(val) && errno != EAGAIN check already covers every
  reachable outcome. I'd keep it as is, unless we ever drop EFD_NONBLOCK on this fd.

  (Minor: the proposed loop as written exits on EINTR instead of retrying, and would
  spin on a genuine error. If we ever did need EINTR handling, it'd be do { n =
  write(...); } while (n < 0 && errno == EINTR); then the same EAGAIN check.)

modules/infra/datapath/main_loop.c
maxime-leroy: The worker's wakeup_fd is closed in worker_destroy(). Closing an fd auto-removes
  it from any epoll set, so its registration needs no explicit EPOLL_CTL_DEL.
  - The rxq interrupt fds belong to the PMD, which closes them on dev_stop(); closed
  fds drop out of the epoll set as well. We deliberately don't DEL them: queues can
  share a single portal eventfd (dpaa2), so deleting one would clobber its siblings.

  There is one real leak, but it's on the DPDK side: the per-thread epoll instance
  itself. Using RTE_EPOLL_PER_THREAD makes DPDK lazily create a per-thread epfd
  (rte_intr_tls_epfd() → epoll_create()), and DPDK never closes it — not on thread
  exit, not in rte_thread_unregister(). So a worker that used adaptive-irq and is then
  torn down at runtime (e.g. a datapath CPU reconfig) leaks that one fd. It's generic
  to any DPDK app using per-thread rx interrupts, arch-independent; adaptive-irq is
  just the first thing in grout to trigger it. It isn't cleanly fixable from grout
  (rte_intr_tls_epfd is __rte_internal and the fd isn't ours) — the fix belongs in
  DPDK (close the tls epfd in __rte_thread_uninit()). I have a small DPDK patch for it
  that I'll send upstream.

modules/infra/datapath/control_input.c
maxime-leroy: Done for the active-count guard: worker_wakeup_any() now early-returns when a worker is already running the graph, so the caller keeps only the policy check.

 I did keep the gr_config.adaptive_irq test in post_to_stack and out of worker_wakeup_any(), because they are different concerns.

 worker_wakeup_any() -> worker_wakeup() -> worker_kick() does two things: signal the per-worker condvar (generic, unparks a worker blocked on graph == NULL, used in poll mode too) and write the wakeup_fd eventfd (adaptive-irq specific, already self-gated inside worker_kick()). So the wakeup primitive is generic - its siblings worker_wakeup_all() / worker_rearm_all() build on the same worker_kick() - and I'd rather not couple it to the feature flag.

The adaptive_irq test is policy: only an adaptive-irq worker can sit idle-blocked on the control-input path and need an explicit kick; in poll mode a running worker drains the ring on its next walk. That belongs to the caller that knows the context.

rjarry (approved):

PR #632, opened 2026-06-25
frr: only redirect to VXLAN for remote L3 nexthops

zeeke:
The VXLAN interface override for L3 nexthops in VRFs with an L3VNI was
applied unconditionally, before checking whether the nexthop is actually
remote. A local nexthop (e.g. a directly connected host resolved via
ARP) would get its iface_id rewritten to the VXLAN tunnel, causing
traffic to be encapsulated instead of forwarded through the local port.

Move the iface_id override inside the RMAC lookup block so it only takes
effect when l3vni_rmac_get returns a cached remote MAC, which is the
condition that also sets GR_NH_F_REMOTE.

- Moved VXLAN `iface_id` override for L3 nexthops behind the remote-MAC lookup in `frr/rt_grout.c`, so only nexthops with a cached RMAC are redirected to the VXLAN tunnel.
- Extended `smoke/evpn_l3vpn_frr_test.sh` to cover a tenant VRF route toward a host-local subnet and verify traffic uses the local port (`p1`) instead of being VXLAN-encapsulated.

### DIFF
--- a/frr/rt_grout.c
+++ b/frr/rt_grout.c
@@ -714,8 +714,6 @@ grout_add_nexthop(uint32_t nh_id, gr_nh_origin_t origin, const struct nexthop *n
 		// the VRF (SVI in FRR's model) to the VXLAN interface. Grout
 		// routes packets directly through the VXLAN tunnel.
 		vxlan_iface_id = l3vni_get_vxlan(req->nh.vrf_id);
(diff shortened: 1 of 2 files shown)


rjarry (commented):
That fix looks good to me in practice. Could you add this trailer before your signoff?

```
Fixes: b468374cf98d ("frr: add L3VPN support")
```

rjarry (approved):

PR #633, opened 2026-06-26
frr: fix EVPN type-5 forwarding for IPv6 prefixes

zeeke:
depends on:
- https://github.com/DPDK/grout/pull/632

EVPN type-5 routes for IPv6 prefixes use v4-mapped nexthop
addresses (::ffff:X.X.X.X) when the VTEP uses IPv4.
These failed for two reasons: grout_add_nexthop() looked up the
RMAC cache with an IPADDR_V6 key but grout_neigh_update_ctx()
stored it under IPADDR_V4, so the hash never matched; and the
nexthop carried GR_AF_IP6, causing vxlan_output to look for an
IPv6 underlay route that does not exist.

Fix both by detecting v4-mapped addresses: extract the embedded
IPv4, store the nexthop as GR_AF_IP4, and look up the RMAC with
the matching IPADDR_V4 key. Add the VXLAN interface redirect that
was missing from the IPv6 path. In ip6_output, derive the VTEP
address family from l3->af instead of hardcoding GR_AF_IP6.

EVPN Type-5 (IP prefix) nexthops now handle IPv6 v4-mapped next-hops (`::ffff:X.X.X.X`) correctly when the VTEP uses IPv4: in `grout_add_nexthop`, `IS_MAPPED_IPV6` triggers `ipv4_mapped_ipv6_to_ipv4`, and the L3 nexthop is stored as IPv4 (`l3->af = GR_AF_IP4` with `l3->ipv4` set to the embedded IPv4) rather than forcing IPv6 (`GR_AF_IP6`). RMAC cache lookup is performed with the matching VTEP address representation by setting the VTEP key to `vtep.ipa_type = IPADDR_V4` / `vtep.ipaddr_v4` before calling `l3vni_rmac_get`, so remote nexthops resolve against the RMAC cached under the IPv4 VTEP entry instead of failing and causing VXLAN forwarding to use an invalid IPv6 underlay route. Native (non-mapped) IPv6 nexthops continue to use `GR_AF_IP6` with IPv6 VTEP data.

The missing VXLAN interface redirect on the IPv6 (including mapped) path was fixed: `req->nh.iface_id` is now set to `vxlan_iface_id` only inside the `if (rmac != NULL)` block (guarded by `vxlan_iface_id != GR_IFACE_ID_UNDEF`), alongside setting the remote nexthop MAC and `GR_NH_F_REMOTE` flag.

VXLAN datapath VTEP address-family handling was generalized to match the L3 address family: `ip_output_process` and `ip6_output_process` now set `eth_data->vtep.af = l3->af` and populate `eth_data->vtep.ipv4` when `l3->af == GR_AF_IP4`, otherwise populate `eth_data->vtep.ipv6` (removing hardcoded IPv4/IPv6 assumptions on the output path).

EVPN L3VPN smoke coverage was extended to dual-stack: `smoke/evpn_l3vpn_frr_test.sh` enables IPv6 forwarding and adds IPv6 addressing/default routes on both namespaces, updates FRR to redistribute and advertise both IPv4/IPv6 unicast in `address-family l2vpn evpn`, adds IPv6 Type-5 readiness/installation waits for `fd00:16::/64` and `fd00:48::/64` (via `wait_event` using `route6 add`), validates IPv6 VRF route installation with `ip -n evpn-peer -6 route show`, and extends connectivity checks with IPv6 pings (`ping -6`) across the VXLAN L3 overlay.

### DIFF
--- a/frr/rt_grout.c
+++ b/frr/rt_grout.c
@@ -736,11 +736,24 @@ grout_add_nexthop(uint32_t nh_id, gr_nh_origin_t origin, const struct nexthop *n
 		case NEXTHOP_TYPE_IPV6:
 		case NEXTHOP_TYPE_IPV6_IFINDEX:
 			l3 = (struct gr_nexthop_info_l3 *)req->nh.info;
-			l3->af = GR_AF_IP6;
-			memcpy(&l3->ipv6, &nh->gate.ipv6, sizeof(l3->ipv6));
+			// EVPN type-5 IPv6 prefixes use v4-mapped nexthops
+			// (::ffff:X.X.X.X) because the VTEP is always IPv4.
+			// Store as native IPv4 so vxlan_output uses IPv4
+			// encapsulation to reach the remote VTEP.
+			if (IS_MAPPED_IPV6(&nh->gate.ipv6)) {
+				struct in_addr v4;
+				ipv4_mapped_ipv6_to_ipv4(&nh->gate.ipv6, &v4);
+				l3->af = GR_AF_IP4;
+				l3->ipv4 = v4.s_addr;
+				vtep.ipa_type = IPADDR_V4;
+				vtep.ipaddr_v4 = v4;
+			} else {
+				l3->af = GR_AF_IP6;
+				memcpy(&l3->ipv6, &nh->gate.ipv6, sizeof(l3->ipv6));
+				vtep.ipa_type = IPADDR_V6;
+				vtep.ipaddr_v6 = nh->gate.ipv6;
+			}
 			// Apply cached RMAC from EVPN NEIGH install if available.
-			vtep.ipa_type = IPADDR_V6;
-			vtep.ipaddr_v6 = nh->gate.ipv6;
 			rmac = l3vni_rmac_get(req->nh.vrf_id, &vtep);
 			if (rmac != NULL) {
 				memcpy(&l3->mac, rmac, sizeof(l3->mac));
--- a/frr/rt_grout.c
+++ b/frr/rt_grout.c
@@ -745,6 +745,8 @@ grout_add_nexthop(uint32_t nh_id, gr_nh_origin_t origin, const struct nexthop *n
 			if (rmac != NULL) {
 				memcpy(&l3->mac, rmac, sizeof(l3->mac));
 				l3->flags |= GR_NH_F_REMOTE;
+				if (vxlan_iface_id != GR_IFACE_ID_UNDEF)
+					req->nh.iface_id = vxlan_iface_id;
 			}
 			break;
 		case NEXTHOP_TYPE_IFINDEX:
--- a/modules/ip6/datapath/ip6_output.c
+++ b/modules/ip6/datapath/ip6_output.c
@@ -117,8 +117,11 @@ ip6_output_process(struct rte_graph *graph, struct rte_node *node, void **objs, 
 			eth_data->dst = l3->mac;
 		eth_data->ether_type = RTE_BE16(RTE_ETHER_TYPE_IPV6);
 		if (iface->type == GR_IFACE_TYPE_VXLAN) {
-			eth_data->vtep.af = GR_AF_IP6;
-			eth_data->vtep.ipv6 = l3->ipv6;
+			eth_data->vtep.af = l3->af;
+			if (l3->af == GR_AF_IP4)
+				eth_data->vtep.ipv4 = l3->ipv4;
+			else
+				eth_data->vtep.ipv6 = l3->ipv6;
 		} else {
 			eth_data->vtep.af = GR_AF_UNSPEC;
 		}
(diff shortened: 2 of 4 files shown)


modules/ip6/datapath/ip6_output.c
```diff
@@ -117,8 +117,11 @@ ip6_output_process(struct rte_graph *graph, struct rte_node *node, void **objs,
 			eth_data->dst = l3->mac;
 		eth_data->ether_type = RTE_BE16(RTE_ETHER_TYPE_IPV6);
 		if (iface->type == GR_IFACE_TYPE_VXLAN) {
-			eth_data->vtep.af = GR_AF_IP6;
-			eth_data->vtep.ipv6 = l3->ipv6;
+			eth_data->vtep.af = l3->af;
+			if (l3->af == GR_AF_IP4)
+				eth_data->vtep.ipv4 = l3->ipv4;
+			else
+				eth_data->vtep.ipv6 = l3->ipv6;
```
rjarry, line 124: There should be a similar fix in ip_output_process?

```diff
diff --git modules/ip/datapath/ip_output.c modules/ip/datapath/ip_output.c
index 78214edc7c1b..9a3bcc24a4fb 100644
--- modules/ip/datapath/ip_output.c
+++ modules/ip/datapath/ip_output.c
@@ -135,8 +135,11 @@ ip_output_process(struct rte_graph *graph, struct rte_node *node, void **objs, u
                eth_data->dst = l3->mac;
                eth_data->ether_type = RTE_BE16(RTE_ETHER_TYPE_IPV4);
                if (iface->type == GR_IFACE_TYPE_VXLAN) {
-                       eth_data->vtep.af = GR_AF_IP4;
-                       eth_data->vtep.ipv4 = l3->ipv4;
+                       eth_data->vtep.af = l3->af;
+                       if (l3->af == GR_AF_IP4)
+                               eth_data->vtep.ipv4 = l3->ipv4;
+                       else
+                               eth_data->vtep.ipv6 = l3->ipv6;
                } else {
                        eth_data->vtep.af = GR_AF_UNSPEC;
                }
```

frr/rt_grout.c
```diff
@@ -736,15 +735,29 @@ grout_add_nexthop(uint32_t nh_id, gr_nh_origin_t origin, const struct nexthop *n
...
+				memcpy(&l3->ipv6, &nh->gate.ipv6, sizeof(l3->ipv6));
+				vtep.ipa_type = IPADDR_V6;
+				vtep.ipaddr_v6 = nh->gate.ipv6;
+			}
 			rmac = l3vni_rmac_get(req->nh.vrf_id, &vtep);
 			if (rmac != NULL) {
 				memcpy(&l3->mac, rmac, sizeof(l3->mac));
 				l3->flags |= GR_NH_F_REMOTE;
+				if (vxlan_iface_id != GR_IFACE_ID_UNDEF)
+					req->nh.iface_id = vxlan_iface_id;
```
rjarry, line 762: This should have been part of #632 (I missed it during review). Could you split it into a separate commit that has `Fixes: 45afc2b00cbf ("frr: only redirect to VXLAN for remote L3 nexthops")` ?

frr/rt_grout.c
```diff
@@ -721,7 +721,6 @@ grout_add_nexthop(uint32_t nh_id, gr_nh_origin_t origin, const struct nexthop *n
 			l3 = (struct gr_nexthop_info_l3 *)req->nh.info;
 			l3->af = GR_AF_IP4;
 			memcpy(&l3->ipv4, &nh->gate.ipv4, sizeof(l3->ipv4));
-			// Apply cached RMAC from EVPN NEIGH install if available.
```
rjarry, line 724: Keep this comment, it cannot hurt.

frr/rt_grout.c
```diff
@@ -736,15 +735,29 @@ grout_add_nexthop(uint32_t nh_id, gr_nh_origin_t origin, const struct nexthop *n
 		case NEXTHOP_TYPE_IPV6:
 		case NEXTHOP_TYPE_IPV6_IFINDEX:
 			l3 = (struct gr_nexthop_info_l3 *)req->nh.info;
-			l3->af = GR_AF_IP6;
-			memcpy(&l3->ipv6, &nh->gate.ipv6, sizeof(l3->ipv6));
-			// Apply cached RMAC from EVPN NEIGH install if available.
```
rjarry, line 741: Same here, keep the comment just above l3vni_rmac_get():

https://github.com/DPDK/grout/blob/45afc2b00cbf65004e07de462579b19f9db6aaec/frr/rt_grout.c#L741
https://github.com/DPDK/grout/blob/45afc2b00cbf65004e07de462579b19f9db6aaec/frr/rt_grout.c#L744

rjarry (changes requested):

modules/ip6/datapath/ip6_output.c
zeeke: sure, that's to encapsulate ipv4 packets into ipv6, right?

rjarry: yes

rjarry (approved):

PR #638, opened 2026-07-03
Fix L3VNI mapping and restrict to one VXLAN per VRF

rjarry:
The L3VNI forward table maps a VRF to a single VXLAN interface for EVPN type-5 nexthop redirection. Two issues exist with this mapping:

When a VXLAN interface leaves VRF mode (e.g. moved to a bridge as an L2 VNI) or is deleted while not in VRF mode, the mapping is never cleaned up or is cleaned up using the wrong key. This is fixed by adding l3vni_del_iface() which scans the forward table by interface ID and calling it on mode changes and deletion.

Additionally, nothing prevents creating multiple VXLAN interfaces in VRF mode in the same VRF. On Linux this situation does not arise because all VXLAN devices sit in bridges and the L2/L3 distinction comes from FRR configuration ("vrf / vni"), not from how the device is attached. In grout, VRF mode itself implies L3 VNI, so having two VXLANs in VRF mode in the same VRF silently overwrites the forward mapping and breaks L3 routing for the previous one. This is now rejected with EADDRINUSE at configuration time.

A smoke test exercises L2 VNI and L3 VNI coexistence in the same VRF, verifying both type-5 inter-subnet routing and type-2/type-3 intra-subnet bridging.

- Added `l3vni_del_iface()` to remove L3VNI forward-table entries by VXLAN interface ID, and wired it into `grout_link_change()` to clear the mapping on VXLAN mode transitions to `GR_IFACE_MODE_BOND`/`GR_IFACE_MODE_BRIDGE`, and to broaden VXLAN deletion cleanup: when `gr_vxlan` exists on the delete path, it now unconditionally clears by interface ID via `l3vni_del_iface(gr_if->id)` (instead of only calling `l3vni_del(gr_if->base.vrf_id)` when the interface was in VRF mode).
- Rejected reconfiguring a VXLAN into VRF mode when another VXLAN is already in `GR_IFACE_MODE_VRF` for the same `vrf_id`; `iface_vxlan_reconfig` now returns `EADDRINUSE` to prevent silent overwrites of the L3VNI forward mapping.
- Added `smoke/evpn_l2l3vni_frr_test.sh` to verify L2 VNI + L3 VNI coexistence within the same tenant VRF, covering EVPN Type-5 inter-subnet routing and type-2/type-3 intra-subnet bridging, plus end-to-end reachability checks.

### DIFF
--- a/smoke/evpn_l2l3vni_frr_test.sh
+++ b/smoke/evpn_l2l3vni_frr_test.sh
@@ -1,0 +1,300 @@
+#!/bin/bash
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2026 Robin Jarry
+
+# This test verifies that L2 VNI and L3 VNI coexist correctly in the same VRF.
+# When both an L3 VNI (1000) and an L2 VNI (100) are present, EVPN type-5
+# (IP prefix) L3VPN routing must use the L3 VNI VXLAN interface and L2
+# bridging must use the L2 VNI.
+#
+# Success criteria:
+#   - Type-5 routes are exchanged and installed via the L3 VNI.
+#   - Host-A and Host-B can ping each other through the L3 VXLAN overlay.
+#   - Type-2/type-3 routes are exchanged via the L2 VNI.
+#   - Host-C and Host-D can ping each other through the L2 VXLAN overlay.
+#   - Host-A and Host-B can talk to both Host-C and Host-D crossing the L3
+#     VXLAN boundary.
+#
+#                                BGP/VXLAN underlay
+#                    +---------+    172.16.0.0/24    +---------+
+#                    |  x-p0   |-+-+-+-+-+-+-+-+-+-+-|  p0     |
+#   .----------------|         |--------.    .-------|         |--------------.
+#   | evpn-peer      +---------+        |    |       +---------+       grout  |
+#   |    ............... .1 ....        |    |    ...... .2 ..............    |
+#   |    .                     .        |    |    .                      .    |
+#   | .- . - - - - - - - - - - . - - .  |    | .- . - - - - - - - - - - -.-.  |
+#   | '  .        vrf tenant   .     '  |    | '  .  vrf tenant          .  ' |
+#   | '  .                     .     '  |    | '  .                      .  ' |
+#   | '  .                     .     '  |    | '  .                      .  ' |
+#   | '  .                     .     '  |    | '  .                      .  ' |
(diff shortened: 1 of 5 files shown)


rjarry:
@zeeke fyi

smoke/evpn_l2l3vni_frr_test.sh
```diff
@@ -0,0 +1,281 @@
...
+
+# -- Verify L3 connectivity through L3 VNI VXLAN overlay ----------------------
+ip netns exec host-b ping -i0.1 -c3 -W1 16.0.0.2
+ip netns exec host-a ping -i0.1 -c3 -W1 48.0.0.2
+
+# -- Verify L2 connectivity through L2 VNI VXLAN overlay ----------------------
+wait_event "flood add: vtep vrf=main 172.16.0.1 vni=100"
+
+ip netns exec host-c ping -i0.1 -c3 -W1 10.0.0.3
+ip netns exec host-d ping -i0.1 -c3 -W1 10.0.0.2
```
zeeke, line 287: I'm wondering if we can/should add cross VNI connectivity, after setting the bridge under the `tenant` VRF

```suggestion
ip netns exec host-d ping -i0.1 -c3 -W1 10.0.0.2

# -- Verify cross VNI connectivity ----------------------
ip netns exec host-a ping -i0.1 -c3 -W1 10.0.0.2
ip netns exec host-a ping -i0.1 -c3 -W1 10.0.0.3
ip netns exec host-b ping -i0.1 -c3 -W1 10.0.0.2
ip netns exec host-b ping -i0.1 -c3 -W1 10.0.0.3

```

smoke/evpn_l2l3vni_frr_test.sh
```diff
@@ -0,0 +1,281 @@
...
+EOF
+
+# right side (grout) setup L3VPN + L2VNI ---------------------------------------
+create_vrf tenant
+
+# L3 VNI VXLAN in VRF mode (no bridge needed in grout)
+grcli interface add vxlan vxlan-l3 vni 1000 local 172.16.0.2 vrf tenant
+
+# L2 VNI: create directly in the bridge domain.
+grcli interface add bridge br100
```
zeeke: Can we put the bridge into the tenant VRF? This way, the packets that are not routed to the underlying L2 network can be routed via the VRF routing table. 
```suggestion
grcli interface add bridge br100 vrf tenant
```

zeeke (approved):
LGTM

PR #639, opened 2026-07-06
L3vpn ipv6 fix

rjarry:
- `frr/rt_grout.c`: In `grout_neigh_update_ctx`, the IPv4-only VTEP check is now gated by FRR version. It is applied only for FRR versions older than `10.6.0` (non-IPv4 VTEPs are logged and skipped). For `10.6.0` and newer, this function no longer performs the IPv4 VTEP rejection at that point.
- `smoke/evpn_l3vpn_vxlan6_frr_test.sh`: Added EVPN Type-5 L3VPN smoke coverage over a VXLAN overlay with IPv4 and IPv6, including FRR version gating (`zebra` < `10.6.0` exits with code `125`), VNI/L3 VNI recognition waits, Type-5 route installation checks via `wait_event`, nexthop validation, and end-to-end IPv4/IPv6 ping between hosts.
- `smoke/bgp6_srv6_frr_test.sh`: Adjusted BGP route-verification retry behavior for SRv6 (`seg6`) checks: fewer maximum retries and fixed retry timing changes, plus a loop-structure adjustment to increment attempts after sleeping.
- `smoke/evpn_l2l3vni_frr_test.sh`: Replaced FRR `vtysh` polling for EVPN Type-5 routes with dataplane/route-event synchronization: wait for `route4 add` events, then poll the peer Linux routing table for the expected learned prefix.
- `smoke/evpn_l3vpn_frr_test.sh`: Replaced broad FRR/BGP CLI polling for EVPN Type-5 route exchange (IPv4/IPv6) with narrower `wait_event` synchronization for Type-5 route installation, followed by peer Linux routing-table verification for the corresponding prefixes.
- `smoke/evpn_vxlan6_frr_test.sh`: Updated synchronization/validation to be event- and dataplane-driven:
  - EVPN type-3 flood VTEP exchange now waits for Grout `flood add` before checking Linux bridge FDB learning (tightened retries and fail-fast behavior).
  - EVPN type-2 validation now prioritizes peer Linux bridge FDB checks (`extern` entries) instead of polling FRR `vtysh` EVPN Type-2 routes.
- `smoke/evpn_vxlan_frr_test.sh`: Similar to the VXLAN6 test—type-3 flood VTEP uses `wait_event` timing before FDB checks, and type-2 validation relies on `wait_event`/FDB-based checks for remote MAC entries with reduced retry budgets.
- `smoke/ospf6_frr_test.sh`: Switched OSPFv6 convergence verification from polling FRR JSON output to `wait_event` for a `route6 add` event for the expected prefix with origin `ospf`.
- `smoke/ospf_frr_test.sh`: Switched OSPFv2 convergence verification from polling FRR JSON output to `wait_event` for a `route4 add` event with origin `ospf`.

### DIFF
--- a/frr/rt_grout.c
+++ b/frr/rt_grout.c
@@ -1060,10 +1060,12 @@ enum zebra_dplane_result grout_neigh_update_ctx(struct zebra_dplane_ctx *ctx) {
 	bool add = dplane_ctx_get_op(ctx) != DPLANE_OP_NEIGH_DELETE;
 	uint16_t vrf_id = vrf_frr_to_grout(dplane_ctx_get_vrf(ctx));
 
+#if CURRENT_FRR_VERSION < MAKE_FRRVERSION(10, 6, 0)
 	if (addr->ipa_type != IPADDR_V4) {
 		gr_log_debug("only IPv4 VTEP addresses supported, skip");
 		return ZEBRA_DPLANE_REQUEST_SUCCESS;
 	}
+#endif
 
 	// Cache the RMAC for later use by grout_add_nexthop. We cannot
 	// create a separate nexthop here because grout's L3 nexthop hash
--- a/smoke/bgp6_srv6_frr_test.sh
+++ b/smoke/bgp6_srv6_frr_test.sh
@@ -195,20 +195,15 @@ wait_event -t 20 'route4 add: vrf=gr-vrf1 16.0.0.0/24 origin=bgp via type=SRv6 '
 wait_event -t 20 'route6 add: vrf=main 2001:db8:2:2:100::/128 origin=bgp via type=SRv6-local '
 
 attempts=0
-while ! vtysh -N bgp-peer -c "show ip route vrf vrf1" | grep -q "B>\* 16.1.0.0/24 .*seg6" ; do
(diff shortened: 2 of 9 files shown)


frr/rt_grout.c
```diff
@@ -1060,10 +1060,12 @@ enum zebra_dplane_result grout_neigh_update_ctx(struct zebra_dplane_ctx *ctx) {
 	bool add = dplane_ctx_get_op(ctx) != DPLANE_OP_NEIGH_DELETE;
 	uint16_t vrf_id = vrf_frr_to_grout(dplane_ctx_get_vrf(ctx));
 
+#if CURRENT_FRR_VERSION < MAKE_FRRVERSION(10, 6, 0)
```
christophefontaine, line 1063: Do we want to support FRR < 10.6 ?
@maxime-leroy any opinion about that ?

rjarry: I was thinking we keep frr 10.5 until 10.7 is out and then shift to 10.6 as default with 10.7 as supported as well ?

maxime-leroy: I havent tested yet frr 10.6 on my target.

maxime-leroy (approved):

PR #640, opened 2026-07-09
ctlplane: set loose reverse path filtering on TAP representors

rjarry:
When rp_filter=1 (strict mode) is enabled, the kernel does a reverse path FIB lookup on incoming packets to verify that the source address is reachable via the same interface. With only oif rules, this lookup falls through to the main/VRF table where the default route points to the TUN loopback, causing an interface mismatch and dropping the packet.

Add a matching iif rule for each TAP so that the rp_filter FIB lookup hits table 999 and finds the default route via the same TAP the packet arrived on. This does not affect actual routing of return packets which still go through the TUN loopback into the datapath as before.

Update the stale rule flush logic to also collect and clean up iif rules on restart.

Enable rp_filter in all smoke tests to exercise the added iif rules.

- `smoke/_init.sh`: after bringing up `lo`, sets IPv4 strict reverse path filtering for namespace defaults (`net.ipv4.conf.all.rp_filter=1` and `net.ipv4.conf.default.rp_filter=1`); when `run_grout` is enabled, exports `GROUT_OVERRIDE_RP_FILTER=true` into the grout environment.
- `Containerfile` + `main/grout.default`: set `GROUT_OVERRIDE_RP_FILTER=true` in the container/grout defaults (with comments that it’s required when `net.ipv4.conf.all.rp_filter` is `1`/strict where grout runs).
- `docs/grout.8.scdoc`: documents `GROUT_OVERRIDE_RP_FILTER` and that it configures `rp_filter=2` (loose) on control plane TAP interfaces when strict (`rp_filter=1`) is enabled globally.
- `main`: adds `gr_config.override_rp_filter`, parsed from `GROUT_OVERRIDE_RP_FILTER`; when set, emits a `NOTICE` that reverse path filtering will be bypassed.
- `modules/infra/control/ctlplane.c`: during control-plane TAP creation, when `gr_config.override_rp_filter` is enabled, writes `/proc/sys/net/ipv4/conf/<iface>/rp_filter=2` for the TAP (after the existing `/keep_addr_on_down` sysctl write) and logs `WARNING` if the `rp_filter` sysctl can’t be opened/written.
- `modules/infra/control/netlink.c`: configures control-plane policy routing using table `999` for both IPv4 and IPv6 by installing per-TAP policy rules that match `oif=<tap>` via `FRA_OIFNAME` and a `default dev <tap>` route in table `999`; on init, flushes stale state for table `999` by dumping `RTM_GETRULE`/`RTM_GETROUTE`, collecting entries where `table == 999`, and deleting the corresponding policy rules (by saved `FRA_OIFNAME`) and routes (in table `999`) to avoid orphan kernel entries across crash/restart cycles.

### DIFF
--- a/modules/infra/control/ctlplane.c
+++ b/modules/infra/control/ctlplane.c
@@ -321,6 +321,21 @@ static void cp_create(struct iface *iface) {
 		LOG(WARNING, "fopen(%s): %s", path, strerror(errno));
 	}
 
+	if (gr_config.override_rp_filter) {
+		// Set loose reverse path filtering on the TAP so that packets
+		// delivered by grout are not dropped by rp_filter. The effective
+		// rp_filter mode is max(conf.all, conf.<iface>), so setting the
+		// per-interface value to 2 (loose) overrides a global strict setting.
+		snprintf(path, sizeof(path), "/proc/sys/net/ipv4/conf/%s/rp_filter", iface->name);
+		f = fopen(path, "w");
+		if (f != NULL) {
+			fputs("2", f);
+			fclose(f);
+		} else {
+			LOG(WARNING, "fopen(%s): %s", path, strerror(errno));
+		}
+	}
+
 	iface->cp_ev = event_new(
 		ev_base,
 		iface->cp_fd,
--- a/smoke/_init.sh
+++ b/smoke/_init.sh
@@ -33,6 +33,8 @@ if [ "${INTERACTIVE:-false}" = true ] && [ -z "${SMOKE_TMUX_SOCK:-}" ]; then
 fi
 
 ip link set lo up
+sysctl -qw net.ipv4.conf.all.rp_filter=1
+sysctl -qw net.ipv4.conf.default.rp_filter=1
 
 : "${test_frr:=false}"
 
@@ -265,6 +267,7 @@ llocal_addr() {
 if [ "$run_grout" = true ]; then
 	smoke_setenv GROUT_SOCK_PATH "$tmp/grout.sock"
 	smoke_setenv GROUT_OVERRIDE_DEFAULT_ROUTE true
+	smoke_setenv GROUT_OVERRIDE_RP_FILTER true
 fi
 if [ -n "${builddir}" ]; then
 	smoke_setenv PATH "$builddir:$PATH"
--- a/Containerfile
+++ b/Containerfile
@@ -23,5 +23,6 @@ FROM scratch
 COPY --from=builder /tmp/null/ /
 STOPSIGNAL SIGRTMIN+3
 ENV GROUT_OVERRIDE_DEFAULT_ROUTE=true
+ENV GROUT_OVERRIDE_RP_FILTER=true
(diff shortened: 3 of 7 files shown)


smoke/_init.sh
```diff
@@ -33,6 +33,8 @@ if [ "${INTERACTIVE:-false}" = true ] && [ -z "${SMOKE_TMUX_SOCK:-}" ]; then
 fi
 
 ip link set lo up
+sysctl -qw net.ipv4.conf.all.rp_filter=1
+sysctl -qw net.ipv4.conf.default.rp_filter=1
```
christophefontaine, line 37: In that commit, should we remove the other rp_filter config ?
For example:
`smoke/bgp6_srv6_frr_test.sh:ip netns exec bgp-peer sysctl -w net.ipv4.conf.all.rp_filter=0`

rjarry: These settings are for the bgp-peer netns that does not run grout. Just linux stack. I can try removing them but I think they are needed because of the assymetric vxlan routing.

christophefontaine: ack, my bad !

christophefontaine (approved):

smoke/_init.sh
maxime-leroy: The default value is zero on kernel, why always enabled it ?

rjarry: Setting it to 2 (loose) is the only way to override anything set by system admins.

modules/infra/control/ctlplane.c
```diff
@@ -321,6 +321,19 @@ static void cp_create(struct iface *iface) {
 		LOG(WARNING, "fopen(%s): %s", path, strerror(errno));
 	}
 
+	// Set loose reverse path filtering on the TAP so that packets
+	// delivered by grout are not dropped by rp_filter. The effective
+	// rp_filter mode is max(conf.all, conf.<iface>), so setting the
+	// per-interface value to 2 (loose) overrides a global strict setting.
+	snprintf(path, sizeof(path), "/proc/sys/net/ipv4/conf/%s/rp_filter", iface->name);
```
maxime-leroy: rp_filter is disable by default, why you want to force at loose for tap. -> NACK

maxime-leroy (changes requested):

modules/infra/control/ctlplane.c
rjarry: Setting it to 2 (loose) is the **only** way to override `net.ipv4.conf.all=1` set by system admins.

maxime-leroy: Just don't put loose when it's disabled in all.

rjarry: I cannot know if the admin will enable it later. Why do you care about setting it loose? It has no impact at all. It will just verify there *is* one route to reach the source address. Since grout always installs a default route via the VRF tun interface, the check passes all the time.

maxime-leroy (approved):

PR #641, opened 2026-07-10
Add MPLS support

mmuzila:


vjardin:
Please, can you explain how you would integrate it with frrouting ?

mmuzila:
> Please, can you explain how you would integrate it with frrouting ?

AFAIK the FRR zebra dplane plugin already handles IP routes, nexthops, and SRv6 bidirectionally.
MPLS would follow the same pattern:

FRR -> grout : FRR sends DPLANE_OP_LSP_INSTALL/UPDATE/DELETE when LDP, OSPF-SR,
ISIS-SR, or static MPLS configures a label switch. The plugin would extract the incoming
label and NHLFEs from the dplane context, create GR_NH_T_MPLS nexthops in grout (via
GR_NH_ADD), then bind them to the label with GR_MPLS_LABEL_ROUTE_ADD. For ECMP (multiple
NHLFEs), wrap individual MPLS nexthops in a GR_NH_T_GROUP.

grout -> FRR: Subscribe to GR_EVENT_MPLS_ROUTE_ADD/DEL in the zebra notification
handler, then call FRR's mpls_lsp_install()/mpls_lsp_uninstall() to reflect
externally-created label routes (e.g. from grcli) into FRR's LSP table.

Startup sync: After the existing nexthop/route sync passes, add a pass that streams
GR_MPLS_LABEL_ROUTE_LIST from grout and injects each entry into FRR via mpls_lsp_install().

The main complexity is that FRR's LSP NHLFEs don't carry nexthop IDs, so the plugin needs a
deterministic ID scheme (e.g. base_offset | (label << 4) | nhlfe_index) to
create/update/delete the grout nexthops. All of this goes in frr/rt_grout.c and
frr/zebra_dplane_grout.c, following the exact patterns already used for IP routes and SRv6.

modules/mpls/api/gr_mpls.h
```diff
@@ -0,0 +1,103 @@
...
+#define GR_MPLS_MAX_STACK_DEPTH 30
+
+#define GR_MPLS_LABEL_MAX 0xFFFFF
+
+// Reserved MPLS label values (RFC 3032).
+#define GR_MPLS_LABEL_IPV4_EXPLICIT_NULL 0
+#define GR_MPLS_LABEL_ROUTER_ALERT 1
+#define GR_MPLS_LABEL_IPV6_EXPLICIT_NULL 2
+#define GR_MPLS_LABEL_IMPLICIT_NULL 3
+#define GR_MPLS_LABEL_FIRST_UNRESERVED 16
```
rjarry: Does it make sense to store these in an enum?

modules/mpls/api/gr_mpls.h
```diff
@@ -0,0 +1,103 @@
...
+
+struct gr_nexthop_info_mpls {
+	uint8_t n_labels; // Number of output labels (0 = pop).
+	uint8_t ttl; // Initial TTL (0 = copy from payload).
+	addr_family_t via_af; // Gateway address family (GR_AF_IP4 or GR_AF_IP6).
+	uint8_t _pad;
+	union {
+		ip4_addr_t ipv4;
+		struct rte_ipv6_addr ipv6;
+	} via; // Gateway address for ARP/NDP resolution.
```
rjarry: You could replace all of this with:

```c
	struct l3_addr via;
```

modules/mpls/control/mpls.h
```diff
@@ -23,3 +24,5 @@ int mpls_rib_delete(uint16_t vrf_id, uint32_t label, bool missing_ok);
 
 typedef int (*mpls_rib_iter_cb)(uint16_t, uint32_t, const struct nexthop *, void *);
 int mpls_rib_iter(uint16_t vrf_id, mpls_rib_iter_cb cb, void *priv);
+
+extern control_input_t mpls_output_node;
```
rjarry: This internal identifier probably shouldn't be exposed in a header. See below...

modules/mpls/control/resolve.c
```diff
@@ -0,0 +1,122 @@
...
+
+static int mpls_resubmit_cb(struct rte_mbuf *m, struct nexthop *) {
+	l3_mbuf_data(m)->nh = mpls_hold_mbuf_data(m)->mpls_nh;
+	mbuf_data(m)->iface = NULL;
+	if (post_to_stack(mpls_output_node, m) < 0) {
+		LOG(ERR, "post_to_stack: %s", strerror(errno));
+		return -errno;
+	}
+	return 0;
+}
```
rjarry: Could you move this to `datapath/mpls_output.c` and keep the control_input_t value `static`? See ndp_na_output for reference:

https://github.com/DPDK/grout/blob/94d13676c8efcfbf21e7c6f7be6b64656453340b/modules/ip6/datapath/ndp_na_output.c#L24-L40

modules/mpls/control/resolve.c
```diff
@@ -0,0 +1,122 @@
...
+static struct nexthop_af_ops mpls_af_ops = {
+	.resolve = mpls_resolve_cb,
+	.solicit = mpls_solicit,
+	.resubmit = mpls_resubmit_cb,
+	.cleanup_routes = mpls_rib_cleanup,
+};
+
+static void mpls_resolve_init(struct event_base *) {
+	mpls_output_node = gr_control_input_register_handler("mpls_output", true);
+}
```
rjarry: This should be moved as a `.register_callback` in datapath/mpls_output.c as it is done in `ndp_na_output`:

https://github.com/DPDK/grout/blob/94d13676c8efcfbf21e7c6f7be6b64656453340b/modules/ip6/datapath/ndp_na_output.c#L114-L116

modules/mpls/datapath/mpls_input.c
```diff
@@ -0,0 +1,189 @@
...
+						mbuf->packet_type = RTE_PTYPE_L3_IPV6;
+						edge = IP6_INPUT;
+					} else {
+						edge = BAD_LABEL;
+					}
+				} else {
+					edge = BAD_LABEL;
+				}
+				break;
+			}
```
rjarry, line 92: This looks like it could be rewritten with a `switch` `case` block.

modules/mpls/datapath/mpls_input.c
```diff
@@ -0,0 +1,189 @@
...
+	rte_edge_t edge;
+
+	for (uint16_t i = 0; i < nb_objs; i++) {
+		mbuf = objs[i];
+		uint8_t depth = 0;
+		edge = BAD_LABEL;
+
+		struct rte_mpls_hdr trace_hdr = *rte_pktmbuf_mtod(mbuf, struct rte_mpls_hdr *);
+
+		while (true) {
```
rjarry: I am not fond of `while (true)` blocks in datapath nodes. Since we know there is a limit of `GR_MPLS_MAX_LABELS`, why not clamp the loop to this value so that the compiler has a chance to unroll the loop.

modules/mpls/datapath/mpls_input.c
```diff
@@ -0,0 +1,189 @@
...
+				mpls->ttl = ttl;
+				l3_mbuf_data(mbuf)->nh = nh;
+				mbuf->packet_type = RTE_PTYPE_TUNNEL_MPLS_IN_GRE;
+				edge = MPLS_OUTPUT;
+				break;
+			}
+
+			if (bos) {
+				rte_pktmbuf_adj(mbuf, sizeof(*mpls));
+				uint8_t ver = *rte_pktmbuf_mtod(mbuf, uint8_t *) >> 4;
```
rjarry: This is fragile. The MPLS FIB should contain the necessary info about what payload is next. Maybe `info->via.af`?

I don't know how other stacks do this but it seems to me we shouldn't try to guess what is next.

modules/mpls/datapath/mpls_input.c
```diff
@@ -0,0 +1,189 @@
...
+					ip = rte_pktmbuf_mtod(mbuf, struct rte_ipv4_hdr *);
+					uint16_t old_w = *(uint16_t *)&ip->time_to_live;
+					ip->time_to_live = ttl;
+					uint16_t new_w = *(uint16_t *)&ip->time_to_live;
+					uint32_t csum = (uint16_t)~ip->hdr_checksum;
+					csum += (uint16_t)~old_w + new_w;
+					csum = (csum >> 16) + (csum & 0xffff);
+					csum += csum >> 16;
+					ip->hdr_checksum = ~(uint16_t)csum;
+					mbuf->packet_type = RTE_PTYPE_L3_IPV4;
```
rjarry, line 135: Why do you need this convoluted casting? Maybe this could be simplified by reusing the existing fixup_checksum_* functions from nat_datapath.h (which should be moved into a more generic place first):

```c
struct rte_ipv4_hdr *ip;
ip = rte_pktmbuf_mtod(mbuf, struct rte_ipv4_hdr *);
ip->hdr_checksum = fixup_checksum_16(
	ip->hdr_checksum,
	rte_cpu_to_be_16(ip->time_to_live << 8),
	rte_cpu_to_be_16(ttl << 8)
);
ip->time_to_live = ttl;
```

rjarry (changes requested):

maxime-leroy:
smoke/graph_svg_test.sh -> Tests are failing because you must update the graph (make update-graph) when adding new node to the graph. https://github.com/DPDK/grout/blob/main/docs/graph.svg

PR #642, opened 2026-07-10
srv6: add per-nexthop encap source address

rjarry:
FRR 10.8 added a per-nexthop encapsulation source address to nexthop_add_srv6_seg6(). This series adds per-nexthop encap_src support to grout's SRv6 output nexthops and plumbs it through the FRR dplane plugin.

When an SRv6 output nexthop has an explicit encap_src set, the datapath uses it directly as the outer IPv6 source instead of the global tunnel source or the interface preferred address. An internal flags field avoids a 16-byte comparison in the hot path.

The series also updates FRR build support: bumps 10.5 to 10.5.4, adds 10.7 (rc1), and changes the default build version to 10.6.

## Summary

- Add per-nexthop SRv6 `encap-src` support end-to-end (CLI → control plane API structs → FRR dplane plugin → datapath encapsulation).
- Propagate an explicit `encap_src` as the outer IPv6 header source; when `encap_src` is unspecified, fall back to the configured tunnel source / interface preferred address.
- Extend SRv6 nexthop output state with an `encap_src` field plus `SR_ENCAP_F_SRC` presence flag, and in `srv6_output_process()` use the flagged `encap_src` directly to avoid doing a 16-byte IPv6 comparison in the datapath hot path.
- Update SRv6 nexthop equality/import/export logic to include `encap_src`, setting `SR_ENCAP_F_SRC` only when it’s not the unspecified IPv6 address.
- Add `rte_ipv6_addr_is_unspec()` helper in `api/gr_net_compat.h` to detect unspecified IPv6 addresses by checking all 16 bytes.
- Update CLI parsing and display for SRv6 nexthops/seglists to accept an optional `SRC` parameter and to show `ENCAP_SRC` only when `encap_src` is not unspecified.
- Extend FRR dplane SRv6 SR output nexthop handling to propagate `encap_src` to `nexthop_add_srv6_seg6()` for FRR `>= 10.8.0`, and populate the plugin-side `encap_src` from the FRR seglist source when building `GR_NH_T_SR6_OUTPUT`.
- Bump API handshake version `GR_API_VERSION` from `3` to `4`.
- Add smoke tests validating per-nexthop `encap-source` behavior: `smoke/srv6_test.sh` and `smoke/srv6_encapsrc_frr_test.sh` (skips when installed FRR < `10.8.0`).
- Update FRR build/compat support: default `frr_version` to `10.6`, add `10.7` (rc1) support, bump `10.5` wrap to `10.5.4`, and adjust CI build matrix to use `10.7` for the `gcc-14` aarch64 configuration.

### DIFF
--- a/smoke/srv6_encapsrc_frr_test.sh
+++ b/smoke/srv6_encapsrc_frr_test.sh
@@ -1,0 +1,76 @@
+#!/bin/bash
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2026 Robin Jarry
+
+set -e
+zebra=$(PATH="$1/frr_install/sbin:$1/frr_install/bin:$PATH" command -v zebra)
+frr_version=$($zebra --version | sed -En 's/zebra version //p')
+min_version=$(printf '%s\n%s\n' "$frr_version" "10.8.0" | sort -V | head -n1)
+if ! [ "$min_version" = "10.8.0" ]; then
+	echo "$0: FRR $frr_version does not support per-nexthop encap source"
+	exit 125
+fi
+
+. $(dirname $0)/_init_frr.sh
+
+create_interface p0
+create_interface p1
+
+for n in 0 1; do
+	p=x-p$n
+	ns=n$n
+	netns_add $ns
+	move_to_netns $p $ns
+done
+ip -n n0 addr add 192.168.61.2/24 dev x-p0
+ip -n n1 addr add fd00:102::2/32 dev x-p1
+
+set_ip_address p0 192.168.61.1/24
+set_ip_address p1 fd00:102::1/32
+
+#
+# network layout:
+#  (client) p0(netns) <--> p0 <grout> p1 <--->  p1(netns)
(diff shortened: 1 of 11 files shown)


smoke/srv6_encapsrc_frr_test.sh
```diff
@@ -0,0 +1,81 @@
...
+
+# underlay route
+set_ip_route fd00:202::/32 fd00:102::2
+
+# SRv6 route with explicit per-nexthop encap source via FRR
+encap_src=fd00:102::42
+mark_events
+vtysh <<EOF
+configure terminal
+ip route 192.168.0.0/16 p1 segments fd00:202:200:: encap-source $encap_src
```
maxime-leroy: set_srv6_route add an option parameter for this function please.

maxime-leroy (changes requested):

smoke/srv6_encapsrc_frr_test.sh
rjarry: done

maxime-leroy (approved):

PR #647, opened 2026-07-16
cli: fix ping response time display

rjarry:
The response_time value is now in nanoseconds.

Fixes: fe0c7f9a3654 ("api: replace clock_t with high-res type")

@MortenBroerup fyi

- Update ICMP (modules/ip/cli/icmp.c) round-trip time formatting to treat `reply_resp->response_time` as nanoseconds by computing a `double roundtrip_ms` using `GR_NS_PER_S`, then use `roundtrip_ms` in the `RTE_ICMP_TYPE_ECHO_REPLY` (traceroute and non-traceroute) and `RTE_ICMP_TYPE_TTL_EXCEEDED` output paths instead of printing `response_time / 1000.` inline.
- Update ICMPv6 (modules/ip6/cli/icmp6.c) traceroute timing to precompute a `roundtrip_ms` from nanosecond `reply_resp->response_time` using `GR_NS_PER_S`, and pass `roundtrip_ms` into the traceroute echo-reply and traceroute error/timeout `printf` formatting rather than recomputing the ms conversion inline.

### DIFF
--- a/modules/ip/cli/icmp.c
+++ b/modules/ip/cli/icmp.c
@@ -85,13 +85,14 @@ static cmd_status_t icmp_send(
 			else
 				printf("timeout: icmp_seq=%d\n", i);
 		} else {
+			double roundtrip_ms = (reply_resp->response_time * 1000.) / GR_NS_PER_S;
 			switch (reply_resp->type) {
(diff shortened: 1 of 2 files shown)


MortenBroerup (approved):
Improves readability, yes.

PR #649, opened 2026-07-20
rpm: fix build without systemd

rjarry:
Fix the following error when building `--without=systemd`:

```
Processing files: grout-frr-0.16.1-20260720120706.local.el10.x86_64
error: File not found: /BUILDROOT/.../usr/lib/systemd/system/frr.service.d/grout.conf
```

Fixes: a6643d4302f5 ("frr: add systemd dropin to rpm/deb installed files")

Conditionally includes `%{_unitdir}/frr.service.d/grout.conf` in the RPM `%files frr` section of `rpm/grout.spec` only when `%{with systemd}` is enabled, fixing RPM builds with `--without=systemd` by avoiding the missing drop-in file.


zeeke (approved):
LGTM

PR #650, opened 2026-07-20
l2: push bridge mac address to members

david-marchand:
Similar to bonding, underlying iface members may need to know about this bridge mac addresses.

- Adds bulk helpers in `modules/l2/control/bridge.c` to add/delete a specified Ethernet MAC across all bridge member interfaces (`bridge_all_member_add_mac` / `bridge_all_member_del_mac`), and wires them into new bridge iface callbacks (`iface_type_bridge.{add_eth_addr,del_eth_addr}` mapped to `bridge_mac_add` / `bridge_mac_del`).
- On member attach/detach and bridge finalization, synchronizes member interface MAC lists with the bridge by adding/removing both the bridge’s existing per-MAC entries (`bridge->macs`) and the bridge primary MAC (`br->mac`) on the affected member(s).
- Reworks `bridge_set_eth_addr`: generates a random MAC when the input is zero, no-ops when the requested MAC already matches `bridge->mac`, otherwise deletes the previous `bridge->mac` from all members (when non-zero), adds the new MAC to all members, and updates `bridge->mac` (propagating errors from the bulk add/delete helpers).

### DIFF
--- a/modules/l2/control/bridge.c
+++ b/modules/l2/control/bridge.c
@@ -10,6 +10,32 @@
 
 LOG_TYPE("bridge");
 
+static int
+bridge_all_member_add_mac(const struct iface_info_bridge *br, const struct rte_ether_addr *mac) {
+	for (uint16_t i = 0; i < br->n_members; i++) {
+		if (iface_add_eth_addr(br->members[i], mac) < 0)
+			return errno_set(errno);
+	}
+	return 0;
+}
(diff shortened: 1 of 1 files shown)


rjarry (commented):
Too bad we cannot rely on promisc for this :trollface:

rjarry (approved):

PR #652, opened 2026-07-21
fdb: also program mac in hw when an entry is updated

rjarry:
When an FDB entry is create via regular local mac learning in a VXLAN port (via flood vteps), the associated address is *not* installed in the hardware filters because the entry does not have GR_FDB_F_EXTERN.

Later, when an EVPN type 2 route is exchanged via BGP, the FDB entry is *updated* (not created) and the GR_FDB_F_EXTERN flag is added.

But since fdb_event_cb only fires on creation and deletion, the hardware filter is never updated. Subscribe to GR_EVENT_FDB_UPDATE to fix this.

Fixes: 10be5b5883ce ("l2: push external mac addresses in hardware")

Adds `GR_FDB_F_HW` to `gr_fdb_flags_t` to mark MAC entries programmed into hardware. Extends `fdb_match` so flush/list requests can filter on `GR_FDB_F_HW`, and updates `fdb_event_cb` to subscribe to `GR_EVENT_FDB_UPDATE` and apply the same add/delete hardware programming path as existing events: it derives whether the update is an add or delete, early-returns when the entry’s current `GR_FDB_F_HW` already matches the requested transition, programs the MAC to bridge members when needed, and sets/clears `GR_FDB_F_HW` after the corresponding hardware add/delete. Also updates `modules/l2/cli/fdb.c` to render the `hw` flag and accept `fdb show hw` filtering based on `GR_FDB_F_HW`.

### DIFF
--- a/modules/l2/control/fdb.c
+++ b/modules/l2/control/fdb.c
@@ -256,6 +256,8 @@ static inline bool fdb_match(
 		return false;
 	if ((flags & GR_FDB_F_EXTERN) && !(e->flags & GR_FDB_F_EXTERN))
 		return false;
+	if ((flags & GR_FDB_F_HW) && !(e->flags & GR_FDB_F_HW))
+		return false;
 	if (bridge_id != GR_IFACE_ID_UNDEF && e->bridge_id != bridge_id)
 		return false;
 	if (iface_id != GR_IFACE_ID_UNDEF && e->iface_id != iface_id)
@@ -352,7 +354,8 @@ static void push_mac_to_hw(struct iface *iface, const struct rte_ether_addr *mac
 
 static void fdb_event_cb(uint32_t event, const void *obj) {
 	const struct iface_info_bridge *bridge_info;
-	const struct gr_fdb_entry *fdb = obj;
+	struct gr_fdb_entry *fdb = (void *)obj;
+	bool add = event != GR_EVENT_FDB_DEL;
 	const struct iface *bridge;
 
 	bridge = iface_from_id(fdb->bridge_id);
@@ -363,6 +366,10 @@ static void fdb_event_cb(uint32_t event, const void *obj) {
 
 	if ((fdb->flags & GR_FDB_F_EXTERN) == 0)
 		return;
+	if (add && (fdb->flags & GR_FDB_F_HW))
+		return;
+	if (!add && !(fdb->flags & GR_FDB_F_HW))
+		return;
 
 	bridge_info = iface_info_bridge(bridge);
 	for (unsigned i = 0; i < bridge_info->n_members; i++) {
@@ -370,11 +377,13 @@ static void fdb_event_cb(uint32_t event, const void *obj) {
 
 		// we have no clear idea what to do with a vlan_id if one got pushed by FRR
 		assert(fdb->vlan_id == 0);
-		// FIXME: ideally, we should ask for adding the mac in hw only when getting the
-		// *first* transition from !EXTERN to EXTERN to handle both learning mac on the
-		// wire and from FRR.
-		push_mac_to_hw(member, &fdb->mac, event != GR_EVENT_FDB_DEL);
+		push_mac_to_hw(member, &fdb->mac, add);
 	}
+
+	if (add)
+		fdb->flags |= GR_FDB_F_HW;
+	else
+		fdb->flags &= ~GR_FDB_F_HW;
 }
 
 void fdb_sync_hardware(const struct iface *bridge, struct iface *member, bool add) {
@@ -472,6 +481,7 @@ RTE_INIT(init) {
 	api_handler(GR_FDB_CONFIG_GET, fdb_config_get);
 	api_handler(GR_FDB_CONFIG_SET, fdb_config_set);
 	event_subscribe(GR_EVENT_FDB_ADD, fdb_event_cb);
+	event_subscribe(GR_EVENT_FDB_UPDATE, fdb_event_cb);
 	event_subscribe(GR_EVENT_FDB_DEL, fdb_event_cb);
 	event_serializer(GR_EVENT_FDB_ADD, NULL);
 	event_serializer(GR_EVENT_FDB_DEL, NULL);
--- a/modules/l2/api/gr_l2.h
+++ b/modules/l2/api/gr_l2.h
@@ -65,6 +65,7 @@ typedef enum : uint8_t {
 	GR_FDB_F_STATIC = GR_BIT8(0), // User-configured, never aged out.
 	GR_FDB_F_LEARN = GR_BIT8(1), // Learned via local bridge.
 	GR_FDB_F_EXTERN = GR_BIT8(2), // Programmed by external control plane.
+	GR_FDB_F_HW = GR_BIT8(3), // Mac address programmed in hardware filter.
 } gr_fdb_flags_t;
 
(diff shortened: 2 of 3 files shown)


david-marchand:
I have one concern with this change.

Any update on a fdb entry will end up as an addition in the underlying ifaces right?
Won't it mess up the refcounting?

rjarry:
> Any update on a fdb entry will end up as an addition in the underlying ifaces right? Won't it mess up the refcounting?

Maybe we could track if the mac was installed in HW in an internal flag of the FDB entry and only install it if it wasn't done already? Same thing at deletion?

modules/l2/control/fdb.c
```diff
@@ -333,20 +333,27 @@ static struct api_out fdb_config_set(const void *request, struct api_ctx *) {
...
+	int ret = 0;
+
+	// we have no clear idea what to do with a vlan_id if one got pushed by FRR
+	assert(fdb->vlan_id == 0);
 
-	if (add)
-		ret = iface_add_eth_addr(iface, mac);
-	else
-		ret = iface_del_eth_addr(iface, mac);
+	if (add && !(fdb->flags & GR_FDB_F_HW))
```
david-marchand: If we have multiple members, on the first addition, the fdb entry will get flagged as GR_FDB_F_HW.
Then on the second member, the if (add && !(fdb->flags & GR_FDB_F_HW)) is false, and the code will try to remove the mac address from this second member.

modules/l2/control/fdb.c
```diff
@@ -333,20 +333,27 @@ static struct api_out fdb_config_set(const void *request, struct api_ctx *) {
...
-		LOG(DEBUG,
+		LOG(NOTICE,
 		    "failed to %s mac " ETH_F " to %s: %s",
 		    add ? "add" : "del",
-		    mac,
+		    &fdb->mac,
 		    iface->name,
 		    strerror(errno));
+	} else if (add) {
+		((struct gr_fdb_entry *)fdb)->flags |= GR_FDB_F_HW;
```
david-marchand: A bit sad that we need to cast... though I have no better idea.

modules/l2/control/fdb.c
rjarry: Ah right. I had missed this. We need a flag "per member" and "per fdb entry".

modules/l2/control/fdb.c
david-marchand: This flag update is not atomic. No race condition possible?

rjarry: Since this runs in the control plane thread, I don't think there is any race possible.

modules/l2/control/fdb.c
david-marchand: The GR_FDB_F_HW considerations should probably be moved one level upper, in fdb_event_cb.
And leave fdb_sync_hardware() untouched.

david-marchand: > Ah right. I had missed this. We need a flag "per member" and "per fdb entry".

I don't think we need per member.

Failing to add to some member should not prevent from adding in other member.
We just want to detect a transition !EXTERN -> EXTERN.

rjarry: > I don't think we need per member.
> 
> Failing to add to some member should not prevent from adding in other member. We just want to detect a transition !EXTERN -> EXTERN.

But you need a way to avoid adding the mac twice on the same member, otherwise the ref counting is messed up.

david-marchand: So in the end, something like:

```
diff --git a/modules/l2/control/fdb.c b/modules/l2/control/fdb.c
index 823e8c81..16108bc1 100644
--- a/modules/l2/control/fdb.c
+++ b/modules/l2/control/fdb.c
@@ -351,9 +351,10 @@ static void push_mac_to_hw(struct iface *iface, const struct rte_ether_addr *mac
 }
 
 static void fdb_event_cb(uint32_t event, const void *obj) {
+       struct gr_fdb_entry *fdb = (struct gr_fdb_entry *)obj;
        const struct iface_info_bridge *bridge_info;
-       const struct gr_fdb_entry *fdb = obj;
        const struct iface *bridge;
+       bool add;
 
        bridge = iface_from_id(fdb->bridge_id);
        if (bridge == NULL) {
@@ -364,17 +365,25 @@ static void fdb_event_cb(uint32_t event, const void *obj) {
        if ((fdb->flags & GR_FDB_F_EXTERN) == 0)
                return;
 
+       add = (event != GR_EVENT_FDB_DEL);
+       if (add && (fdb->flags & GR_FDB_F_HW))
+               return;
+       if (!add && !(fdb->flags & GR_FDB_F_HW))
+               return;
+
        bridge_info = iface_info_bridge(bridge);
        for (unsigned i = 0; i < bridge_info->n_members; i++) {
                struct iface *member = bridge_info->members[i];
 
                // we have no clear idea what to do with a vlan_id if one got pushed by FRR
                assert(fdb->vlan_id == 0);
-               // FIXME: ideally, we should ask for adding the mac in hw only when getting the
-               // *first* transition from !EXTERN to EXTERN to handle both learning mac on the
-               // wire and from FRR.
-               push_mac_to_hw(member, &fdb->mac, event != GR_EVENT_FDB_DEL);
+               push_mac_to_hw(member, &fdb->mac, add);
        }
+
+       if (add)
+               fdb->flags |= GR_FDB_F_HW;
+       else
+               fdb->flags &= ~GR_FDB_F_HW;
 }
 
 void fdb_sync_hardware(const struct iface *bridge, struct iface *member, bool add) {
@@ -472,6 +481,7 @@ RTE_INIT(init) {
        api_handler(GR_FDB_CONFIG_GET, fdb_config_get);
        api_handler(GR_FDB_CONFIG_SET, fdb_config_set);
        event_subscribe(GR_EVENT_FDB_ADD, fdb_event_cb);
+       event_subscribe(GR_EVENT_FDB_UPDATE, fdb_event_cb);
        event_subscribe(GR_EVENT_FDB_DEL, fdb_event_cb);
        event_serializer(GR_EVENT_FDB_ADD, NULL);
        event_serializer(GR_EVENT_FDB_DEL, NULL);

```

rjarry: But how do you deal if new ports are added to the bridge?

david-marchand: bridge_detach_member/bridge_attach_member call fdb_sync_hardware which only pushes EXTERN entries.

david-marchand: > > I don't think we need per member.
> > Failing to add to some member should not prevent from adding in other member. We just want to detect a transition !EXTERN -> EXTERN.
> 
> But you need a way to avoid adding the mac twice on the same member, otherwise the ref counting is messed up.

In which case would the fdb add the mac twice?
This would be a bug in the bridge code.

modules/l2/control/fdb.c
david-marchand: If you are sure of it, ok for me.

modules/l2/control/fdb.c
rjarry: I have implemented your proposal. Seems correct.

rjarry:
@david-marchand what do you think? The FRR master tests are broken but it has nothing to do with the FDB change (see #651).

david-marchand (approved):
Maybe highlight in the title that it is a fix.
But otherwise lgtm.

PR #654, opened 2026-07-21
readme: update build dependencies

rjarry:
The build dependencies section didn't mention libecoli indirect dependency to libedit nor any of the FRR build dependencies which are now "required" by default (since FRR is now enabled by default since commit ab2ef19fca14 ("frr: always build from subproject when system headers are missing").

@MortenBroerup fyi

## Summary

- Updated README build instructions with separate DNF/RPM and APT/DEB dependency groups.
- Documented the indirect `libecoli` → `libedit` dependency using conditional package-install fallbacks.
- Added FRR build dependencies required when FRR builds from the subproject.
- Marked FRR as enabled by default when compiled with `-Dfrr=enabled`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,19 +264,43 @@ image.
 ### Install build dependencies
 
 ```sh
+# grout direct build dependencies
 dnf install git gcc make meson ninja-build pkgconf \
         python3-pyelftools scdoc libmnl-devel \
-        libcmocka-devel libedit-devel libevent-devel numactl-devel \
+        libcmocka-devel libevent-devel numactl-devel \
         libarchive-devel rdma-core-devel
+
+# grcli build dependencies
+if ! dnf install libecoli-devel; then
+        # libecoli is not available, install its dependencies to build from source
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -264,19 +264,37 @@ image.
...
 
 ```sh
+# grout direct build dependencies
 dnf install git gcc make meson ninja-build pkgconf \
         python3-pyelftools scdoc libmnl-devel \
-        libcmocka-devel libedit-devel libevent-devel numactl-devel \
+        libcmocka-devel libevent-devel numactl-devel \
         libarchive-devel rdma-core-devel
+
+# libecoli build dependencies
```
MortenBroerup: Maybe mention that libecoli is a CLI build dependency, since the "root cause" of the libedit-devel dependency is the CLI. (Same below in line 291.)

rjarry: Let me know if this is better.

MortenBroerup (approved):
LGTM now. Thanks.

PR #655, opened 2026-07-23
l2: do not push locally learnt mac to hardware

david-marchand:
Avoid pushing back mac addresses coming from the same iface. This is unneeded and may confuse some NICs.

With this approach, pushing mac to hardware becomes way simpler: a decapsulated packet mac address coming from a VXLAN iface simply has to be pushed to all members of the bridge, when it is first learnt.

Similarly, a mac address learnt by external means (FRR) can be pushed to all but the originating interface.

If we had a setup with a bridge with two member ifaces backed by two separate NICs, a mac address learnt from an interface belonging to the first NIC will be pushed to the interface backed by the second NIC.

As a consequence of this approach, we don't need to listen to FDB updates. The GR_FDB_F_HW flag becomes unneeded since only add/del ops matter.

Fixes: a4c609a5a5ac ("l2: program all fdb entries in bridge ports")
Fixes: 3f6447663762 ("fdb: fix missing mac in hw when an entry is updated")
Fixes: 10be5b5883ce ("l2: push external mac addresses in hardware")

## Summary

- Removed `GR_FDB_F_HW` from `gr_fdb_flags_t` (`modules/l2/api/gr_l2.h`) and updated CLI/view handling so `fdb_format_flags()` and `fdb show` no longer format or parse an `"hw"` flag (`modules/l2/cli/fdb.c`).
- Reworked L2 FDB control-plane state in `modules/l2/control/fdb.c` by wrapping `gr_fdb_entry` in an internal `struct fdb_entry` that tracks `prev_iface_id` for learned MACs whose learned interface (`iface_id`) changes.
- Updated `fdb_event_cb` to treat events as `const struct fdb_entry *`, remove all `GR_FDB_F_HW`-based gating, and program bridge members by skipping the learned/originating interface (`member->id == fdb->iface_id`).
  - For `GR_EVENT_FDB_ADD`/`GR_EVENT_FDB_DEL`: add/delete the MAC across all bridge members except the learned/originating interface.
  - For `GR_EVENT_FDB_UPDATE`: use `prev_iface_id` to delete from the newly learned interface and add to the previously learned interface, reflecting a MAC move between interfaces.
- Updated `fdb_match()` to no longer filter based on `GR_FDB_F_HW`, and updated `fdb_sync_hardware()` to no longer depend on `GR_FDB_F_HW` state and to also skip pushing updates to the learned/originating interface.
- Adjusted event subscription ordering at `RTE_INIT` so `GR_EVENT_FDB_DEL` is subscribed before `GR_EVENT_FDB_UPDATE`.
- Bumped `GR_API_VERSION` from `4` to `5` in `api/gr_api.h`.

### DIFF
--- a/modules/l2/control/fdb.c
+++ b/modules/l2/control/fdb.c
@@ -21,6 +21,11 @@ struct fdb_key {
 	struct rte_ether_addr mac;
 };
 
+struct fdb_entry {
+	BASE(gr_fdb_entry);
+	uint16_t prev_iface_id;
+};
+
 static unsigned fdb_max_entries;
 static struct rte_hash *fdb_hash;
 static struct rte_mempool *fdb_pool;
@@ -50,7 +55,7 @@ static int fdb_reconfig(unsigned max_entries) {
 	struct rte_mempool *p = rte_mempool_create(
 		name,
 		rte_align32pow2(max_entries) - 1,
-		sizeof(struct gr_fdb_entry),
+		sizeof(struct fdb_entry),
 		0, // cache size
 		0, // priv size
 		NULL, // mp_init
@@ -112,7 +117,7 @@ void fdb_learn(
 	const struct l3_addr *vtep
 ) {
 	const struct fdb_key key = {bridge_id, vlan_id, *mac};
-	struct gr_fdb_entry *fdb;
+	struct fdb_entry *fdb;
 	void *data;
 
 	if (rte_hash_lookup_data(fdb_hash, &key, &data) < 0) {
@@ -120,6 +125,7 @@ void fdb_learn(
 			return; // pool exhausted
 
 		fdb = data;
+		fdb->prev_iface_id = GR_IFACE_ID_UNDEF;
 		fdb->bridge_id = bridge_id;
 		fdb->vlan_id = vlan_id;
 		fdb->mac = *mac;
@@ -143,6 +149,7 @@ void fdb_learn(
 	if ((fdb->flags & GR_FDB_F_LEARN)
 	    && (fdb->iface_id != iface_id || !l3_addr_eq(&fdb->vtep, vtep))) {
 		// update in case the mac address has moved
+		fdb->prev_iface_id = fdb->iface_id;
 		fdb->iface_id = iface_id;
 		fdb->vtep = *vtep;
 		event_push(GR_EVENT_FDB_UPDATE, fdb);
@@ -150,7 +157,7 @@ void fdb_learn(
 }
 
 void fdb_purge_iface(uint16_t iface_id) {
-	struct gr_fdb_entry *fdb;
+	struct fdb_entry *fdb;
 	uint32_t next = 0;
 	const void *key;
 	void *data;
@@ -160,11 +167,14 @@ void fdb_purge_iface(uint16_t iface_id) {
 		if (fdb->iface_id == iface_id) {
 			rte_hash_del_key(fdb_hash, key);
 		}
+		if (fdb->prev_iface_id == iface_id) {
+			fdb->prev_iface_id = GR_IFACE_ID_UNDEF;
+		}
 	}
 }
 
 void fdb_purge_bridge(uint16_t bridge_id) {
-	struct gr_fdb_entry *fdb;
+	struct fdb_entry *fdb;
 	uint32_t next = 0;
 	const void *key;
 	void *data;
@@ -180,7 +190,7 @@ void fdb_purge_bridge(uint16_t bridge_id) {
 static struct api_out fdb_add(const void *request, struct api_ctx *) {
 	const struct gr_fdb_add_req *req = request;
 	const struct iface *iface;
-	struct gr_fdb_entry *e;
+	struct fdb_entry *e;
 	void *data;
 	int ret;
 
@@ -205,7 +215,8 @@ static struct api_out fdb_add(const void *request, struct api_ctx *) {
 			return api_out(-ret, 0, NULL);
 
 		e = data;
-		*e = req->fdb;
+		e->prev_iface_id = GR_IFACE_ID_UNDEF;
+		e->base = req->fdb;
 		e->bridge_id = iface->id;
 		e->last_seen = gr_clock_ns();
 
@@ -217,7 +228,8 @@ static struct api_out fdb_add(const void *request, struct api_ctx *) {
 		event_push(GR_EVENT_FDB_ADD, e);
 	} else if (req->exist_ok) {
 		e = data;
-		*e = req->fdb;
+		e->prev_iface_id = e->iface_id;
+		e->base = req->fdb;
 		e->bridge_id = iface->id;
 		e->last_seen = gr_clock_ns();
 
@@ -286,7 +298,7 @@ static struct api_out fdb_flush(const void *request, struct api_ctx *) {
 
 static struct api_out fdb_list(const void *request, struct api_ctx *ctx) {
 	const struct gr_fdb_list_req *req = request;
-	struct gr_fdb_entry *fdb;
+	struct fdb_entry *fdb;
 	uint32_t next = 0;
 	const void *key;
 	void *data;
@@ -296,7 +308,7 @@ static struct api_out fdb_list(const void *request, struct api_ctx *ctx) {
 			continue;
 
 		fdb = data;
-		api_send(ctx, sizeof(*fdb), fdb);
+		api_send(ctx, sizeof(fdb->base), fdb);
 	}
 
 	return api_out(0, 0, NULL);
@@ -352,8 +364,9 @@ static void push_mac_to_hw(struct iface *iface, const struct rte_ether_addr *mac
 
 static void fdb_event_cb(uint32_t event, const void *obj) {
 	const struct iface_info_bridge *bridge_info;
-	const struct gr_fdb_entry *fdb = obj;
+	const struct fdb_entry *fdb = obj;
 	const struct iface *bridge;
+	struct iface *member;
 
 	bridge = iface_from_id(fdb->bridge_id);
 	if (bridge == NULL) {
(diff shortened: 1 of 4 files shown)


modules/l2/control/fdb.c
```diff
@@ -364,24 +361,18 @@ static void fdb_event_cb(uint32_t event, const void *obj) {
...
 		struct iface *member = bridge_info->members[i];
 
+		// skip the interface where the MAC was learned
+		if (member->id == fdb->iface_id)
+			continue;
+
 		// we have no clear idea what to do with a vlan_id if one got pushed by FRR
 		assert(fdb->vlan_id == 0);
-		push_mac_to_hw(member, &fdb->mac, add);
+		push_mac_to_hw(member, &fdb->mac, event != GR_EVENT_FDB_DEL);
```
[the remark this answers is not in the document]
david-marchand: The commitlog may not be accurate enough. VXLAN iface must be skipped too...

david-marchand:
Coderabbit had one interesting comment I can't find here...
So copying from mail:

477-478: 🎯 Functional Correctness | 🟠 Major | ⚡ Quick win

Preserve hardware filters when a learned MAC moves.

fdb_learn() still emits GR_EVENT_FDB_UPDATE after replacing iface_id. With no update subscriber, a MAC moving from member A to B remains installed on B and is never installed on A; a later delete also skips B, leaving the stale filter. Propagate the previous interface in the move event and reconcile both members, or otherwise perform that delta before overwriting the entry.
🤖 Prompt for AI Agents

Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@modules/l2/control/fdb.c` around lines 477 - 478, Update fdb_learn() and the
GR_EVENT_FDB_UPDATE handling around fdb_event_cb so a learned MAC move preserves
the previous interface identity, reconciles hardware filters on both the old and
new members, and ensures later deletes remove any stale filter; do not overwrite
iface_id before the move delta is processed.

modules/l2/control/fdb.c
```diff
@@ -391,9 +382,13 @@ void fdb_sync_hardware(const struct iface *bridge, struct iface *member, bool ad
 	void *data;
 
 	while (rte_hash_iterate(fdb_hash, &key, &data, &next) >= 0) {
+		if (!fdb_match(data, GR_FDB_F_EXTERN, bridge->id, GR_IFACE_ID_UNDEF, NULL))
+			continue;
+
```
rjarry: This should be kept out. We want to sync all entries, not only GR_FDB_F_EXTERN.

modules/l2/control/fdb.c
```diff
@@ -391,9 +382,13 @@ void fdb_sync_hardware(const struct iface *bridge, struct iface *member, bool ad
...
 		fdb = data;
 
-		if (fdb->bridge_id != bridge->id)
+		// skip the interface where the MAC was learned
+		if (member->id == fdb->iface_id)
```
rjarry, line 417: And change to:

```c
		if (fdb->bridge_id != bridge->id || member->id == fdb->iface_id)
```

rjarry (changes requested):
About coderabbit's feedback, indeed, mac addresses can move in between interfaces. I don't know if we can do something in the current state of things since we don't know what was the previous interface ID to be able to remove it from the hardware filter.

modules/l2/control/fdb.c
david-marchand: Gah, I had it correctly before "cleaning up"...

Can't we use fdb_match(data, 0, bridge->id, GR_IFACE_ID_UNDEF, NULL) ?

david-marchand:
Yes, we would need the old iface... I see the fdb object has enough space in the current hole, would it be acceptable to track the prev_iface_id?

rjarry:
> Yes, we would need the old iface... I see the fdb object has enough space in the current hole, would it be acceptable to track the prev_iface_id?

I don't mind adding a field, but the problem is it will be exposed in the API since it is a public struct. This field may contain garbage most of the time it would only make sense internally.

modules/l2/control/fdb.c
rjarry: We could, but it would be simpler to just to do:

```c
if (fdb->bridge_id != bridge->id || member->id == fdb->iface_id)
        continue;
```

david-marchand:
Yes, exposing internal stuff is not elegant / a risk someone relies on it.
There is also an open question (I don't think we flush fdb in case an iface disappears, do we?) about iface_id validity in case a port gets deleted... and prev_iface_id would be a concern too.

For this current issue with FDB, I see two other options:
- allow extra data to be passed along the event,
- remove the UPDATE event and emits two separate ADD/DEL events,

modules/l2/control/fdb.c
david-marchand: This form aggregates two checks together, kind of hiding the intent, between filtering fdb from this bridge, and not pushing on the iface the FDB comes from.

I would at least have two blocks:

```c
                if (fdb->bridge_id != bridge->id)
                        continue;
                // skip the interface where the MAC was learned
                if (member->id == fdb->iface_id)
                        continue;
```

david-marchand:
None of those two options look good to me either...
- the notification mechanism would need a lot of changes,
- we would need two fdb objects, since handling of the del and add events could happen at different point in time,

I'll leave this topic aside for this PR, the problem about mac move already exists even before the changes.

rjarry:
There is a third solution: wrap `struct gr_fdb_entry` into an internal version that holds private data.

```c
struct fdb_entry {
	BASE(gr_fdb_entry);
	uint16_t prev_iface_id;
};
```

And then, maintain that value when updating existing entries, e.g.:

```diff
@@ -143,6 +149,7 @@ void fdb_learn(
        if ((fdb->flags & GR_FDB_F_LEARN)
            && (fdb->iface_id != iface_id || !l3_addr_eq(&fdb->vtep, vtep))) {
                // update in case the mac address has moved
+               fdb->prev_iface_id = fdb->iface_id;
                fdb->iface_id = iface_id;
                fdb->vtep = *vtep;
                event_push(GR_EVENT_FDB_UPDATE, fdb);
```

Finally, you can use this to update the mac filters:

```c
static void fdb_event_cb(uint32_t event, const void *obj) {
	const struct iface_info_bridge *bridge_info;
	const struct fdb_entry *fdb = obj;
	const struct iface *bridge;

	bridge = iface_from_id(fdb->bridge_id);
	bridge_info = iface_info_bridge(bridge);

	if (event == GR_EVENT_FDB_UPDATE) {
		if (fdb->iface_id == fdb->prev_iface_id)
			return;
		for (unsigned i = 0; i < bridge_info->n_members; i++) {
			struct iface *member = bridge_info->members[i];

			if (member->id == fdb->prev_iface_id)
				push_mac_to_hw(member, &fdb->mac, true);
			else if (member->id == fdb->iface_id)
				push_mac_to_hw(member, &fdb->mac, false);
		}
		return;
	}

	for (unsigned i = 0; i < bridge_info->n_members; i++) {
		struct iface *member = bridge_info->members[i];

		// skip the interface where the MAC was learned
		if (member->id == fdb->iface_id)
			continue;

		push_mac_to_hw(member, &fdb->mac, event != GR_EVENT_FDB_DEL);
	}
}
```

david-marchand:
> There is a third solution: wrap `struct gr_fdb_entry` into an internal version that holds private data.
> 
> ```c
> struct fdb_entry {
> 	BASE(gr_fdb_entry);
> 	uint16_t prev_iface_id;
> };
> ```

Yes, interesting.
I'll have a try.


> 
> And then, maintain that value when updating existing entries, e.g.:
> 
> ```diff
> @@ -143,6 +149,7 @@ void fdb_learn(
>         if ((fdb->flags & GR_FDB_F_LEARN)
>             && (fdb->iface_id != iface_id || !l3_addr_eq(&fdb->vtep, vtep))) {
>                 // update in case the mac address has moved
> +               fdb->prev_iface_id = fdb->iface_id;
>                 fdb->iface_id = iface_id;
>                 fdb->vtep = *vtep;
>                 event_push(GR_EVENT_FDB_UPDATE, fdb);
> ```
> 
> Finally, you can use this to update the mac filters:
> 
> ```c
> static void fdb_event_cb(uint32_t event, const void *obj) {
> 	const struct iface_info_bridge *bridge_info;
> 	const struct fdb_entry *fdb = obj;
> 	const struct iface *bridge;
> 
> 	bridge = iface_from_id(fdb->bridge_id);
> 	bridge_info = iface_info_bridge(bridge);
> 
> 	if (event == GR_EVENT_FDB_UPDATE) {
> 		if (fdb->iface_id == fdb->prev_iface_id)
> 			return;
> 		for (unsigned i = 0; i < bridge_info->n_members; i++) {
> 			struct iface *member = bridge_info->members[i];
> 
> 			if (member->id == fdb->prev_iface_id)
> 				push_mac_to_hw(member, &fdb->mac, true);
> 			else if (member->id == fdb->iface_id)
> 				push_mac_to_hw(member, &fdb->mac, false);
> 		}
> 		return;

I had an implementation with a iface lookup on prev and new iface.
What is more costly?

rjarry:
> I had an implementation with a iface lookup on prev and new iface. What is more costly?

The difference is probably negligible. Lookup on prev and new will be cleaner.

david-marchand:
Just a warning, I did not take time to test the new code.

rjarry (approved):
This looks correct to me.

david-marchand:
I tested those two patches on a simple setup with vxlan (mlx5 VF), this looks ok.
However, I can't think of an easy way to test mac movement between ports in the setup I have.

rjarry:
> I tested those two patches on a simple setup with vxlan (mlx5 VF), this looks ok. However, I can't think of an easy way to test mac movement between ports in the setup I have.

Thanks for testing! I'll merge it as is. I don't foresee any problem with the mac movement code.

PR #658, opened 2026-07-23
ip6: punt received router advertisements to control plane

### LINKED ISSUE #657: Received IPv6 Router Advertisements are dropped, breaking FRR BGP unnumbered over grout ports
## Summary

grout's datapath drops all received ICMPv6 Router Advertisements (type 134) as
"unsupported", and never punts them to the per-interface control-plane TAP. Because
FRR's BGP unnumbered (`neighbor <iface> interface remote-as ...`) relies on receiving
the peer's Router Advertisement to auto-discover the peer's link-local next-hop, BGP
sessions over grout ports stay stuck in `Idle` forever. Numbered eBGP over the same
port works fine, which isolates the problem to RA handling specifically.

## Details / root cause

In `modules/ip6/datapath/icmp6_input.c`, Router Advertisements fall into the
`default`/`ICMP6_TYPE_ROUTER_ADVERT` branch, which only punts to the control plane if a
callback has been registered for that ICMPv6 type — otherwise the packet is sent to the
`icmp6_input_unsupported` DROP node:

    case ICMP6_TYPE_ROUTER_ADVERT:
    default:
        if (icmp6_cb[icmp6->type] != NULL) {
            control_output_set_cb(mbuf, icmp6_cb[icmp6->type], gr_clock_ns());
            next = CONTROL;
        } else {
            next = UNSUPPORTED;          // -> icmp6_input_unsupported (GR_DROP_REGISTER)
        }

The only registered ICMPv6 callbacks are ECHO_REPLY and the four error types
(`modules/ip6/control/icmp6.c`):

    icmp6_input_register_callback(ICMP6_TYPE_ECHO_REPLY, icmp6_input_cb);
    icmp6_input_register_callback(ICMP6_ERR_DEST_UNREACH, icmp6_input_cb);
    icmp6_input_register_callback(ICMP6_ERR_TTL_EXCEEDED, icmp6_input_cb);
    icmp6_input_register_callback(ICMP6_ERR_PKT_TOO_BIG, icmp6_input_cb);
    icmp6_input_register_callback(ICMP6_ERR_PARAM_PROBLEM, icmp6_input_cb);

There is no `ICMP6_TYPE_ROUTER_ADVERT` callback anywhere, and `router_advert.c` is
send-only (it generates grout's own RAs; there is no RA receiver). So received RAs are
always dropped and are never delivered to the kernel control-plane TAP where FRR listens.

By contrast, local TCP (e.g. BGP :179) is correctly punted to FRR:
`modules/l4/l4_input_local.c` sends non-UDP local traffic to the `MANAGEMENT` edge
(`l4_loopback_output` -> `iface_cp_tx` -> the interface `cp_fd` TAP). ICMPv6 is
registered as its own local proto handler (`ip6_input_local_add_proto(IPPROTO_ICMPV6,
"icmp6_input")`), so RAs are intercepted before ever reaching that generic punt path.

## Steps to reproduce

1. Create a grout net_tap port and connect it (L2) to an FRR/Linux peer that sends
   IPv6 Router Advertisements on the link (e.g. an FRR leaf using `neighbor <iface>
   interface remote-as external`, which enables RA on the interface).
2. Run FRR with the grout dataplane plugin inside grout's netns and configure BGP
   unnumbered over the grout port:
       router bgp 64497
        neighbor uplink1 interface remote-as external
3. Observe: `vtysh -c "show bgp summary"` shows the neighbor stuck in `Idle` (never up).
4. `grcli graph stats` shows a growing `icmp6_input_unsupported` counter and no
   corresponding control-plane RX (`iface_cp_rx`) for the RAs. tcpdump on the peer
   confirms RAs are being sent toward grout, but FRR never learns the peer link-local.
5. Switch the same link to numbered p2p (assign IPs to the grout port and the peer)
   and the BGP session comes up immediately — confirming only RA delivery is broken.

## Expected behaviour

Received Router Advertisements on a grout port should be punted to that interface's
control-plane TAP (like other local control traffic) so a userspace routing daemon
(FRR/BIRD) can consume them, enabling BGP unnumbered / SLAAC-style next-hop discovery.
At minimum, a way to register an RA punt callback (or a per-interface "punt RA to control
plane" knob) would unblock this.

## Environment
- FRR: BGP unnumbered eBGP over the grout uplink port.

### Version

grout:edge (main / DPDK 25.11). Also reproduced against grout 0.16.2.

chdxD1:
### Problem

BGP unnumbered peering (RFC 5549) relies on IPv6 Router Advertisements to
discover the peer's link-local next hop. grout was dropping all received RAs
in `icmp6_input` (dispatched to the `unsupported` edge), so a routing daemon
such as FRR running on top of grout would never receive them. The BGP session
would remain stuck in `Idle (Waiting for peer address)` indefinitely.

A second issue caused zebra to crash when programming a link-local neighbor
learned over such a session: `if_nbr_mac_to_ipv4ll_neigh_update()` calls
`kernel_neigh_update()` directly, bypassing the dplane abstraction, which
does not work with `dplane_grout`.

### Solution

**`ip6: punt received router advertisements to control plane`**

Add a new `ndp_ra_input` graph node that validates incoming Router
Advertisements per [RFC 4861 §6.1.2](https://www.rfc-editor.org/rfc/rfc4861#section-6.1.2):

- IP Hop Limit must be 255 (not forwarded by a router)
- ICMPv6 Code must be 0
- Packet length ≥ 4-byte ICMPv6 header + 12-byte RA fields
- IP source address must be link-local

Valid packets are punted to the control plane via the existing
`l4_loopback_output` path. Invalid packets are dropped via
`ndp_ra_input_inval`. A smoke test verifies an unnumbered eBGP session comes
up end-to-end between an FRR+grout instance and a pure FRR peer.

**`frr: patch zebra neighbor update crash with dplane_grout`**

Carry patches for FRR 10.5–10.7+ that route link-local neighbor updates
through the dplane API instead of calling `kernel_neigh_update()` directly.
(based on https://github.com/FRRouting/frr/pull/22773)

### Closes

Closes #657

## Summary

- Add validated IPv6 Router Advertisement handling. Forward valid packets to the control-plane TAP through `l4_loopback_output` and drop invalid packets.
- Apply FRR zebra patches for versions 10.5–10.7 and master. Route synthetic IPv4 link-local neighbor updates through the dplane API.
- Add a smoke test for unnumbered eBGP peering between grout and FRR.


modules/ip6/datapath/icmp6_input.c
```diff
@@ -80,6 +81,15 @@ icmp6_input_process(struct rte_graph *graph, struct rte_node *node, void **objs,
...
 		case ICMP6_TYPE_ROUTER_ADVERT:
+			// Grout does not process router advertisements itself.
+			// Punt them to the control plane so that a routing daemon
+			// (e.g. FRR) running on top of grout can consume them. BGP
+			// unnumbered peering relies on received RAs to discover the
+			// peer link-local next hop; without this the packets would
+			// be dropped as unsupported and the session would never come
+			// up.
+			next = ROUTER_ADVERT;
+			break;
```
rjarry, line 92: Your approach is fine but I would prefer if we could add an intermediate node to check for the RA packets compliance with RFCs. See my other PR for reference:

https://github.com/DPDK/grout/pull/660/changes/f5309fec06417a18bef87d3c6484c2e8b9edaa0b#diff-ae0101fec65d9c9f9e7645dff184ef9454a808762ad3162dbe93d129874eb37cR41-R55

smoke/bgp_unnumbered_frr_test.sh
```diff
@@ -0,0 +1,77 @@
...
+#  .------. |       .------------.             .------------.  | bgpd |   |
+#  | bgpd | |       |     p0     |   net_tap   |    x-p0    |  '------'   |
+#  '------' |       |    (LL)    +-------------+    (LL)    | .-------.   |
+#         .------.  '------------'             '------------' | zebra |   |
+#         | main |        |                          |       '-------'   |
+#         '------'        |                          |     .----------.  |
+#            |  ping <------------------------------------>|    lo    |  |
+#            '------------'                          |     | 16.0.0.1 |  |
+#                                                    |     '----------'  |
+#                                                    '-------------------'
```
rjarry: Also, the issue we have at the moment is that the test you added will fail all the time until there is a proper fix in zebra.

I think my approach in https://github.com/FRRouting/frr/pull/22773 is correct but I'd like some upstream reviews.

In the meantime, we "could" keep a local patch for FRR https://github.com/DPDK/grout/pull/660/changes/487aac58c75ed1178964e2b49dcf782bcd6796c2 so that we can test it in CI.

rjarry (changes requested):
Hi @chdxD1,

sorry not to have replied earlier.

I will close my PR https://github.com/DPDK/grout/pull/660 since we are basically doing the same thing.

Could you rebase your code on top of the latest main?

Thanks!

smoke/bgp_unnumbered_frr_test.sh
chdxD1: Sounds good, I can cherry pick your commit after rebasing :)

rjarry (changes requested):
Could you split in 3 commits? That way, the source tree always succeeds build/test at every step.

1. Punt RAs to control plane representors
2. Add zebra patch
3. Add smoke test

Thanks!

chdxD1:
> Could you split in 3 commits? That way, the source tree always succeeds build/test at every step.
> 
> 1. Punt RAs to control plane representors
> 2. Add zebra patch
> 3. Add smoke test
> 
> Thanks!

And I also noticed that I need to re-generated the svg. Will do!

chdxD1:
The CI failure here is not caused by this series. The smoke test added in the last commit is the first one to configure an address on the `main` VRF loopback, which uncovers a pre-existing leak in `iface_vrf_fini()`:

```
Direct leak of 128 byte(s) in 1 object(s) allocated from:
    #1 in event_new
    #2 in iface_loopback_create modules/infra/control/loopback.c:257
```

`netlink_vrf_del()` fails with `ESRCH` at teardown (`delete route on main failed: No such process`) and `iface_vrf_fini()` returns early, so `iface_loopback_destroy()` never runs.

Fixed separately in #681.

chdxD1:
@rjarry I think we need to re-order the commits:
1. zebra patch
2. then the actual RA punt commit
3. smoketest

Otherwise other tests fail as well when receiving RAs on the rebase build-and-test run.

Also this was very hard to figure out, given that it is not really obvious that only this specific check has this particular behaviour.

chdxD1:
And I think I found a generic issue with the FRR wrapping in meson, it does not rebuild when patch files or source files are changed.

rjarry:
> And I think I found a generic issue with the FRR wrapping in meson, it does not rebuild when patch files or source files are changed.

Oh, I thought it would have worked with this:

https://github.com/DPDK/grout/blob/6d5f98c801900c23ea4571721f6cc737595ce099/.github/workflows/check.sh#L11-L21

rjarry:
Hmm, possibly the FRR patch isn't applied:

```
Run-time dependency frr found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency frr
Downloading frr-10.5 source from https://github.com/FRRouting/frr/archive/refs/tags/frr-10.5.4.tar.gz

Executing subproject frr-10.5 

frr-10.5| Project name: frr
frr-10.5| Project version: 10.5.4
frr-10.5| C compiler for the host machine: ccache gcc-13 (gcc 13.3.0 "gcc-13 (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0")
frr-10.5| C linker for the host machine: gcc-13 ld.bfd 2.42
frr-10.5| Build targets in project: 185
frr-10.5| Subproject frr-10.5 finished.

Dependency frr from subproject subprojects/frr-frr-10.5.4 found: YES 10.5.4
Dependency systemd found: YES 255 (cached)
Program abidiff found: YES (/usr/bin/abidiff)
Program git found: YES (/usr/lib/git-core/git)
Program scdoc found: YES (/usr/bin/scdoc)
Run-time dependency cmocka found: YES 1.1.7
Build targets in project: 244
```

I don't know why...

rjarry:
On another test, it is:

```
Run-time dependency frr found: NO (tried pkgconfig and cmake)
Looking for a fallback subproject for the dependency frr
Downloading frr-10.5 source from https://github.com/FRRouting/frr/archive/refs/tags/frr-10.5.4.tar.gz
Applying diff file "frr/10.5-zebra-route-IPv4-link-local-neighbor-updates-through.patch"

Executing subproject frr-10.5 

frr-10.5| Project name: frr
frr-10.5| Project version: 10.5.4
frr-10.5| C compiler for the host machine: ccache clang-16 (clang 16.0.6 "Ubuntu clang version 16.0.6 (23ubuntu4)")
frr-10.5| C linker for the host machine: clang-16 ld.bfd 2.42
frr-10.5| Build targets in project: 185
frr-10.5| Subproject frr-10.5 finished.

Dependency frr from subproject subprojects/frr-frr-10.5.4 found: YES 10.5.4
Dependency systemd found: YES 255 (cached)
Program abidiff found: YES (/usr/bin/abidiff)
```

chdxD1:
From my understanding: the stamps live in the build folder, that is not purged by `meson subprojects purge [...]`.

Anyway, fixing the meson build will fix issues in local dev or in future PRs, not this particular bug.
Because the SRv6 test also triggers the crash in zebra now that FRR receives RAs. That's why we need the reordering for this PR here in general.

chdxD1:
One idea was to include the source files (.c / .h) of FRR in build depend_files.

By only including .c / .h, changes to files caused by autoreconf are not triggering a build. (There might still be a risk that a rebuild of FRR is skipped because of changes not to .c/.h files.

Another option is to forcefully remove the build directory in check.sh (or the stamps), however this of course increases the build time of consecutive builds (in CI only the rebase job)

```diff
 --- a/subprojects/packagefiles/frr/meson.build
 +++ b/subprojects/packagefiles/frr/meson.build
 @@ -18,6 +18,12 @@ prefix = meson.global_build_root()  / 'frr_install'
  # make sure the out-of-tree build directory exists
  run_command('mkdir', '-p', builddir, check: true)
  
 +find = run_command('find', srcdir, '-name', '*.[ch]', check: true)
 +sources = find.stdout().strip().split('\n')
 +
  bootstrap = custom_target(
    'bootstrap',
    output: '.bootstrap-stamp',
 @@ -85,6 +91,7 @@ build = custom_target(
    output: build_stamp,
    command: [ 'sh', '-c', build_cmd ],
    depends: configure,
 +  depend_files: sources,
    build_by_default: true,
  )
  alias_target('build', build
```

rjarry:
Yeah, it has nothing to do with your PR. Please do reorganize the commits so that the FRR patch is first. I'll merge your changes and find a proper fix for that CI bug.

rjarry (approved):

rjarry:
Thanks @chdxD1 !

PR #660, opened 2026-07-23
Ipv6 ra rx

rjarry:
@chdxD1 fyi

## Summary

- Added IPv6 Router Advertisement datapath handling with validation for hop limit, ICMP code, packet length, and link-local source address; valid packets proceed to control-plane loopback, while invalid packets are dropped.
- Added a smoke test for BGP unnumbered IPv6 peering using Router Advertisement-discovered link-local neighbors.
- Updated FRR integration for version-specific route cleanup APIs and routed IPv4 link-local neighbor updates through the dplane abstraction.
- Registered the required FRR patches for supported dependency versions.


rjarry:
Closing in favor of #658

PR #664, opened 2026-07-24
frr: sync nexthop groups from grout on startup

rjarry:
Recent FRR versions wrap every singleton nexthop in a nexthop group. Routes reference the group nh_id rather than the underlying L3 nexthop directly. When zebra restarts and the plugin re-reads routes from grout, those routes carry the group nh_id. If the group is not registered in zebra's NHG hash table, rib_sweep_route fails to resolve it and the orphan route is never purged.

The startup sync was not registering group nexthops at all. Individual NHs were synced per-VRF via grout_sync_nhs/grout_nexthop_change, but groups were silently skipped because grout_gr_nexthop_to_frr_nexthop only produced a dummy entry for them.

Split the NH sync into two passes: first individual nexthops (per-VRF, as before), then a single global pass for groups via the new grout_nexthop_group_add function which builds a proper struct nh_grp array and calls zebra_nhg_kernel_find. Groups must come second because they reference member nh_ids that must already exist in zebra's hash table. Groups are not per-VRF in grout (they have vrf_id=UNDEF), so the group pass queries once with GR_VRF_ID_UNDEF.

## Summary

- Added startup nexthop-group registration via `grout_nexthop_group_add(struct gr_nexthop *gr_nh, bool startup)` (declared in `frr/rt_grout.h`, implemented in `frr/rt_grout.c`).
  - Interprets `gr_nh->info` as a group nexthop payload (`gr_nexthop_info_group`) and builds a local `struct nh_grp grp[g->n_members]` from member nexthop IDs plus weights.
  - Derives the Zebra NHG `type` using `origin2zebra(gr_nh->origin, AF_UNSPEC, true)` (address-family handling kept as `UNSPEC` rather than reduced to IP).
  - Converts the grout VRF to FRR VRF (`vrf_grout_to_frr()`) and installs/registers the group with `zebra_nhg_kernel_find(..., startup, ...)` using the assembled member list.

- Split startup synchronization into two phases in `frr/zebra_dplane_grout.c`:
  1. Per-VRF pass registers individual nexthops only by skipping entries with type `GR_NH_T_GROUP` in `grout_sync_nhs`.
  2. Global pass (`grout_sync_nh_groups`) streams group nexthops once using `.vrf_id = GR_VRF_ID_UNDEF` (group entries referencing member IDs) and registers each group, then starts route synchronization after the group pass.

- Routes referencing group IDs are resolved correctly by ensuring group NHGs are registered only after their member nexthop IDs already exist in Zebra’s nexthop hash table; route sync starts after the group pass to avoid orphan routes across Zebra restarts.

### DIFF
--- a/frr/rt_grout.c
+++ b/frr/rt_grout.c
@@ -905,6 +905,32 @@ enum zebra_dplane_result grout_add_del_nexthop(struct zebra_dplane_ctx *ctx) {
 	return grout_add_nexthop(nh_id, origin, dplane_ctx_get_nhe_ng(ctx)->nexthop);
 }
 
+void grout_nexthop_group_add(struct gr_nexthop *gr_nh, bool startup) {
+	const struct gr_nexthop_info_group *g = (const struct gr_nexthop_info_group *)gr_nh->info;
+	struct nh_grp grp[g->n_members];
+	int type;
+
+	gr_log_debug("add group nh_id %u n_members %u", gr_nh->nh_id, g->n_members);
+
+	for (uint32_t i = 0; i < g->n_members; i++) {
+		grp[i].id = g->members[i].nh_id;
+		grp[i].weight = g->members[i].weight;
+	}
+
+	type = origin2zebra(gr_nh->origin, AF_UNSPEC, true);
+	zebra_nhg_kernel_find(
+		gr_nh->nh_id,
+		NULL,
+		grp,
+		g->n_members,
+		vrf_grout_to_frr(gr_nh->vrf_id),
+		AFI_UNSPEC,
+		type,
+		startup,
+		NULL
+	);
+}
+
 void grout_nexthop_change(bool new, struct gr_nexthop *gr_nh, bool startup) {
 	struct nexthop *nh = NULL;
 	afi_t afi = AFI_UNSPEC;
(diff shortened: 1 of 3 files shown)


frr/rt_grout.c
```diff
@@ -905,6 +905,32 @@ enum zebra_dplane_result grout_add_del_nexthop(struct zebra_dplane_ctx *ctx) {
...
+	int type;
+
+	gr_log_debug("add group nh_id %u n_members %u", gr_nh->nh_id, g->n_members);
+
+	for (uint32_t i = 0; i < g->n_members; i++) {
+		grp[i].id = g->members[i].nh_id;
+		grp[i].weight = g->members[i].weight;
+	}
+
+	type = origin2zebra(gr_nh->origin, AF_INET, true);
```
christophefontaine: Do we always have AF_INET, and never AF_INET6 for next hop groups ?

frr/rt_grout.c
```diff
@@ -905,6 +905,32 @@ enum zebra_dplane_result grout_add_del_nexthop(struct zebra_dplane_ctx *ctx) {
...
+	zebra_nhg_kernel_find(
+		gr_nh->nh_id,
+		NULL,
+		grp,
+		g->n_members,
+		vrf_grout_to_frr(gr_nh->vrf_id),
+		AFI_IP,
```
christophefontaine: Same as above, do we have AFI_IP only ?

rjarry:
I fixed both AF_* values to use UNSPEC. It does not make any sense to reduce to IP.

christophefontaine (approved):

PR #665, opened 2026-07-24
srv6: resolve nexthop groups in sr6_output

rjarry:
When FRR wraps every nexthop in a group, fib6_lookup returns a group nexthop instead of the underlying L3 nexthop. In sr6_output, the secondary FIB lookup for the SID prefix is followed by a call to sr_tunsrc_get which reads nh->iface_id directly. Group nexthops have iface_id set to GR_IFACE_ID_UNDEF, causing sr_tunsrc_get to fail and dropping the packet at sr6_source_no_route.

Resolve the group to its member nexthop right after the fib6_lookup so that subsequent field accesses see the actual L3 nexthop.

### DIFF
--- a/modules/srv6/datapath/srv6_output.c
+++ b/modules/srv6/datapath/srv6_output.c
@@ -6,6 +6,7 @@
 #include "ip6.h"
 #include "ip6_datapath.h"
 #include "l3.h"
+#include "nexthop.h"
 #include "srv6.h"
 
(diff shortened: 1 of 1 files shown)


rjarry:
Same issue may happen in many places. I'll close this and rework it to include the nexthop group resolution inline with fib{4,6}_lookup.

PR #666, opened 2026-07-27
ip,ip6: resolve nexthop groups in fib lookup

rjarry:
Move ECMP load balancing from dedicated datapath nodes and scattered call sites into fib4_lookup() and fib6_lookup() themselves. Both functions now take a flow_id argument (typically mbuf->hash.rss) and resolve GR_NH_T_GROUP nexthops before returning, so callers always get a concrete nexthop.

This removes the ip_loadbalance and ip6_loadbalance graph nodes entirely, along with the ad-hoc group resolution in icmp_local_send and icmp6_local_send. More importantly, it fixes bugs in call sites that never handled nexthop groups at all (e.g. srv6_output, vxlan_output, dnat44) and would crash or silently misbehave when a route resolved to an ECMP group.

## Summary

- Moved ECMP nexthop-group resolution into `fib4_lookup()` / `fib6_lookup()` by:
  - Extending `fib4_lookup(vrf_id, ip, flow_id)` and `fib6_lookup(vrf_id, iface_id, ip, flow_id)` to accept a `flow_id`.
  - After converting the FIB nexthop id to a `struct nexthop *`, detecting `GR_NH_T_GROUP` and selecting a concrete per-`flow_id` member via `nexthop_group_get_nh(..., flow_id)`.
  - Returning `ENONET` when group-to-member selection returns `NULL` (instead of returning an unresolved group nexthop pointer).

- Updated nexthop lookup call sites to pass a flow identifier:
  - ICMP control local-send:
    - IPv4 `icmp_send` uses `fib4_lookup(req->vrf, req->addr, req->ident)` (replacing `rib4_lookup(req->vrf, req->addr)`) and keeps the existing null lookup / `-errno` error return behavior.
    - IPv6 `icmp6_send` uses `fib6_lookup(req->vrf, req->iface, &req->addr, req->ident)` (replacing the prior `rib6_lookup(...)` usage).
  - Packet datapath/encapsulation/output and error paths:
    - Passes `mbuf->hash.rss` into `fib4_lookup()` / `fib6_lookup()` for ICMP output/error, IPv4/IPv6 input, IP-in-IP output, VXLAN underlay output, DNAT44 dynamic/static, and SRv6 output.

- Removed ECMP load-balancing graph nodes and ad-hoc group handling from local-send paths:
  - Deleted `ip_loadbalance` and `ip6_loadbalance` datapath nodes (including their build entries and `ip*_lb_no_nexthop` drop registrations).
  - Removed `GR_NH_T_GROUP` handling blocks from `icmp_local_send.c` / `icmp6_local_send.c`, so local-send no longer resolves group nexthops via `nexthop_group_get_nh()`. Group nexthops now follow the existing `gw->type != GR_NH_T_L3` validation, failing via the `ENONET` path.

- Added optimized single-member group shortcut behavior:
  - `modules/infra/control/nexthop.h`: extended `GR_NH_T_GROUP` info with a single-member `nh` shortcut; `nexthop_group_get_nh()` returns the shortcut immediately when `n_members == 1` and returns `NULL` when `n_members == 0`.
  - `modules/infra/control/group_nexthop.c`: special-cased `group->n_members == 1` to store the sole member directly in `pvt->nh` (avoiding RETA allocation logic) and updated member removal/import handling accordingly.


modules/ipip/datapath_out.c
```diff
@@ -75,7 +75,7 @@ ipip_output_process(struct rte_graph *graph, struct rte_node *node, void **objs,
 		IFACE_STATS_INC(tx, self, mbuf, iface);
 
 		// Resolve nexthop for the encapsulated packet.
-		ip_data->nh = fib4_lookup(iface->vrf_id, ipip->remote);
+		ip_data->nh = fib4_lookup(iface->vrf_id, ipip->remote, mbuf->hash.rss);
 		edge = IP_OUTPUT;
```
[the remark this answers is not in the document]
rjarry: This was already like this before. The return value was not checked and ip_output deals with it.

christophefontaine (approved):
@maxime-leroy any feedback about this PR ?
Mainly if this bothers you, as we may want to introduce a more complex (configurable) algorithm for ECMP ?

rjarry:
> we may want to introduce a more complex (configurable) algorithm for ECMP ?

If that is the case, this could be handled transparently in `nexthop_group_get_nh()`. The removed `*_loadbalance` nodes didn't do anything else than call this function anyway.

modules/ip/control/route.c
```diff
@@ -152,7 +152,14 @@ const struct nexthop *fib4_lookup(uint16_t vrf_id, ip4_addr_t ip) {
 	if (nh_id == 0)
 		return errno_set_null(EHOSTUNREACH);
 
-	return nh_id_to_ptr(nh_id);
+	const struct nexthop *nh = nh_id_to_ptr(nh_id);
+	if (nh->type == GR_NH_T_GROUP) {
+		nh = nexthop_group_get_nh(nexthop_info_group(nh), flow_id);
```
maxime-leroy, line 157: Performance impact ?

modules/ipip/datapath_out.c
```diff
@@ -75,7 +75,7 @@ ipip_output_process(struct rte_graph *graph, struct rte_node *node, void **objs,
 		IFACE_STATS_INC(tx, self, mbuf, iface);
 
 		// Resolve nexthop for the encapsulated packet.
-		ip_data->nh = fib4_lookup(iface->vrf_id, ipip->remote);
+		ip_data->nh = fib4_lookup(iface->vrf_id, ipip->remote, mbuf->hash.rss);
```
maxime-leroy, line 78: What about tunneling decap ? what about port not supporting rss hash ?

For example, hardware hashing only outer should not use hash.rss for inner packet.
But for hardware hashing inner/and outer is fine.

modules/ip/control/route.c
rjarry: If we want to support nexthop groups, we have no choice. The check must be done *at some point*. This patch fixes the many places where the type of nexthop was not checked.

Mind that FRR 10.8+ will always configure nexthop groups. Even for single nexthops.

I had considered "optimizing" this either in the FRR plugin or in the grout control plane. But it will be super complicated and fragile. I don't want to spend 2 months fixing bugs just to avoid one if condition.

modules/ip6/control/route.c
```diff
@@ -158,7 +158,14 @@ fib6_lookup(uint16_t vrf_id, uint16_t iface_id, const struct rte_ipv6_addr *ip)
 	if (nh_id == 0)
 		return errno_set_null(EHOSTUNREACH);
 
-	return nh_id_to_ptr(nh_id);
+	const struct nexthop *nh = nh_id_to_ptr(nh_id);
+	if (nh->type == GR_NH_T_GROUP) {
+		nh = nexthop_group_get_nh(nexthop_info_group(nh), flow_id);
```
maxime-leroy, line 163: static inline struct nexthop *
nexthop_group_get_nh(struct nexthop_info_group *nhg, uint32_t flow_id) {
	if (unlikely(nhg->n_members == 0))
		return NULL;
	return nhg->reta[flow_id & (nhg->reta_size - 1)];
}


it seems possible to add : if (likely(nhg->n_members == 1)) return return nhg->reta[0];

Any idea to avoid this indirection ?

modules/ipip/datapath_out.c
rjarry: Then it will take whatever value is in mbuf->hash.rss and use that as flow_id. It will return one of the nexthops of the group.

maxime-leroy (changes requested):

modules/ip6/control/route.c
rjarry: it would mean adding another test. If you are concerned about the access to the large reta array, we could think of a middle ground that reduces the size of reta to 1 when there is a single nexthop in the group. That would give the same result.

rjarry: Scratch that. The `reta` array is already taking the correct amount of space based on the number of nexthops and their respective weights. If you have a single nexthop with weight 1, `reta_size=1`

So `nhg->reta[flow_id & (nhg->reta_size - 1)]` is *exactly* `nhg->reta[0]` regardless of `flow_id`.

maxime-leroy:
Two bugs in "group: bypass redirection table for single nexthops".

  1. group_import_info(): pvt->nh is stored ~50 lines before the block that
     publishes n_members/reta_size/reta. On a 1 -> N update the datapath sees
     n_members == 1 with nh == NULL until the reta is built, and blackholes.
     And if the reta alloc fails, pvt->nh stays NULL while n_members is still
     1, so a failed (no-op) update kills the group for good. Please build it in a
     local and publish it with the others.

  2. remove_group_member_cb() does not update pvt->nh. On interface deletion
     nh_cleanup_interface_cb() destroys a member even though the group holds a
     ref, so a 2-member group ends up n_members == 1 with nh == NULL: whole
     route blackholed instead of forwarding over the survivor. Repro:
     smoke/ip_loadbalance_test.sh + grcli interface del p1.

Also, ip_output no longer registers an edge for GR_NH_T_GROUP, so a group
reaching it is read as an L3 nexthop (garbage MAC) instead of dropped. Cheap to
keep a drop edge, and under 10.8 that invariant covers all traffic.

maxime-leroy:
Pool sizing. The parent/leaf split roughly doubles nexthop objects for the same number of distinct nexthops. DEFAULT_MAX_COUNT (modules/infra/control/nexthop.c:34) is worth revisiting against a full-view deployment before
 someone hits the mempool exhaustion.

maxime-leroy:
```

static void ip_input_route_bulk(void **objs, uint16_t n, const struct nexthop **nhs) {
        uint32_t keys[IP_INPUT_LOOKUP_BULK];
        uint16_t vrfs[IP_INPUT_LOOKUP_BULK];
        struct rte_ipv4_hdr *ip;
        struct rte_mbuf *mbuf;
        uint16_t i, r;

        // Collect the lookup keys. Packets that header validation will reject are
        // looked up as well: filtering them out here would mean duplicating the
        // checks, and their result is never read.
        for (i = 0; i < n; i++) {
                mbuf = objs[i];
                if (unlikely(rte_pktmbuf_data_len(mbuf) < sizeof(struct rte_ipv4_hdr))) {
                        // Header not readable: no key, no iface deref. The walk
                        // drops these with BAD_LENGTH.
                        vrfs[i] = GR_VRF_ID_UNDEF;
                        nhs[i] = NULL;
                        continue;
                }
                ip = rte_pktmbuf_mtod(mbuf, struct rte_ipv4_hdr *);
                keys[i] = rte_be_to_cpu_32(ip->dst_addr);
                vrfs[i] = eth_input_mbuf_data(mbuf)->iface->vrf_id;
        }
  
        // One bulk lookup per run of consecutive packets sharing the same VRF.
        for (i = 0; i < n; i = r) {
                if (vrfs[i] == GR_VRF_ID_UNDEF) {
                        r = i + 1;
                        continue;
                }
                for (r = i + 1; r < n && vrfs[r] == vrfs[i]; r++)
                        ;
                fib4_lookup_bulk(vrfs[i], &keys[i], &nhs[i], r - i);
        }

        // Get every fib result in flight before dereferencing any of them.
        for (i = 0; i < n; i++)
                rte_prefetch0(nhs[i]);

        // Then resolve the groups and prefetch the members they point to, so that
        // the walk below finds both lines warm.
        for (i = 0; i < n; i++) {
                if (nhs[i] == NULL)
                        continue;
                mbuf = objs[i];
                nhs[i] = nexthop_resolve(nhs[i], mbuf->hash.rss);
                rte_prefetch0(nhs[i]);
        }
}


```

maxime-leroy:
nexthop_new() does memset(nh, 0, sizeof(*nh)) (nexthop.c:324) before calling nexthop_update(), which is what dispatches to group_import_info(). So on a create the group's private info starts all zeroes, and by the time you reach
  the commit block:

        n_tmp = pvt->n_members;   // 0
        tmp = pvt->members;       // NULL
        old_reta = pvt->reta;     // NULL
        ...
        rte_rcu_qsbr_synchronize(gr_datapath_rcu(), RTE_QSBR_THRID_INVALID);

        for (uint32_t i = 0; i < n_tmp; i++)   // does not run
                nexthop_decref(tmp[i].nh);
  
        rte_free(old_reta);   // no-op on NULL
        rte_free(tmp);        // no-op on NULL

  Everything the synchronize exists to protect is a no-op. It is there so that no datapath thread is still walking the old members/reta when they get freed. On a create there is no old anything, so the wait guards nothing and just
  blocks the API thread until every worker has reported a quiescent state.

  On an update of an existing group, tmp and old_reta are the previous arrays, the datapath may still be dereferencing them through nexthop_group_get_nh(), and the wait is genuinely required.

  So the two cases are create versus update, and the guard is just "did we take anything away that still needs reclaiming":

        if (tmp != NULL || old_reta != NULL)
                rte_rcu_qsbr_synchronize(gr_datapath_rcu(), RTE_QSBR_THRID_INVALID);

christophefontaine (approved):

PR #667, opened 2026-07-27
arp: only reply for IPs owned by the receiving iface

ramperher:
We have the following test scenario where, because of resource limitations, we have to split a network (/24) in two subnets (/28), representing two different networks but in the same VLAN.

Here's a diagram of the scenario:

```
                       VLAN 3801 (grout-test-net1)          VLAN 3802 (grout-test-net2)
                       172.16.16.0/24                       SAME L2 — two /28s carved on it
                       ============================         =====================================
                                                            172.16.21.0/28   |  172.16.21.16/28
                                                            (east LAN)       |  (interconnect)
                                                            -----------------+------------------
grout-west-endpoint                    grout-west-router                 grout-east-router                grout-east-endpoint
-------------------                    -----------------                 -----------------                -------------------
net1 VF                                net1 VF          net2 VF          net1 VF          net2 VF         net1 VF
172.16.16.10/24                        172.16.16.1/24   172.16.21.18/28  172.16.21.1/28   172.16.21.19/28 172.16.21.10/28
MAC :16:10                             MAC :16:01       MAC :21:18       MAC :21:01       MAC :21:19      MAC :21:10
     |                                      |                |                |                |               |
     +--------------------------------------+                |                |                |               |
            west LAN (VLAN 3801)                             |                |                |               |
                                                             +----------------+----------------+---------------+
                                                                        east LAN + interconnect
                                                                        (VLAN 3802 — one broadcast domain)
                                                                        PF ens7f0, VFs 8-15
IP routes:
  west-endpoint:  0.0.0.0/0          via 172.16.16.1
  east-endpoint:  0.0.0.0/0          via 172.16.21.1
  west-router:    172.16.21.0/28     via 172.16.21.19   (east-router.net2)
  east-router:    172.16.16.0/24     via 172.16.21.18   (west-router.net2)
```

Based on this, we're finding a problem. When several grout ports share one L2 domain (e.g. two SR-IOV VFs on the same VLAN), an ARP request for a local IP can be received on any of those ports:

```
Bug on VLAN 3802:
  ARP who-has 172.16.21.1  ----flooded to BOTH east-router VFs---->
       p0 (:21:01) owns .1  -> reply  .1 is-at :21:01   (correct)
       p1 (:21:19) sees .1  -> reply  .1 is-at :21:19   (wrong, before fix)
East-router is the problem node: two VFs, one VLAN, two IPv4 subnets.
```

Lookup was VRF-wide and only checked `GR_NH_F_LOCAL`, so a port that does not own the target IP could still reply using its own MAC. That produced wrong neighbor bindings and intermittent one-way ping failures.

This change scopes ARP replies to addresses configured on the receiving interface (`local->iface_id == iface->id`), in both `arp_input_request` and `arp_probe_input_cb`, matching the existing IPv6 NDP behavior. A smoke test covers two ports on a shared Linux bridge, and CI/docs gain the `arping` dependency.

## Summary

- Restrict ARP replies to local nexthops configured on the receiving interface in both request and probe handling.
- Add a regression smoke test for two ports sharing an L2 bridge, including interface-down scenarios.
- Add `arping` dependencies to CI and development setup documentation.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -83,7 +83,7 @@ jobs:
             make gcc gdb ccache ninja-build meson git scdoc inotify-tools \
             libibverbs-dev libasan8 libcmocka-dev libedit-dev libarchive-dev \
             libevent-dev libmnl-dev libnuma-dev python3-pyelftools \
-            socat tcpdump traceroute graphviz iproute2 iputils-ping ndisc6 jq \
+            socat tcpdump traceroute graphviz iproute2 iputils-ping iputils-arping ndisc6 jq \
             dnsmasq systemd-coredump abigail-tools \
             "linux-modules-extra-$(uname -r)"
           # FRR dependencies
--- a/README.md
+++ b/README.md
@@ -287,13 +287,13 @@ In order to run the `smoke-tests`, `lint`, `check-patches` and `update-graph`
 targets, you'll need additional packages:
 
 ```sh
-dnf install gawk gdb clang-tools-extra jq codespell curl traceroute graphviz ndisc6 abidiff inotify-tools
+dnf install gawk gdb clang-tools-extra jq codespell curl traceroute graphviz ndisc6 iputils abidiff inotify-tools
 ```
 
 or
 
 ```sh
-apt install gawk gdb clang-format jq codespell curl traceroute graphviz ndisc6 abigail-tools inotify-tools
+apt install gawk gdb clang-format jq codespell curl traceroute graphviz ndisc6 iputils-arping abigail-tools inotify-tools
 ```
 
 ### Build
--- a/modules/ip/control/nexthop.c
+++ b/modules/ip/control/nexthop.c
@@ -195,17 +195,25 @@ void arp_probe_input_cb(void *obj, uintptr_t, const struct control_queue_drain *
 	}
 
 	if (arp->arp_opcode == RTE_BE16(RTE_ARP_OP_REQUEST)) {
-		// send a reply for our local ip
+		// Reply only if the target is a local address on this iface.
+		// Sender learning above still applies when we skip the reply.
 		struct nexthop *local = nh4_lookup(iface->vrf_id, arp->arp_data.arp_tip);
-		struct arp_reply_mbuf_data *d = arp_reply_mbuf_data(m);
-		d->local = local;
-		d->iface = iface;
-		if (post_to_stack(arp_output_reply_node, m) < 0) {
-			LOG(ERR, "post_to_stack: %s", strerror(errno));
-			goto free;
+		const struct nexthop_info_l3 *local_l3;
+
+		if (local != NULL) {
+			local_l3 = nexthop_info_l3(local);
+			if ((local_l3->flags & GR_NH_F_LOCAL) && local->iface_id == iface->id) {
+				struct arp_reply_mbuf_data *d = arp_reply_mbuf_data(m);
+				d->local = local;
+				d->iface = iface;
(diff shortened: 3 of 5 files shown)


rjarry:
Hey Ramon,

on the principle yes, we should only reply once. But I fear your changes will break the resolution of addresses configured on vrf interfaces (loopbacks). As you can never receive packets on these interfaces, grout will never respond to an ARP request on an IP address configured on them.

I don't know how other networking stacks handle this edge case, but it would be good to get inspiration.

rjarry:
It looks like grout behaves like the default linux stack:

By default Linux responds to ARP requests on all interfaces, even if the IP is only configured on one of them. This is known as ARP flux.

### Default behavior (arp_ignore=0, arp_announce=0)

Linux uses a host-based (weak host model) approach to ARP, not a per-interface one. When an ARP request arrives on any interface, the kernel checks if the requested IP exists anywhere on the host (same netns/VRF). If it does, it replies — regardless of
which interface the address is configured on.

So with two interfaces `eth0` (`10.0.0.1`) and `eth1` (no address) on the same L2 segment:

- ARP `who-has 10.0.0.1?` arriving on `eth0` → reply from `eth0` MAC
- ARP `who-has 10.0.0.1?` arriving on `eth1` → reply from `eth1` MAC

Both reply. The remote host ends up with whichever reply arrived last, and traffic may land on the "wrong" interface.

### Loopback addresses

Same story. An IP on `lo` is still a local address in the host's routing table. Both `eth0` and `eth1` will respond to ARP requests for it. This is actually a common pattern in load-balancing setups (e.g., DSR / Direct Server Return): you put the VIP on
loopback and suppress ARP for it with sysctl tunables.

Two sysctls under `/proc/sys/net/ipv4/conf/<iface>/`:

- `arp_ignore` — controls whether to respond:
  - `0` (default): reply for any local address, on any interface
  - `1`: reply only if the address is configured on the incoming interface
  - `2`: like `1`, plus the source IP must be on the same subnet
- `arp_announce` — controls source address selection in outgoing ARP requests:
  - `0` (default): use any local address
  - `1`: prefer an address on the outgoing interface
  - `2`: always use the best local address for the target subnet

For the loopback VIP / DSR case, the typical config is:

```
sysctl net.ipv4.conf.all.arp_ignore=1
sysctl net.ipv4.conf.all.arp_announce=2
```

This prevents the interfaces from answering ARP for the loopback VIP while still allowing normal ARP for their own addresses.

NB: When interfaces are enslaved to a VRF device, the ARP scope is limited to that VRF's routing table — but within that VRF, the same weak-host behavior applies unless arp_ignore/arp_announce are tuned.

ramperher:
@rjarry , thanks for your feedback and for sharing some light in the behavior of grout in this scenario. So, I assume it is better to properly configure `arp_ignore` and `arp_announce` in the interfaces, rather than applying this patch on grout. Is it correct?

rjarry:
No, I was highlighting the **Linux** stack behavior. Grout does not have any settings to configure arp_ignore and arp_announce yet.

rjarry:
@ramperher could you try changing `grout-east-router` to have a single VF with both its IP addresses configured on it? It should fix your problem.

ramperher:
> @ramperher could you try changing grout-east-router to have a single VF with both its IP addresses configured on it? It should fix your problem.

Yes, we will try to do some modifications in the deployment itself, rather than patching grout, at least for the moment. If some features like arp_ignore are eventually interesting for grout, we'll think on proposing a patch for that.

Closing this for the moment.

PR #672, opened 2026-07-29
Mbuf pool drain

rjarry:
This series fixes the following NULL pointer dereference when an ICMP message arrives while reconfiguring CPU affinity:

```
(gdb) bt -full
#0  get_icmp_response (ident=19013, seq_num=0, timestamp=<synthetic pointer>) at ../modules/ip/control/icmp.c:55
        icmp = 0x0
        i = 0x17ff88b80
        tmp = 0x17ff88c40
        mbuf = 0x0
        i = <optimized out>
        tmp = <optimized out>
        mbuf = <optimized out>
        icmp = <optimized out>
        ip = <optimized out>
#1  icmp_recv (request=<optimized out>) at ../modules/ip/control/icmp.c:114
        icmp_req = <optimized out>
        pkt_timestamp = <optimized out>
        rcv_timestamp = <optimized out>
        resp = 0x0
        icmp = <optimized out>
        ip = <optimized out>
--Type <RET> for more, q to quit, c to continue without paging--bt -full
        m = <optimized out>
        len = 0
        ret = 0
#2  0x000055fd55bb36df in read_cb (bev=0x55fd580b4570, priv=0x55fd580b48b0) at ../main/api.c:315
        input = 0x55fd580b4790
        ctx = 0x55fd580b48b0
        req_payload = 0x55fd58330740
        __PRETTY_FUNCTION__ = "read_cb"
        __func__ = "read_cb"
        out = <optimized out>
        handler = 0x7f0b260bc100
        send = <optimized out>
        resp = {for_id = 0, status = 0, payload_len = 1}
#3  0x00007f0b271c7f2c in bufferevent_run_readcb_ () from /lib64/libevent_core-2.1.so.7
No symbol table info available.
#4  0x00007f0b271cac18 in bufferevent_readcb.lto_priv () from /lib64/libevent_core-2.1.so.7
No symbol table info available.
#5  0x00007f0b271d4847 in event_process_active_single_queue () from /lib64/libevent_core-2.1.so.7
No symbol table info available.
#6  0x00007f0b271d662f in event_base_loop () from /lib64/libevent_core-2.1.so.7
No symbol table info available.
#7  0x000055fd55ab61ff in main (argc=8, argv=<optimized out>) at ../main/main.c:349
        ev_base = 0x55fd57fbec40
        ret = 1
        err = 0
        __func__ = "main"
        shutdown = <optimized out>
```

The timeline of the crash:

1. An ICMP echo reply arrives, mbuf is allocated from the port's mempool
2. Mbuf goes through the graph -> control_output -> control queue ring -> icmp_input_cb -> stored in icmp_queue
3. An API request to change CPU affinity arrives -- we're now inside read_cb in the libevent loop
4. worker_queue_distribute stops all workers, then calls port_configure for each port
5. port_configure calls gr_pktmbuf_pool_release(p->pool, ...) -- if the ref count drops to zero, the pool backing the ICMP mbuf is freed
6. read_cb returns, libevent loop resumes
7. Later, icmp_recv -> get_icmp_response accesses i->mbuf -- the mbuf's memory has been freed/unmapped -> buf_addr is 0 -> rte_pktmbuf_mtod returns 0 -> segfault

The icmp_queue (and potentially any pending items in the control queue ring) holds mbufs from the old pool that nobody drains before the pool is freed. The RCU mechanism doesn't help here because these mbufs are held by the control plane itself, not the datapath.

## Summary

- Resize existing packet-buffer mempools instead of releasing and recreating them during port reconfiguration.
- Defer mempool destruction until all outstanding mbufs are released. Track pending pools and report pools that remain busy.
- Add shared APIs for adding, resetting, flushing, and iterating L3 nexthop hold queues.
- Remove queued ICMP and ICMPv6 packets when their interface is removed.
- Expire stale ICMP queues with garbage-collection timers.

### DIFF
--- a/modules/dhcp/control/client.c
+++ b/modules/dhcp/control/client.c
@@ -527,7 +527,7 @@ static void dhcp_fini(struct event_base *) {
 
 static struct module dhcp_module = {
 	.name = "dhcp",
-	.depends_on = "graph,ip_address,iface",
+	.depends_on = "graph,ip_address,iface,mempool",
 	.init = dhcp_init,
 	.fini = dhcp_fini,
 };
--- a/modules/ip/control/icmp.c
+++ b/modules/ip/control/icmp.c
@@ -60,6 +60,18 @@ static void icmp_event_cb(uint32_t ev_type, const void *obj) {
 	}
 }
 
+#define ICMP_QUEUE_TIMEOUT (10 * GR_NS_PER_S)
+
+static void icmp_queue_gc(evutil_socket_t, short, void *) {
+	gr_clock_ns_t now = gr_clock_ns();
+	struct icmp_queue_item *i, *tmp;
+
+	STAILQ_FOREACH_SAFE (i, &icmp_queue, next, tmp) {
+		if (now - i->timestamp >= ICMP_QUEUE_TIMEOUT)
+			icmp_queue_pop(i, true);
+	}
+}
+
 // Search for the oldest ICMP response matching the given identifier.
 // If found, the packet is removed from the queue.
 static struct rte_mbuf *
@@ -169,8 +181,9 @@ out:
 }
 
 #define ICMP_LOCAL_QUEUE_SIZE 1024
+static struct event *gc_timer;
 
-static void icmp_init(struct event_base *) {
+static void icmp_init(struct event_base *ev_base) {
 	pool = rte_mempool_create(
 		"icmp_queue", // name
 		ICMP_LOCAL_QUEUE_SIZE,
@@ -186,9 +199,18 @@ static void icmp_init(struct event_base *) {
 	);
 	if (pool == NULL)
 		ABORT("rte_mempool_create(icmp_queue) failed");
+
+	gc_timer = event_new(ev_base, -1, EV_PERSIST | EV_FINALIZE, icmp_queue_gc, NULL);
+	if (gc_timer == NULL)
+		ABORT("event_new() failed");
+	if (event_add(gc_timer, &(struct timeval) {.tv_sec = 1}) < 0)
+		ABORT("event_add() failed");
 }
 
 static void icmp_fini(struct event_base *) {
+	if (gc_timer)
+		event_free(gc_timer);
+
 	if (pool != NULL) {
 		struct icmp_queue_item *i, *tmp;
 		STAILQ_FOREACH_SAFE (i, &icmp_queue, next, tmp)
--- a/modules/ip/control/icmp.c
+++ b/modules/ip/control/icmp.c
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: BSD-3-Clause
 // Copyright (c) 2024 Christophe Fontaine
 
+#include "event.h"
 #include "ip4.h"
 #include "ip4_datapath.h"
 #include "log.h"
@@ -29,9 +30,15 @@ static void icmp_queue_pop(struct icmp_queue_item *i, bool free_mbuf) {
 
 // Callback invoked by control plane for each ICMP packet received for a local address.
 // The packet is added at the end of a linked list.
-static void icmp_input_cb(void *m, uintptr_t timestamp, const struct control_queue_drain *) {
+static void icmp_input_cb(void *m, uintptr_t timestamp, const struct control_queue_drain *drain) {
 	struct icmp_queue_item *i;
 	void *data;
+
+	if (drain != NULL && drain->event == GR_EVENT_IFACE_REMOVE
+	    && mbuf_data(m)->iface == drain->obj) {
+		rte_pktmbuf_free(m);
+		return;
+	}
 
 	while (rte_mempool_get(pool, &data) < 0)
 		icmp_queue_pop(STAILQ_FIRST(&icmp_queue), true);
@@ -40,6 +47,17 @@ static void icmp_input_cb(void *m, uintptr_t timestamp, const struct control_que
 	i->mbuf = m;
 	i->timestamp = timestamp;
 	STAILQ_INSERT_TAIL(&icmp_queue, i, next);
+}
+
+static void icmp_event_cb(uint32_t ev_type, const void *obj) {
+	struct icmp_queue_item *i, *tmp;
+
+	if (ev_type == GR_EVENT_IFACE_REMOVE) {
+		STAILQ_FOREACH_SAFE (i, &icmp_queue, next, tmp) {
+			if (mbuf_data(i->mbuf)->iface == obj)
+				icmp_queue_pop(i, true);
+		}
+	}
 }
 
 // Search for the oldest ICMP response matching the given identifier.
@@ -190,6 +208,7 @@ RTE_INIT(icmp_module_init) {
 	module_register(&icmp_module);
 	api_handler(GR_IP4_ICMP_SEND, icmp_send);
 	api_handler(GR_IP4_ICMP_RECV, icmp_recv);
+	event_subscribe(GR_EVENT_IFACE_REMOVE, icmp_event_cb);
 	icmp_input_register_callback(RTE_ICMP_TYPE_DEST_UNREACHABLE, icmp_input_cb);
 	icmp_input_register_callback(RTE_ICMP_TYPE_TTL_EXCEEDED, icmp_input_cb);
 	icmp_input_register_callback(RTE_ICMP_TYPE_ECHO_REPLY, icmp_input_cb);
--- a/.clang-format
+++ b/.clang-format
@@ -81,11 +81,12 @@ ForEachMacros:
   - STAILQ_FOREACH_SAFE
   - TAILQ_FOREACH
   - TAILQ_FOREACH_SAFE
-  - rte_graph_foreach_node
   - gr_api_client_stream_foreach
+  - gr_nh_flags_foreach
+  - nexthop_l3_hold_queue_foreach
+  - rte_graph_foreach_node
   - vec_foreach
   - vec_foreach_ref
-  - gr_nh_flags_foreach
 IfMacros:
   - KJ_IF_MAYBE
 IncludeBlocks: Regroup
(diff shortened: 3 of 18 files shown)


modules/ip/control/icmp.c
```diff
@@ -29,10 +29,25 @@ static void icmp_queue_pop(struct icmp_queue_item *i, bool free_mbuf) {
...
 	void *data;
 
+	if (drain != NULL) {
+		switch (drain->event) {
+		case GR_EVENT_IFACE_REMOVE:
+			STAILQ_FOREACH_SAFE (i, &icmp_queue, next, tmp) {
+				if (mbuf_data(i->mbuf)->iface == drain->obj)
+					icmp_queue_pop(i, true);
+			}
+			if (mbuf_data(m)->iface == drain->obj) {
```
david-marchand: I am not sure I understand the draining API and the relation with the control queue.

For a drain event, the m input parameter should be NULL ?

rjarry: The drain api is purging the control ring which invokes all the registered callbacks with their drained mbufs. So no, it will not be NULL.

https://github.com/DPDK/grout/blob/44f173bd2db423c07b703e25c57a6f2513971a97/main/control_queue.c#L102-L105

https://github.com/DPDK/grout/blob/44f173bd2db423c07b703e25c57a6f2513971a97/main/control_queue.c#L67-L68

david-marchand: If all mbuf of the control queue are drained, I don't understand this code.

Why are we walking through the icmp queue?
Are not all the mbuf in the control queue referenced in the icmp queue?

rjarry: The icmp queue is separate from the control queue. The problem is we cannot "peek" into the control ring. We must pop mbufs from it and store them into a linked list so that get_icmp_response() can find them. But we may receive them out of order if there are multiple ping sessions in parallel.

So the drain procedure must *first* walk through the icmp linked list and purge any mbufs that belong to the deleted pool, and *then* if the current mbuf (the one just popped from the control ring) is also part of the deleted pool, also purge it.

Does that make sense?

david-marchand: Not sure it makes sense, but ok, I understand we need to do it.

main/control_queue.h
```diff
@@ -6,6 +6,10 @@
 #include <sched.h>
 #include <stdint.h>
 
+// Internal drain event: a pktmbuf pool is about to be freed.
+// drain->obj points to the struct rte_mempool being released.
+#define EVENT_MBUF_POOL_RELEASE UINT32_C(0xbad0caca)
```
david-marchand: The control queue was not aware of what objects are passed so far.
Starting to expose mempool consideration looks wrong (layer violation).

I have no suggestion for now... need to look more at this code.

rjarry: This constant should be defined in mempool.h. I'll change that.

modules/dhcp/control/client.c
```diff
@@ -527,7 +530,7 @@ static void dhcp_fini(struct event_base *) {
 
 static struct module dhcp_module = {
 	.name = "dhcp",
-	.depends_on = "graph,ip_address,iface",
+	.depends_on = "graph,ip_address,iface,control_queue",
```
david-marchand: We were already relying on the control queue.

Why do you need this now?

rjarry: Because now `control_queue_drain()` is called before a mempool is freed. Since the dhcp module allocates a mempool (and releases it on tear down), we must make sure that the control ring is still accessible when the dhcp module is torn down to avoid use-after-free.

david-marchand:
Taking a step back on the mempool release topic.
Something does not add up.


For the case you mention, when a reconfigure event triggers a mempool "re-"creation... the iface itself is valid (I guess no IFACE REMOVE was emitted right?), and so the associated packets and their references in the stack are still valid.
Forcing to drop all those packets is probably not a big deal, but still there is nothing that requires to destroy them.


Besides, we should not be calling rte_mempool_free() if grout still references some mbuf.
I don't see a check on rte_mempool_full() in gr_pktmbuf_pool_release().
And this will likely bite us again.


With such a check, the next problems are likely:
- how to defer this mempool suppression (maybe schedule a once per main thread loop event check? and maybe WARN after 5 minutes ...),
- creating a new mempool may be impacted, as growing a mempool with the same name is only possible if we remove the previous object, here, we may need to add the count of objects in the mempool name,

If we put this in place, I don't think we need a new special mempool release event.

rjarry:
> For the case you mention, when a reconfigure event triggers a mempool "re-"creation... the iface itself is valid (I guess no IFACE REMOVE was emitted right?), and so the associated packets and their references in the stack are still valid. Forcing to drop all those packets is probably not a big deal, but still there is nothing that requires to destroy them.

Changing the datapath CPU affinity cause a reconfiguration of all ports to adjust their TX queue count to match the number of CPUs in the affinity. When reconfiguring (not deleting) a port, if the number of queues changes, we release the mempool we got and request a new one:

https://github.com/DPDK/grout/blob/cde2324fd70597a7813d35884765a9d8f539d805/modules/infra/control/port.c#L131-L139

When you have a single port, gr_pktmbuf_pool_release() will cause the pool to reach 0 reserved mbufs and will free the underlying pool:

https://github.com/DPDK/grout/blob/cde2324fd70597a7813d35884765a9d8f539d805/modules/infra/control/mempool.c#L116-L128

So the in-flight packets are no longer "valid" since we reconfigure all queues with a different mempool. We must make sure these mbufs are dereferenced from all places in grout.

> Besides, we should not be calling rte_mempool_free() if grout still references some mbuf. I don't see a check on rte_mempool_full() in gr_pktmbuf_pool_release(). And this will likely bite us again.

The issue is that the mbufs are referenced by the control plane. Not datapath. So we cannot use a deferred RCU. The cleanup must be done synchronously. Which is what this PR is about.

> With such a check, the next problems are likely:
> 
> * how to defer this mempool suppression (maybe schedule a once per main thread loop event check? and maybe WARN after 5 minutes ...),
> * creating a new mempool may be impacted, as growing a mempool with the same name is only possible if we remove the previous object, here, we may need to add the count of objects in the mempool name,
> 
> If we put this in place, I don't think we need a new special mempool release event.

We already need to purge various places of the control plane structures when objects are deleted (interfaces mostly). I see that mempool release event as a similar mechanism.

rjarry:
In addition of this fix, to avoid mempool free/realloc churn, we could add a gr_pktmbuf_pool_resize() function that check whether we have enough space to pass a bigger (or smaller) pool chunk to the caller. That would eliminate the need to free/realloc when we simply change the number of queues (provided it fits in the available pools).

david-marchand:
> > For the case you mention, when a reconfigure event triggers a mempool "re-"creation... the iface itself is valid (I guess no IFACE REMOVE was emitted right?), and so the associated packets and their references in the stack are still valid. Forcing to drop all those packets is probably not a big deal, but still there is nothing that requires to destroy them.
> 
> Changing the datapath CPU affinity cause a reconfiguration of all ports to adjust their TX queue count to match the number of CPUs in the affinity. When reconfiguring (not deleting) a port, if the number of queues changes, we release the mempool we got and request a new one:
> 
> https://github.com/DPDK/grout/blob/cde2324fd70597a7813d35884765a9d8f539d805/modules/infra/control/port.c#L131-L139
> 
> When you have a single port, gr_pktmbuf_pool_release() will cause the pool to reach 0 reserved mbufs and will free the underlying pool:
> 
> https://github.com/DPDK/grout/blob/cde2324fd70597a7813d35884765a9d8f539d805/modules/infra/control/mempool.c#L116-L128

This check does not cover that the mbufs are still referenced somewhere in grout.

A remaining reference should be a red flag in the first place.
By construction, grout modules should have in place the right event handlers / cleanup mechanisms to clear any mbuf referencing a grout object.

There should not be many references in a grout mbuf... a iface, a nexthop... what else?


> So the in-flight packets are no longer "valid" since we reconfigure all queues with a different mempool. We must make sure these mbufs are dereferenced from all places in grout.

I am not following, there is no reference to a queue in a dpdk mbuf.
Where is it tracked in grout?

Be it on datapath or control path side, I don't see why it would matter to mbuf "validity".

In the icmp case for example, mbufs are probably stored in some icmp queue referencing a L3 iface.
If a l2 port gets reconfigured, the icmp packet is still valid.


> > With such a check, the next problems are likely:
> > 
> > * how to defer this mempool suppression (maybe schedule a once per main thread loop event check? and maybe WARN after 5 minutes ...),
> > * creating a new mempool may be impacted, as growing a mempool with the same name is only possible if we remove the previous object, here, we may need to add the count of objects in the mempool name,
> > 
> > If we put this in place, I don't think we need a new special mempool release event.
> 
> We already need to purge various places of the control plane structures when objects are deleted (interfaces mostly). I see that mempool release event as a similar mechanism.

Yes, with a hammer, everything can be a nail :-).

I posted a draft PR in // since I think we will need it regardless.
I won't insist on my approach if you think the event mechanism is the right way to fix the bug you saw.

rjarry:
> > We already need to purge various places of the control plane structures when objects are deleted (interfaces mostly). I see that mempool release event as a similar mechanism.
> 
> Yes, with a hammer, everything can be a nail :-).
> 
> I posted a draft PR in // since I think we will need it regardless. I won't insist on my approach if you think the event mechanism is the right way to fix the bug you saw.

Your approach seems much simpler and more robust.

I'll test it and take it in.

rjarry:
I have tested and it seems to work. However, there is a drawback with this deferred free approach. There are cases where mbufs may remain in queues for a long time:

1. Held packets in L3 nexthops waiting for resolution. Currently, these can sit in the queue forever. This is a separate bug that I will fix first so that eventually, the pool may be garbage collected.
2. ICMP responses waiting to be collected. Same issue. If nobody collects the responses, they will sit in the queue forever. I will add a timer to remove orphan responses.

With these two fixes, the deferred free will work. I will adjust the timeouts to avoid spurious warnings so that nexthops/icmp-responses have time to age out.

christophefontaine (approved):

PR #675, opened 2026-07-31
mempool: defer free until all mbufs are returned

david-marchand:
Here is a draft, untested.

Such a check should help catch any leak / remaining reference of mbufs in grout, rather than blindly shooting the mempools like we do atm.

Now that I think about it, we could probably add a rte_mempool_full() check and log an ERR in mempool_fini, though I did not do it in this PR.


christophefontaine:
Closing as it was merged as part of #672

PR #676, opened 2026-08-03
Control input cleanup

rjarry:
Depends on #672 

## Summary

- Added per-interface mbuf pools and control-plane file descriptors and events to `struct iface`.
- Reworked loopback and TAP paths to use interface-owned pools and resources.
- Replaced control-input type IDs and allocation flags with direct `rte_edge_t` registration and `(edge, mbuf)` queue entries.
- Added typed send helpers for Ethernet, interface, IPv4, IPv6, ARP, and loopback output.
- Updated DHCP, LACP, ICMP, ICMPv6, ARP, router advertisement, and NDP paths to allocate mbufs from the interface pool and store request context in mbuf metadata.
- Removed shared control-input and per-module mempools, heap-allocated control messages, obsolete output handlers, and related module dependencies.

### DIFF
--- a/modules/infra/control/loopback.c
+++ b/modules/infra/control/loopback.c
@@ -163,8 +163,8 @@ static void iface_loopback_poll(evutil_socket_t, short reason, void *ev_iface) {
 	e->iface = iface;
 	e->domain = ETH_DOMAIN_LOOPBACK;
 
-	if (post_to_stack(loopback_get_control_id(), mbuf) < 0) {
-		LOG(ERR, "post_to_stack: %s", strerror(errno));
+	if (loopback_input_send(mbuf) < 0) {
+		LOG(ERR, "loopback_input_send: %s", strerror(errno));
 		goto err;
 	}
 
--- a/modules/infra/control/loopback.c
+++ b/modules/infra/control/loopback.c
@@ -34,7 +34,6 @@ LOG_TYPE("loopback");
 // TUN device naming pattern: gr-loop{iface_id}
 #define GR_LOOPBACK_TUN_NAME_PATTERN GR_LOOPBACK_TUN_NAME_PREFIX "%d"
 
-static struct rte_mempool *loopback_pool;
 static struct event_base *ev_base;
 
 static void finalize_fd(struct event *ev, void * /*priv*/) {
@@ -120,7 +119,7 @@ static void iface_loopback_poll(evutil_socket_t, short reason, void *ev_iface) {
 		return;
 	}
 
-	mbuf = rte_pktmbuf_alloc(loopback_pool);
+	mbuf = rte_pktmbuf_alloc(iface->pool);
 	if (!mbuf) {
 		LOG(ERR, "rte_pktmbuf_alloc %s", rte_strerror(rte_errno));
 		goto err;
@@ -246,6 +245,13 @@ int iface_loopback_create(struct iface *iface) {
 		goto err;
 	}
 
+	iface->pool = gr_pktmbuf_pool_get(SOCKET_ID_ANY, RTE_GRAPH_BURST_SIZE);
+	if (iface->pool == NULL) {
+		LOG(ERR, "gr_pktmbuf_pool_get: %s", strerror(errno));
+		goto err;
+	}
+	iface->pool_size = RTE_GRAPH_BURST_SIZE;
+
 	iface->cp_ev = event_new(
 		ev_base,
 		iface->cp_fd,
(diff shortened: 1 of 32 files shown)


modules/infra/control/loopback.c
```diff
@@ -254,55 +245,61 @@ int iface_loopback_create(struct iface *iface) {
...
+	}
+	if (iface->cp_fd > 0)
+		close(iface->cp_fd);
+	if (iface->pool != NULL) {
+		gr_pktmbuf_pool_release(iface->pool, iface->pool_size);
+		iface->pool = NULL;
 	}
-	if (lo->fd > 0)
-		close(lo->fd);
 	if (ioctl_sock > 0)
```
aharivel, line 281: ioctl_sock is initialised to -1. 
If socket() is created, then it's 0 (good fd). 
But if a "goto err;" happens after that, the socket is not closed ? 
That's intended ?

modules/ip6/datapath/icmp6_local_send.c
```diff
@@ -50,17 +52,25 @@ int icmp6_local_send(
...
-	msg->ident = ident;
-	msg->hop_limit = hop_limit;
-	msg->dst = *dst;
-	msg->src = nexthop_info_l3(local)->ipv6;
-
-	if ((ret = post_to_stack(ctl_icmp6_request, msg)) < 0) {
-		free(msg);
+
+	d = icmp6_send_mbuf_data(m);
+	d->iface = iface;
```
aharivel, line 64: this is storing the raw pointer... 
(see next comment)

modules/ip6/datapath/icmp6_local_send.c
```diff
@@ -84,32 +94,32 @@ static uint16_t icmp6_local_send_process(
...
-		mbuf->hash.rss = msg->ident;
+		mbuf->hash.rss = msg.ident;
 		mbuf->ol_flags |= RTE_MBUF_F_RX_RSS_HASH;
 
 		payload = PAYLOAD(icmp6_echo);
 		*payload = gr_clock_ns();
 
 		data = ip6_local_mbuf_data(mbuf);
-		data->iface = iface_from_id(msg->iface_id);
+		data->iface = msg.iface;
```
aharivel, line 117: If between post_to_stack() and the worker draining the ring, the interface gets destroyed (user deletes it via API, link goes down, etc.), that pointer points to freed memory...

modules/ip/datapath/icmp_local_send.c
```diff
@@ -99,21 +106,22 @@ static uint16_t icmp_local_send_process(
...
 		data->len = sizeof(*icmp) + sizeof(gr_clock_ns_t);
-		data->dst = msg->dst;
-		data->src = msg->src;
-		data->vrf_id = msg->vrf_id;
-		data->ttl = msg->ttl;
+		data->dst = msg.dst;
+		data->src = msg.src;
+		data->vrf_id = msg.vrf_id;
+		data->ttl = msg.ttl;
+		data->iface = msg.iface;
```
aharivel, line 124: same as in icmp6_local_send.c : 
If between post_to_stack() and the worker draining the ring, the interface gets destroyed (user deletes it via API, link goes down, etc.), that pointer points to freed memory...

aharivel: If grout guarantees that interface destruction always drains the control_input ring first (via rte_rcu_qsbr_synchronize() or similar barrier), then the pointer can't go stale.
Worth checking IMO

aharivel (changes requested):
Some minor check before approve

modules/infra/control/loopback.c
rjarry: I am not following. If the socket() call succeeds (first call of the function), ioctl_sock will be >= 0 and it will be closed.

modules/ip/datapath/icmp_local_send.c
rjarry: Yes, we use RCU sync:

https://github.com/DPDK/grout/blob/670bd537783e8156bf8842a50e13639398b67daa/modules/infra/control/iface.c#L702-L717

modules/ip6/datapath/icmp6_local_send.c
rjarry: The flow is:

1. Control plane thread handles ICMP6_SEND, looks up interface pointer, calls `icmp6_local_send()` with that pointer. The pointer gets attached to an mbuf.
2. Datapath thread picks up mbuf, pushes it through graph
3. Control plane thread handles IFACE_DELETE, unreferences interface ID and waits until all datapath threads have "seen" the deletion.

At this point, no mbuf can reference the deleted interface pointer which can be safely freed.

modules/infra/control/loopback.c
aharivel: I think it should be ` if (ioctl_sock >= 0)` because socket() can return 0. 
Ok this would means that fd 0 (stdin) would be already closes when this happens.. unlikely

aharivel (approved):
LGTM - every questions has been answered

PR #679, opened 2026-08-03
infra: use IFA_F_NODAD for control plane tap addresses

rjarry:
The accept_dad sysctl set on the control plane tap is ineffective because the kernel evaluates max(conf/all/accept_dad, conf/<iface>/accept_dad). Since conf/all/accept_dad defaults to 1, DAD still runs after link flaps.

When DAD runs, the kernel sends a NS probe on the control plane tap. Grout's datapath answers with a NA (it owns the same address), and the kernel marks the address as dadfailed. The address never recovers, sendmsg() returns EINVAL permanently and OSPF6 never converges.

Set IFA_F_NODAD in the netlink RTM_NEWADDR message when adding IPv6 addresses to the control plane tap. This is a per-address flag that unconditionally prevents DAD regardless of any sysctl setting.

Fixes: 0480bdffd3c2 ("infra: disable DAD on control plane TAPs")

### DIFF
--- a/modules/infra/control/ctlplane.c
+++ b/modules/infra/control/ctlplane.c
@@ -327,12 +327,6 @@ static void cp_create(struct iface *iface) {
 	// would then fail to bind() sockets on those addresses.
 	sysctl_write("net/ipv6/conf", iface, "keep_addr_on_down", "1");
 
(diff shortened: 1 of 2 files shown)


aharivel (approved):
LGTM

PR #680, opened 2026-08-04
icmp,icmp6: replace response queue with hash-based session pool

rjarry:
The IPv4 and IPv6 ICMP ping handlers share the same pattern: a linked list with linear scan and a mempool to match responses to requests. This series first fixes several issues in the ICMP response parsing (IP options, length validation, timestamp handling, use of ip_local_mbuf_data) then extracts the common logic into a reusable icmp_session_pool backed by rte_hash for O(1) lookup by (ident, seq_num).

The session pool uses a create-before-send model: the send handler registers a session, and incoming responses are matched and stored by a get_key callback. A periodic GC timer reaps stale sessions. Both IPv4 and IPv6 modules are converted to use it, removing significant code duplication.

## Summary

- Added a reusable `icmp_session_pool` backed by `rte_hash` for O(1) lookup by `(ident, seq_num)`.
- Added session lifecycle APIs for registration, deletion, response retrieval, timeout cleanup, and interface removal.
- Replaced separate IPv4 and IPv6 linked-list queues, mempools, and garbage-collection timers with the shared session pool.
- Registered sessions before transmission and removed them when transmission fails.
- Added IPv4 and IPv6 response parsing for embedded echo requests and ICMP errors.
- Validated embedded packet lengths, protocols, ICMP types, and IPv4 options.
- Preserved IPv6 packets and corrected outer-header and timestamp handling.

### DIFF
--- a/modules/infra/control/icmp_session.c
+++ b/modules/infra/control/icmp_session.c
@@ -1,0 +1,239 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026 Robin Jarry
+
+#include "icmp_session.h"
+
+#include <gr_clock.h>
+#include <gr_errno.h>
+#include <gr_infra.h>
+
+#include <event2/event.h>
+#include <rte_errno.h>
+#include <rte_hash.h>
+#include <rte_malloc.h>
+#include <rte_mbuf.h>
+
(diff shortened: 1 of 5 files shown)


aharivel (approved):
LGTM

PR #690, opened 2026-08-10
frr,l2: fix ipv4-mapped ipv6 nexthop behavior

chdxD1:
With #687 I did not really change the behaviour of sending unmapped IPv4-addresses for IPv6 routes down to the datapath, however I identified another bug.

When the nexthop is unmapped, it shares a l3_lookup key. This is unexpected with NHG / NHE id handling together with zebra, leading to a loss of mapping between zebra and grout.

This reverts one of the commits (https://github.com/DPDK/grout/commit/6aaf92bf6998f93273458076a8de403c64c83fd8) of #687, thus fixing a behaviour introduced by https://github.com/DPDK/grout/commit/a1d26c93fa94eaf5432aa7eafb16bc840052b2a4.

Because IPv4-mapped IPv6 addresses are now being passed to the FIB, the datapath now correctly identifies these addresses and uses the IPv4 family now.

In addition a smoke test is provided, which covers this.

### DIFF
--- a/frr/zebra_dplane_grout.c
+++ b/frr/zebra_dplane_grout.c
@@ -100,14 +100,6 @@ void ipaddr_to_l3_addr(struct l3_addr *dst, const struct ipaddr *src) {
 		memcpy(&dst->ipv4, &src->ipaddr_v4, sizeof(dst->ipv4));
 		break;
 	case IPADDR_V6:
-		if (IS_MAPPED_IPV6(&src->ipaddr_v6)) {
-			struct in_addr v4;
-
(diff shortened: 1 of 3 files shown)


chdxD1:
Trying a force-push with the same commits - imo the CI failures are flakes but I can't retrigger CI

rjarry:
@chdxD1 yes, I am trying to fix these in #692 

Could you remove the empty line between `Fixes:` and your `Signed-off-by:` trailers in the first commit?

Thanks!

rjarry (approved):

PR #692, opened 2026-08-11
Control plane netlink fixes

rjarry:


rjarry:
@zeeke fyi

maxime-leroy:
The last fix reminds me of this FRR PR:
https://github.com/FRRouting/frr/pull/19564

Could we check whether this FRR change introduces any regression in grout, and whether an additional fix is needed on the grout side?

rjarry:
> The last fix reminds me of this FRR PR: [FRRouting/frr#19564](https://github.com/FRRouting/frr/pull/19564)
> 
> Could we check whether this FRR change introduces any regression in grout, and whether an additional fix is needed on the grout side?

Interesting that we hit the same issue :)

In this case, it would mean that deleting the last IPv4 address in grout would trigger zebra to wrongfully assume all routes "attached to that interface" (whatever that means in the context of grout) are deleted silently without sending any notification.

This is a separate problem. I'll do some digging.

christophefontaine (approved):

Issue #693, opened 2026-08-12
SRv6 packet gets truncated


zeeke:
### Description

SRv6 packets encapsulatinug IPv6 packets arrive with invalid state at the end of the SRv6 tunnel.
The following is the packet trace going through grout, leafkind1(kernel-FRR), spine(kernel-FRR), leafSRV6(kernel-FRR), hostSRV6_red

```

--------- grout ---------

NOTICE: TRACE: [rx pe-s-100] a6:78:57:4f:2d:f6 > a6:6f:92:41:e9:87 / IPv6 2001:db8:1::4 > 2001:db8:170:20::2 ttl=63 proto=TCP(6), (pkt_len=94)

--------- 07:53:27.092220708 cpu 4 ---------
port_rx-p3q0: RX_L4_CKSUM_GOOD RX_IP_CKSUM_UNKNOWN RX_OUTER_L4_CKSUM_UNKNOWN TX_L4_NO_CKSUM
iface_input: iface=pe-s-100 mode=VRF
eth_input: a6:78:57:4f:2d:f6 > a6:6f:92:41:e9:87 type=IPv6(0x86dd)
ip6_input: 2001:db8:1::4 > 2001:db8:170:20::2 ttl=63 proto=TCP(6)
ip6_forward:
ip6_output: 2001:db8:1::4 > 2001:db8:170:20::2 ttl=62 proto=TCP(6)
sr6_output: match=0.0.0.0/0
ip6_output: 2001:db8:11::5 > fd00:0:10:e001:: ttl=255 proto=RouteOpts(43)
eth_output: aa:c1:ab:c0:b6:4b > aa:c1:ab:5b:cf:b0 type=IPv6(0x86dd)
iface_output: iface=u_toswitch1
port_output:
port_tx-p0q0: shared RX_L4_CKSUM_GOOD RX_IP_CKSUM_UNKNOWN RX_OUTER_L4_CKSUM_UNKNOWN TX_L4_NO_CKSUM

NOTICE: TRACE: [tx u_toswitch1] aa:c1:ab:c0:b6:4b > aa:c1:ab:5b:cf:b0 / IPv6 2001:db8:11::5 > fd00:0:10:e001:: ttl=255 proto=RouteOpts(43) Ext(43 len=24), (pkt_len=158)

--------- leafkind1 ---------
--- not captured

--------- spine ---------
[...]

PR #694, opened 2026-08-12
srv6: fix IPv6 encapsulation payload length

zeeke:
IPv6 payload_len does not include the header itself, unlike IPv4 total_length. The SRv6 output node was using the inner IPv6 payload_len without adding the header size, resulting in a truncated outer payload length for IPv6-in-IPv6 encapsulation.

Fixes: 7d88a14435879 ("srv6: add srv6 module")

Fixes:
- https://github.com/DPDK/grout/issues/693

### DIFF
--- a/modules/srv6/datapath/srv6_output.c
+++ b/modules/srv6/datapath/srv6_output.c
@@ -84,7 +84,7 @@ srv6_output_process(struct rte_graph *graph, struct rte_node *node, void **objs,
 				t->is_dest6 = true;
 			}
 			inner_ip6 = rte_pktmbuf_mtod(m, struct rte_ipv6_hdr *);
-			plen = rte_be_to_cpu_16(inner_ip6->payload_len);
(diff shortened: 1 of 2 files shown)


rjarry:
Nice, can you also include the smoke test updates to make sure we exercise that ipv6-in-ipv6 encap/decap code?

zeeke:
> Nice, can you also include the smoke test updates to make sure we exercise that ipv6-in-ipv6 encap/decap code?

Added an extension to smoek/srv6_test.sh . Let me know if you prefer a brand new file for this test

rjarry:
The existing SRv6 tests are perfect. Your previous implementation https://github.com/zeeke/grout/commit/4269f6193ea10404375df7c098104f6e1c6407a5 checked that the packet sizes were correct (not truncated). Could you use that?

rjarry (approved):

PR #697, opened 2026-08-14
infra: make all sizing parameters configurable via env vars

rjarry:
When running with --test-mode, grout allocates 2048 MB of memory and uses production-sized data structures, resulting in ~887 MB RSS per smoke test instance. This makes it impractical to run many tests in parallel.

All daemon settings and data structure sizing parameters are now configurable via environment variables. CLI flags override environment variables when both are set. Boolean variables accept 1/true/on/yes or 0/false/off/no. Invalid values cause a startup error.

I have tested all smoke tests with the following:

| Resource                  | Default           | Test mode
| ------------------------- | ----------------- | -------------
| EAL memory                | 2048              | 512
| GROUT_MAX_IFACES          | 1024              | 64
| GROUT_MEMPOOL_CHUNK_SIZE  | 64k               | 2047
| GROUT_MAX_NEXTHOPS        | 128k              | 1024
| GROUT_MAX_ROUTES          | 64k per VRF       | 512
| GROUT_MAX_FDB_ENTRIES     | 4k                | 128
| GROUT_MAX_CONNTRACKS      | 16k               | 32
| GROUT_PORT_QUEUE_SIZE     | device default    | 32
| GROUT_FIB4_ALGORITHM      | DIR24_8           | DUMMY
| GROUT_FIB6_ALGORITHM      | TRIE              | DUMMY

Measured RSS after the changes drops from ~887 MB to ~42 MB per test.

@zeeke fyi


main/dpdk.c
```diff
@@ -160,9 +160,9 @@ int dpdk_init(void) {
 		if (!has_m) {
 			vec_add(eal_args, "-m");
 			if (gr_config.max_mtu > 2048)
-				vec_add(eal_args, "4096");
+				vec_add(eal_args, "512");
 			else
-				vec_add(eal_args, "2048");
+				vec_add(eal_args, "256");
```
christophefontaine: shouldn't we reduce the reserved memory only when test-mode is enabled ?
Also, with 1 GB hugepages, is it relevant to reserve less than 1 GB ?

As we configure the memory based on the MTU, should we expose all the knobs that are modified in test mode, instead of fixed values ?

modules/ip/control/route.c
```diff
@@ -62,15 +63,17 @@ static struct rte_fib *get_fib(uint16_t vrf_id) {
...
-			.num_tbl8 = fib4_get_num_tbl8(vrf),
-		},
 	};
+	if (gr_config.test_mode) {
+		conf.type = RTE_FIB_DUMMY;
+	} else {
+		conf.type = RTE_FIB_DIR24_8;
+		conf.dir24_8.nh_sz = RTE_FIB_DIR24_8_8B;
+		conf.dir24_8.num_tbl8 = fib4_get_num_tbl8(vrf);
+	}
```
christophefontaine: While the other changes only impact the performances, does this modify the behavior as we're not exercising the same FIB code ?

modules/ip6/control/route.c
```diff
@@ -65,15 +66,17 @@ static struct rte_fib6 *get_fib6(uint16_t vrf_id) {
...
-			.num_tbl8 = fib6_get_num_tbl8(vrf),
-		},
 	};
+	if (gr_config.test_mode) {
+		conf.type = RTE_FIB6_DUMMY;
+	} else {
+		conf.type = RTE_FIB6_TRIE;
+		conf.trie.nh_sz = RTE_FIB6_TRIE_8B;
+		conf.trie.num_tbl8 = fib6_get_num_tbl8(vrf);
+	}
```
christophefontaine: Same as IPv4

main/dpdk.c
rjarry: This code will only run if --test-mode is specified. So there are no huge pages involved.

Most of the knobs are already exposed. This patch is only changing their default value in test mode. You can still update most of them using grcli commands. The rare exceptions are:

| knob                              | default | test mode |
|----------------------------------|------|----------- |
| mbuf pool chunk size | 64k | 1k |
| per-interface ctlplane mbuf pool size | 256 | 32 |

modules/ip/control/route.c
rjarry: In fact, it does not use the FIB at all when type=DUMMY. It only relies on the internal RIB object which does not need to pre-allocate a 2^24 array of 8 bytes nexthop pointers (128MB of fixed allocation per FIB, so 256MB per VRF including IPv4 and IPv6).

The behavior itself does not change. The APIs are the same, only the internal backend is different. I don't expect any surprises in that area.

smoke/_init.sh
```diff
@@ -378,6 +378,15 @@ if [ "$run_grout" = true ]; then
...
 		grout_extra_options+=" -t"
+		smoke_setenv GROUT_MAX_IFACES 64
+		smoke_setenv GROUT_MEMPOOL_CHUNK_SIZE 2047
+		smoke_setenv GROUT_MAX_NEXTHOPS 1024
+		smoke_setenv GROUT_MAX_ROUTES 512
+		smoke_setenv GROUT_MAX_FDB_ENTRIES 128
+		smoke_setenv GROUT_MAX_CONNTRACKS 32
+		smoke_setenv GROUT_PORT_QUEUE_SIZE 32
+		smoke_setenv GROUT_FIB4_ALGORITHM DUMMY
+		smoke_setenv GROUT_FIB6_ALGORITHM DUMMY
```
christophefontaine: For the smoke tests, shouldn't we keep the TRIE algorithm ?

smoke/_init.sh
```diff
@@ -378,6 +378,15 @@ if [ "$run_grout" = true ]; then
 	smoke_setenv ASAN_OPTIONS disable_coredump=0
 	if [ "$use_hardware_ports" = false ]; then
 		grout_extra_options+=" -t"
+		smoke_setenv GROUT_MAX_IFACES 64
+		smoke_setenv GROUT_MEMPOOL_CHUNK_SIZE 2047
+		smoke_setenv GROUT_MAX_NEXTHOPS 1024
```
christophefontaine: If we want to mimic the remove line 431, GROUT_MAX_NEXTHOPS should be set to 128.

smoke/fib4_fullview_manualtest.sh
```diff
@@ -11,6 +11,8 @@ follow_events=hide
 # Simulate a full IPv4 BGP view: ~1M prefixes.
 count=1000000
 
+grcli nexthop config set max 4096
+
```
christophefontaine: Why do we need to tune this now ?

smoke/_init.sh
rjarry: We can, I had set it to verify the memory reduction. I'll remove these two lines.

smoke/_init.sh
rjarry: Indeed.

smoke/fib4_fullview_manualtest.sh
rjarry: I think this test was already failing since f51c7f085408295147cc23cd6e5c1d0b1a91c9e8. But since we don't run it in the CI, we missed it.

rjarry:
I have tested both `smoke/fib{4,6}_fullview_manualtest.sh` and they work. However, `smoke/fullview_frr_manualtest.sh` fails but because of a gobgp/frr crash.

```
+ ip netns exec bgp-peer gobgp mrt inject global build//latest-bview
failed to parse: unknown evpn subtype: 249
ZEBRA: [QH9AB-Y4XMZ][EC 100663314] STARVATION: task dplane_thread_loop (44d4b0) ran for 33713ms (cpu time 210ms)
BGP: [QH9AB-Y4XMZ][EC 100663314] STARVATION: task bgp_process_packet (49f133) ran for 6135ms (cpu time 521ms)
BGP: [VQK5M-9G0V1][EC 100663315] CPU starvation: {(event *)0x7ffe38cd2550 arg=0x10722440 ready               (bgp_generate_updgrp_packets)() (&connection->t_generate_updgrp_packets) from /home/rjarry/upstream/grout/subprojects/frr-frr-10.6.1/bgpd/bgp_io.c:169} getting executed 8062ms late, warning threshold 4000ms. System load: 68.01, 21.04, 8.82
ZEBRA: [VQK5M-9G0V1][EC 100663315] CPU starvation: {(event *)0x7fffbc326650 arg=0x32bf3e40 ready               wheel_timer_thread() &wheel->timer from /home/rjarry/upstream/grout/subprojects/frr-frr-10.6.1/lib/wheel.c:50} getting executed 10166ms late, warning threshold 4000ms. System load: 68.01, 21.04, 8.82
failed to send: EOF

real	9m11.075s
user	22m54.887s
sys	4m18.091s
smoke/fullview_frr_manualtest.sh: line 1: 1352045 Killed                     ip netns exec bgp-peer gobgpd -f $1/gobgpd.conf

```

smoke/fullview_frr_manualtest.sh
```diff
@@ -40,11 +40,7 @@ netns_add bgp-peer
 move_to_netns x-p0 bgp-peer
 ip -n bgp-peer addr add 172.16.0.2/24 dev x-p0
 
-
-set_ip_address p0 172.16.0.1/24
-set_ip_route 0.0.0.0/0 172.16.0.2
-
-grcli route config set vrf main  rib4-routes 1000000
+grcli interface set vrf main rib4-routes 1000000
```
christophefontaine, line 43: Should we have `grout_max_nexthops=4096` as well ? Would this fix the crash ?

christophefontaine (approved):

smoke/fullview_frr_manualtest.sh
rjarry: I don't think it would. There is a single nexthop in this whole test. The crash seems unrelated.

maxime-leroy:
Why env vars ? Not a file configuration or argc/argv ?

rjarry:
> Why env vars ? Not a file configuration or argc/argv ?

Because it is much simpler, there is nothing to parse :)

Also, it follows the existing variables we already handle and is easy to use along with systemd `EnvironmentFile=` or `podman --env-file`

https://github.com/DPDK/grout/blob/a5f2b8e2ce5a6adb881447faa36ec2cd224bd6fd/main/grout.service#L12

https://github.com/DPDK/grout/blob/a5f2b8e2ce5a6adb881447faa36ec2cd224bd6fd/main/grout.default#L8-L17

Issue #698, opened 2026-08-17
EVPN all-active multihoming support (RFC 7432)


hcaldicott:
### Description

Grout already supports single-homed EVPN/VXLAN bridging with FRR (evpn_vxlan_frr_test.sh). This issue proposes extending that to all-active EVPN multihoming (RFC 7432 / RFC 8365): two grout PEs presenting one 802.3ad LAG to a multihomed device, with FRR providing the EVPN-MH control plane.

We have a working prototype of the complete feature (validated in a three-node namespace lab with a split-chassis LACP peer, including IPv6 underlays, daemon restarts and failure injection) and intend to upstream it as a series of PRs. I am filing this issue so the individual PRs, including bugfixes in existing code that we found along the way, can link back to it.

### Context and Use Cases

A device (host, VM host, or access switch) is attached with a LAG whose members land on two different grout instances. Both PEs forward simultaneously (all-active): remote PEs load-balance unicast toward the segment across both VTEPs, BUM traffic is forwarded onto the segment by exactly one elected PE, and traffic received from the segment is never reflected back to it through the other PE.

### Division of labour
FRR owns BGP EVPN, Ethernet Segment state, Designated Forwarder election and route signalling (Type-1/Type-4). Grout owns LACP, VXLAN, bridging and all packet forwarding; the ES-facing ports stay bound to DPDK.

### Required grout changes

**1. L2 VTEP nexthops and L2 nexthop groups.** A new L2-VTEP nexthop type, reusing the existing RCU-safe weighted group representation, plus an optional NHG ID on external FDB entries. The bridge datapath resolves the group by ID and selects a member per flow, giving MAC ECMP / aliasing toward a remote all-active segment. Mixed L2/L3 groups and cross-type route/FDB references are rejected; deletes are type/origin-conditional so a rejected install cannot later remove another owner's object.

**2. Bridge-port policy for DF and split horizon.** Consume DPLANE_OP_BR_PORT_UPDATE in the FRR provider and apply it atomically in the bridge/VXLAN graph: suppress overlay BUM on a non-DF port, filter unicast and BUM received from an ES peer VTEP, and redirect a local-ES hairpin through FRR's backup NHG (local bias). State is visible via grcli bridge-port show.
 
[...]

PR #699, opened 2026-08-17
bond: preserve configured MAC when attaching members

hcaldicott:
Attaching a member to a bond unconditionally re-derives the bond MAC via
`bond_mac_set(iface, &zero)`, which discards an explicitly configured bond
MAC as soon as the first member arrives, and later members never get the
configured address programmed into their filters.

Keep an explicitly configured MAC when members are attached and program it
into each new member's filter instead. When no MAC was configured, the
existing re-derivation behaviour is unchanged.

Found while prototyping EVPN all-active multihoming (#698), where two
independent LACP instances must present one shared system identity, but the
bug is reproducible on any single bond.

## Testing

- Extended `smoke/bond_lacp_test.sh` to configure an explicit bond MAC and
  assert it survives all member attachments.
- Without the fix (test-only applied to `main`): the smoke test fails —
  after attach the bond MAC becomes the member-derived address
  (`02:09:e8:d0:db:1c`) instead of the configured `02:00:00:00:00:10`.
- With the fix: full unit suite and smoke suite pass (AlmaLinux 9
  container, arm64).

Related: #698

### DIFF
--- a/modules/infra/control/bond.c
+++ b/modules/infra/control/bond.c
@@ -172,9 +172,14 @@ static int bond_attach_member(struct iface *iface, struct iface *member) {
 	m->iface = member;
 	bond->n_members++;
 
-	if (bond_mac_set(iface, &zero) < 0) {
+	if (!bond->mac_explicit) {
+		if (bond_mac_set(iface, &zero) < 0) {
+			bond->n_members--;
+			return errno_log(errno, "bond_mac_set");
+		}
+	} else if (iface_add_eth_addr(member, &bond->mac) < 0) {
 		bond->n_members--;
-		return errno_log(errno, "bond_mac_set");
+		return errno_log(errno, "iface_add_eth_addr(member)");
 	}
 
 	bond_update_active_members(iface);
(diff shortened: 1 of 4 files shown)


rjarry:
Hi Harrison,

thanks a lot for all your contributions.

> Found while prototyping EVPN all-active multihoming (#698), where two independent LACP instances must present one shared system identity, but the bug is reproducible on any single bond.

I am curious about how you managed to test this with grout. Zebra currently (10.8-dev) lacks the dplane API abstractions for multihoming EVPN and does direct linux netlink access instead of honoring the registered dplane provider (e.g. grout).

rjarry (approved):

rjarry (changes requested):
Sorry, I forgot to ask you to remove the `Assisted-by:` trailers.

hcaldicott:
Hiya Robin,

I've started removing these trailers - thanks!

> I am curious about how you managed to test this with grout. Zebra currently (10.8-dev) lacks the dplane API abstractions for multihoming EVPN and does direct linux netlink access instead of honoring the registered dplane provider (e.g. grout).

You're 100% right about this - vanilla FRR doesn't work correctly at the moment. I've been developing with a fork/patch to implement via dplane, which I'm planning to upstream. On the current prototype code, I have a small patch to 10.6.1. See here: [github.com/hcaldicott/grout/tree/evpn-mh-protodown/subprojects/packagefiles/frr](https://github.com/hcaldicott/grout/tree/evpn-mh-protodown/subprojects/packagefiles/frr)

That said, this particular fix doesn't depend on any of that: the bond MAC behavior reproduces with plain grcli on a single bond, which is what the smoke test here does — the MH work is just where I uncovered the problem.

hcaldicott:
Further to this, looks like @christophefontaine has sorted out the FRR issue for me: https://github.com/FRRouting/frr/pull/23056

PR #701, opened 2026-08-17
infra,l2,ip: fix flow hashing for ports without RSS

hcaldicott:
Several datapaths read `m->hash.rss` directly for ECMP and load-balancing
decisions (fib4/fib6 lookups, vxlan underlay routing and source port
selection). On ports without RSS — TAP and other virtual devices — that
field holds stale or zero data, so every flow shares one value: ECMP
collapses onto a single nexthop and all vxlan flows share one UDP source
port. After SRv6 decapsulation, the outer flow label is copied over the
hash while a hardware valid flag survives; with a zero label (as grout's
own encap emits) every inner flow shares hash 0.

The series first extracts the bond Toeplitz hashing into a shared helper
and fixes two fragment-handling bugs in it: packets with the DF flag set
were treated as fragments and hashed L3-only, and first fragments (offset
zero, MF set) were hashed with their L4 ports unlike the rest of their
datagram. It then caches one canonical per-packet hash by storing the
software result in `m->hash.rss` and marking it valid with
`RTE_MBUF_F_RX_RSS_HASH`, exactly as a hardware driver would — no new
mbuf metadata. The bond, vxlan and L3 consumers read it through one
helper. vxlan decapsulation leaves the outer hash in place: a VTEP
following RFC 7348 derives the outer UDP source port from the inner
flow, so an outer RSS value is already a function of the inner flow and
stays valid after decapsulation. SRv6 decapsulation keeps using a
non-zero outer flow label as the flow entropy, now stored through the
shared helper so the valid flag is coherent, and invalidates the hash
when the label is zero so the first consumer hashes the inner packet.

For packets whose driver does provide RSS, the datapath is unchanged:
the same flag test and the same `m->hash.rss` read as before the series,
with no recomputation after encap or decap.

Found while prototyping EVPN all-active multihoming (#698), where bridged
flows must keep per-flow path affinity across bond and vxlan ECMP, but
everything above is reproducible on plain upstream.

## Testing

- New flow_hash unit tests: per-flow stability, DF/MF fragment handling,
  single computation and caching, hardware RSS precedence, hash
  invalidation.
- New srv6 decap unit test: a non-zero outer flow label becomes the flow
  hash with the valid flag set; a zero label drops a stale outer RSS
  value.
- Extended `smoke/vxlan_test.sh`: a second bridged port injects 32
  distinct UDP flows and the test asserts the encapsulated packets use
  several distinct source ports.
- Extended `smoke/ip_loadbalance_test.sh`: 64 distinct UDP flows through
  the ECMP route, asserting both group members carry traffic.
- Without the fixes (tests-only applied to `main`): the vxlan test fails
  with "32 flows shared 1 vxlan source ports", the load-balance test
  fails with "member p1 carried 2 of 64 distinct flows", the MF fragment
  unit test fails, and the srv6 decap test fails with the stale outer
  hash still marked valid.
- With the series: full unit suite and the vxlan, vxlan6 and
  ip_loadbalance smoke tests pass (AlmaLinux 9 container, arm64).

Related: #698

### DIFF
--- a/modules/infra/datapath/flow_hash.c
+++ b/modules/infra/datapath/flow_hash.c
@@ -1,0 +1,145 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026 Harrison Caldicott
+
+#include "flow_hash.h"
+#include "mbuf.h"
+
+#include <gr_macro.h>
+
+#include <rte_ether.h>
+#include <rte_ip4.h>
+#include <rte_ip6.h>
+#include <rte_tcp.h>
+#include <rte_thash.h>
+#include <rte_udp.h>
+
+static const uint8_t rss_key[] = {
+	0x6d, 0x5a, 0x56, 0xda, 0x25, 0x5b, 0x0e, 0xc2, 0x41, 0x67, 0x25, 0x3d, 0x43, 0xa3,
+	0x8f, 0xb0, 0xd0, 0xca, 0x2b, 0xcb, 0xae, 0x7b, 0x30, 0xb4, 0x77, 0xcb, 0x2d, 0xa3,
+	0x80, 0x30, 0xf2, 0x0c, 0x6a, 0x42, 0xb7, 0x3b, 0xbe, 0xac, 0x01, 0xfa,
+};
+
+static uint32_t flow_hash_l3(const struct rte_mbuf *m, uint32_t l3_offset, rte_be16_t eth_type) {
+	union {
+		uint32_t u32;
+		struct rte_ipv4_tuple v4;
+		struct rte_ipv6_tuple v6;
+	} tuple = {};
+	union {
+		const struct rte_ipv4_hdr *ip4;
+		const struct rte_ipv6_hdr *ip6;
+	} l3;
+	union {
+		const struct rte_udp_hdr *udp;
+		const struct rte_tcp_hdr *tcp;
+	} l4;
+	uint32_t len;
+
+	switch (eth_type) {
+	case RTE_BE16(RTE_ETHER_TYPE_IPV4):
+		l3.ip4 = rte_pktmbuf_mtod_offset(m, const struct rte_ipv4_hdr *, l3_offset);
+		tuple.v4.src_addr = l3.ip4->src_addr;
+		tuple.v4.dst_addr = l3.ip4->dst_addr;
+		switch (l3.ip4->next_proto_id) {
+		case IPPROTO_UDP:
+			if (l3.ip4->fragment_offset == 0) {
+				l4.udp = rte_pktmbuf_mtod_offset(
+					m,
+					const struct rte_udp_hdr *,
+					l3_offset + rte_ipv4_hdr_len(l3.ip4)
+				);
+				tuple.v4.sport = l4.udp->src_port;
+				tuple.v4.dport = l4.udp->dst_port;
+			} else {
+				tuple.v4.sport = 0;
+				tuple.v4.dport = 0;
+			}
+			break;
+		case IPPROTO_TCP:
+			if (l3.ip4->fragment_offset == 0) {
+				l4.tcp = rte_pktmbuf_mtod_offset(
+					m,
+					const struct rte_tcp_hdr *,
+					l3_offset + rte_ipv4_hdr_len(l3.ip4)
+				);
+				tuple.v4.sport = l4.tcp->src_port;
+				tuple.v4.dport = l4.tcp->dst_port;
+			} else {
+				tuple.v4.sport = 0;
+				tuple.v4.dport = 0;
+			}
+			break;
+		default:
+			tuple.v4.sport = 0;
+			tuple.v4.dport = 0;
+		}
+		len = sizeof(tuple.v4);
+		break;
+	case RTE_BE16(RTE_ETHER_TYPE_IPV6):
+		l3.ip6 = rte_pktmbuf_mtod_offset(m, const struct rte_ipv6_hdr *, l3_offset);
+		tuple.v6.src_addr = l3.ip6->src_addr;
+		tuple.v6.dst_addr = l3.ip6->dst_addr;
+		switch (l3.ip6->proto) {
+		case IPPROTO_UDP:
+			l4.udp = rte_pktmbuf_mtod_offset(
+				m, const struct rte_udp_hdr *, l3_offset + sizeof(*l3.ip6)
+			);
+			tuple.v6.sport = l4.udp->src_port;
+			tuple.v6.dport = l4.udp->dst_port;
+			break;
+		case IPPROTO_TCP:
+			l4.tcp = rte_pktmbuf_mtod_offset(
+				m, const struct rte_tcp_hdr *, l3_offset + sizeof(*l3.ip6)
+			);
+			tuple.v6.sport = l4.tcp->src_port;
+			tuple.v6.dport = l4.tcp->dst_port;
+			break;
+		default:
+			tuple.v6.sport = 0;
+			tuple.v6.dport = 0;
+		}
+		len = sizeof(tuple.v6);
+		break;
+	default:
+		return 0;
+	}
+
+	return rte_softrss_be(&tuple.u32, len / sizeof(uint32_t), rss_key);
+}
+
+uint32_t gr_mbuf_flow_hash(const struct rte_mbuf *m, gr_mbuf_flow_hash_mode_t mode) {
+	union {
+		uint32_t u32;
+		struct {
+			struct rte_ether_addr mac;
+			rte_be16_t vlan_id;
+		} l2;
+	} tuple;
+	const struct rte_ether_hdr *eth;
+	const struct rte_vlan_hdr *vlan;
+	uint32_t l3_offset;
+	rte_be16_t eth_type;
+
+	if (mode == GR_MBUF_FLOW_HASH_RSS && (m->ol_flags & RTE_MBUF_F_RX_RSS_HASH))
+		return m->hash.rss;
+
+	eth = rte_pktmbuf_mtod(m, const struct rte_ether_hdr *);
+	tuple.l2.mac = eth->dst_addr;
+	if (eth->ether_type == RTE_BE16(RTE_ETHER_TYPE_VLAN)) {
+		vlan = PAYLOAD(eth);
+		tuple.l2.vlan_id = vlan->vlan_tci;
+		eth_type = vlan->eth_proto;
+		l3_offset = sizeof(*eth) + sizeof(*vlan);
+	} else {
+		tuple.l2.vlan_id = 0;
+		eth_type = eth->ether_type;
+		l3_offset = sizeof(*eth);
+	}
+
+	if (mode == GR_MBUF_FLOW_HASH_L2
+	    || (eth_type != RTE_BE16(RTE_ETHER_TYPE_IPV4)
+		&& eth_type != RTE_BE16(RTE_ETHER_TYPE_IPV6)))
+		return rte_softrss_be(&tuple.u32, sizeof(tuple.l2) / sizeof(uint32_t), rss_key);
+
+	return flow_hash_l3(m, l3_offset, eth_type);
+}
--- a/modules/infra/datapath/flow_hash.c
+++ b/modules/infra/datapath/flow_hash.c
@@ -42,7 +42,8 @@ static uint32_t flow_hash_l3(const struct rte_mbuf *m, uint32_t l3_offset, rte_b
 		tuple.v4.dst_addr = l3.ip4->dst_addr;
 		switch (l3.ip4->next_proto_id) {
 		case IPPROTO_UDP:
-			if (l3.ip4->fragment_offset == 0) {
+			if ((rte_be_to_cpu_16(l3.ip4->fragment_offset) & RTE_IPV4_HDR_OFFSET_MASK)
+			    == 0) {
 				l4.udp = rte_pktmbuf_mtod_offset(
 					m,
 					const struct rte_udp_hdr *,
@@ -56,7 +57,8 @@ static uint32_t flow_hash_l3(const struct rte_mbuf *m, uint32_t l3_offset, rte_b
 			}
 			break;
 		case IPPROTO_TCP:
-			if (l3.ip4->fragment_offset == 0) {
+			if ((rte_be_to_cpu_16(l3.ip4->fragment_offset) & RTE_IPV4_HDR_OFFSET_MASK)
+			    == 0) {
 				l4.tcp = rte_pktmbuf_mtod_offset(
 					m,
 					const struct rte_tcp_hdr *,
--- a/modules/infra/datapath/flow_hash.c
+++ b/modules/infra/datapath/flow_hash.c
@@ -147,3 +147,7 @@ uint32_t gr_mbuf_flow_hash(const struct rte_mbuf *m, gr_mbuf_flow_hash_mode_t mo
 
 	return flow_hash_l3(m, l3_offset, eth_type);
 }
+
+uint32_t gr_mbuf_flow_hash_l3(const struct rte_mbuf *m, rte_be16_t eth_type) {
+	return flow_hash_l3(m, 0, eth_type);
+}
--- a/modules/infra/datapath/flow_hash.c
+++ b/modules/infra/datapath/flow_hash.c
@@ -42,7 +42,8 @@ static uint32_t flow_hash_l3(const struct rte_mbuf *m, uint32_t l3_offset, rte_b
 		tuple.v4.dst_addr = l3.ip4->dst_addr;
 		switch (l3.ip4->next_proto_id) {
 		case IPPROTO_UDP:
-			if ((rte_be_to_cpu_16(l3.ip4->fragment_offset) & RTE_IPV4_HDR_OFFSET_MASK)
+			if ((rte_be_to_cpu_16(l3.ip4->fragment_offset)
+			     & (RTE_IPV4_HDR_OFFSET_MASK | RTE_IPV4_HDR_MF_FLAG))
 			    == 0) {
 				l4.udp = rte_pktmbuf_mtod_offset(
 					m,
@@ -57,7 +58,8 @@ static uint32_t flow_hash_l3(const struct rte_mbuf *m, uint32_t l3_offset, rte_b
 			}
 			break;
 		case IPPROTO_TCP:
-			if ((rte_be_to_cpu_16(l3.ip4->fragment_offset) & RTE_IPV4_HDR_OFFSET_MASK)
+			if ((rte_be_to_cpu_16(l3.ip4->fragment_offset)
+			     & (RTE_IPV4_HDR_OFFSET_MASK | RTE_IPV4_HDR_MF_FLAG))
 			    == 0) {
 				l4.tcp = rte_pktmbuf_mtod_offset(
 					m,
--- a/modules/infra/datapath/flow_hash.h
+++ b/modules/infra/datapath/flow_hash.h
@@ -5,6 +5,7 @@
 
 #include <rte_mbuf.h>
 
+#include <stdbool.h>
 #include <stdint.h>
 
 typedef enum : uint8_t {
@@ -13,7 +14,46 @@ typedef enum : uint8_t {
 	GR_MBUF_FLOW_HASH_RSS,
 } gr_mbuf_flow_hash_mode_t;
 
-// Return a stable packet-flow hash. The packet data must start with an
-// Ethernet header. RSS mode uses a hardware hash when present and falls back
-// to a software L3/L4 hash for virtual devices without RSS.
+// Calculate a packet-flow hash without caching it. The packet data must start
+// with an Ethernet header. RSS mode uses a hardware hash when present and falls
+// back to a software L3/L4 hash for virtual devices without RSS.
 uint32_t gr_mbuf_flow_hash(const struct rte_mbuf *, gr_mbuf_flow_hash_mode_t);
+
+static inline bool gr_mbuf_flow_hash_is_valid(const struct rte_mbuf *m) {
+	return m->ol_flags & RTE_MBUF_F_RX_RSS_HASH;
+}
+
+// Store a software flow hash where a hardware driver would have put its RSS
+// value, so every consumer reads m->hash.rss the same way.
+static inline void gr_mbuf_flow_hash_set(struct rte_mbuf *m, uint32_t hash) {
+	m->hash.rss = hash;
+	m->ol_flags |= RTE_MBUF_F_RX_RSS_HASH;
+}
+
+// Drop a hash that no longer describes the packet, typically after tunnel
+// decapsulation exposed an inner frame. The next consumer recomputes a
+// software hash on demand.
+static inline void gr_mbuf_flow_hash_invalidate(struct rte_mbuf *m) {
+	m->ol_flags &= ~RTE_MBUF_F_RX_RSS_HASH;
+}
+
+// Same as gr_mbuf_flow_hash() in L3/L4 mode for packets whose data starts
+// with an IPv4 or IPv6 header. eth_type must match that header.
+uint32_t gr_mbuf_flow_hash_l3(const struct rte_mbuf *, rte_be16_t eth_type);
+
+// Return the flow hash of an Ethernet-framed packet. A hash already present
+// in m->hash.rss is returned as is; otherwise a software L3/L4 hash is
+// calculated once and cached there.
+static inline uint32_t gr_mbuf_flow_hash_get(struct rte_mbuf *m) {
+	if (!gr_mbuf_flow_hash_is_valid(m))
+		gr_mbuf_flow_hash_set(m, gr_mbuf_flow_hash(m, GR_MBUF_FLOW_HASH_L3_L4));
+	return m->hash.rss;
+}
+
+// Same as gr_mbuf_flow_hash_get() for packets whose data starts with an IPv4
+// or IPv6 header. eth_type must match that header.
+static inline uint32_t gr_mbuf_flow_hash_get_l3(struct rte_mbuf *m, rte_be16_t eth_type) {
+	if (!gr_mbuf_flow_hash_is_valid(m))
+		gr_mbuf_flow_hash_set(m, gr_mbuf_flow_hash_l3(m, eth_type));
+	return m->hash.rss;
+}
--- a/modules/infra/datapath/flow_hash.h
+++ b/modules/infra/datapath/flow_hash.h
@@ -1,0 +1,19 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026 Harrison Caldicott
+
+#pragma once
+
+#include <rte_mbuf.h>
+
+#include <stdint.h>
+
+typedef enum : uint8_t {
+	GR_MBUF_FLOW_HASH_L2,
+	GR_MBUF_FLOW_HASH_L3_L4,
+	GR_MBUF_FLOW_HASH_RSS,
+} gr_mbuf_flow_hash_mode_t;
+
+// Return a stable packet-flow hash. The packet data must start with an
+// Ethernet header. RSS mode uses a hardware hash when present and falls back
+// to a software L3/L4 hash for virtual devices without RSS.
+uint32_t gr_mbuf_flow_hash(const struct rte_mbuf *, gr_mbuf_flow_hash_mode_t);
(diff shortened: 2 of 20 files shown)


modules/infra/datapath/flow_hash.h
```diff
@@ -13,6 +15,43 @@ typedef enum : uint8_t {
...
 uint32_t gr_mbuf_flow_hash(const struct rte_mbuf *, gr_mbuf_flow_hash_mode_t);
+
+static inline bool gr_mbuf_flow_hash_is_valid(const struct rte_mbuf *m) {
+	return m->ol_flags & gr_mbuf_flow_hash_flag;
+}
+
+static inline void gr_mbuf_flow_hash_set(struct rte_mbuf *m, uint32_t hash) {
+	*RTE_MBUF_DYNFIELD(m, gr_mbuf_flow_hash_offset, uint32_t *) = hash;
+	m->ol_flags |= gr_mbuf_flow_hash_flag;
+}
```
rjarry, line 31: Why no reuse the existing `m->hash.rss` field? If no driver sets it, you are free to store whatever you want here.

rjarry (changes requested):
There is a bit too much code churn in the successive commits. Could you try and make this more incremental?

Did you test if this series introduces any performance change? I am concerned about the re-computation of RSS after decap/encap even with hardware drivers that support L4 UDP ports.

Also, please drop the `Assisted-by: <LLM>` trailers.

hcaldicott:
Thanks for the review - I will re-work and clean this up. This was cherry-picked from other development work on our MH implementation, so it could use some more scrutiny - sit tight.

modules/infra/datapath/flow_hash.h
hcaldicott: Addressed - good point.

hcaldicott:
> There is a bit too much code churn in the successive commits. Could you try and make this more incremental?

Could you confirm if the commit history is acceptable now? I updated Patch 1 so that it uses the final shape from the beginning to remove most of the churn. This was due to cherry-picking from my upstream MH codebase - apologies.

> Did you test if this series introduces any performance change? I am concerned about the re-computation of RSS after decap/encap even with hardware drivers that support L4 UDP ports.

Fair call - I have reworked so that decap no longer recomputes anything. `vxlan_input` just clears the RSS flag; a software hash for the inner frame is computed lazily at the first consumer that needs one, then cached. Packets with genuine NIC RSS follow the exact upstream code path (same flag test, same field read), so hardware fast paths are untouched. I am going to see how I can benchmark this.

> Also, please drop the `Assisted-by: <LLM>` trailers.

Done.

hcaldicott:
Okay, I have thrown together some AI tests and sanity checked them, seems good to me.

I benchmarked the series against v0.17.1 on a pair of Xeon D-2143IT boxes (X722 10GbE SFP+, back to back), using the per-node cycle counters from `grcli stats software` under ~226 kpps of 8-flow UDP. 

Two topologies: plain routed forwarding (kernel TAP in, 10G wire, TAP out) and a bridged VXLAN overlay mirroring a production EVPN setup. 30 s measured runs after warmup, 3 reps per build per scenario, medians below. X722 RSS was confirmed active (flows spread across both RX queues).

| path | node | v0.17.1 | this series | delta |
|---|---|---|---|---|
| routed ingress, NIC RSS present | `ip_input` | 67.9 c/pkt | ~68 c/pkt | none — one flag test |
| routed ingress, no RSS (TAP) | `ip_input` | 70.3 | 206.6 | +136 c/pkt (~62 ns): the software Toeplitz |
| vxlan encap from bridged no-RSS port | `vxlan_output` | 84.6 | 245.7 | +161 c/pkt: same hash over the inner frame |
| vxlan decap | `vxlan_input` | 47.6 | 51.5 | +4 c/pkt: clearing the flag |
| all other nodes | | | | within noise (±5%) |

Notes:
- c/pkt is CPU cycles per packet - lower being better.
- Where the driver provides RSS, the datapath is unchanged: same flag test, same `m->hash.rss` read, nothing recomputed after encap. An earlier revision of this series showed +5 c/pkt here from the getters being extern functions (call overhead on a test-and-return path); they are now static inline and that is gone.
- The +136/+161 only appear on paths that previously fed stale or zero data into the fib lookups and the source port — the bug being fixed. The hash is computed once and cached, so fib lookup, vxlan source port and bond selection on the same packet share one computation.
- After decap nothing is recomputed unless a consumer actually needs a hash for the inner frame.

Throughput was generator-limited and identical on both builds. The fixed build showed lower loss in the vxlan scenario (<=0.01% vs up to 0.77%): per-flow source ports let the receiving NIC spread flows across both RSS queues instead of piling them onto one.

Incidental finding from the same rig: the X722 PF firmware rejects `rte_eth_allmulticast_enable()` with -ENOTSUP, so `interface add port` on these NICs fails hard on current main — I had to apply #702 to both builds to run the benchmark, so statistics here might be slightly skewed by that change also.

Crux of this: I see about a 62ns (at 2.2ghz) performance _decrease_ with the corrected hashing function in place. However I feel like it could be worthwhile for the correctness aspect of this fix?

modules/infra/datapath/flow_hash.c
```diff
@@ -0,0 +1,143 @@
...
+	uint32_t l3_offset;
+	rte_be16_t eth_type;
+
+	if (mode == GR_MBUF_FLOW_HASH_RSS && (m->ol_flags & RTE_MBUF_F_RX_RSS_HASH))
+		return m->hash.rss;
+
+	eth = rte_pktmbuf_mtod(m, const struct rte_ether_hdr *);
+	tuple.l2.mac = eth->dst_addr;
+	if (eth->ether_type == RTE_BE16(RTE_ETHER_TYPE_VLAN)) {
+		vlan = (const struct rte_vlan_hdr *)(eth + 1);
```
christophefontaine, line 127: `vlan = PAYLOAD(eth);`

modules/infra/datapath/flow_hash.c
```diff
@@ -0,0 +1,143 @@
...
+			tuple.v4.sport = 0;
+			tuple.v4.dport = 0;
+		}
+		len = sizeof(tuple.v4);
+		break;
+	case RTE_BE16(RTE_ETHER_TYPE_IPV6):
+		l3.ip6 = rte_pktmbuf_mtod_offset(m, const struct rte_ipv6_hdr *, l3_offset);
+		tuple.v6.src_addr = l3.ip6->src_addr;
+		tuple.v6.dst_addr = l3.ip6->dst_addr;
+		switch (l3.ip6->proto) {
```
christophefontaine, line 80: Outside of this commit, and eligible for a future discussion / PR.
As we only look at `ip6->proto`, we do not take into account the extension headers which could be before a  valid `IPPROTO_*`.
So, as part of the hash computation, what should we do ?

modules/l2/datapath/vxlan_input.c
```diff
@@ -94,6 +95,7 @@ vxlan_input_process(struct rte_graph *graph, struct rte_node *node, void **objs,
 		}
 
 		rte_pktmbuf_adj(m, sizeof(struct rte_udp_hdr) + sizeof(*vh));
+		gr_mbuf_flow_hash_invalidate(m);
```
christophefontaine, line 98: I'm not comfortable in always invalidating the hash, as it may have been computed by the hardware on the inner packet.
Also, relying on the PTYPE may not be good either: as you said, even if the hash isn't computed on the inner, the remote VTEP may have spread the source UDP port.
```
  // keep a hardware-provided inner-flow hash; only invalidate an outer one
  if (!(m->packet_type & RTE_PTYPE_INNER_L4_MASK))
      gr_mbuf_flow_hash_invalidate(m);
```
And I know that @david-marchand is not fan of using packet_type as well :)

modules/policy/datapath/dnat44_dynamic.c
```diff
@@ -91,7 +92,11 @@ static uint16_t dnat44_dynamic_process(
 		);
 
 		o = l3_mbuf_data(m);
-		o->nh = fib4_lookup(o->iface->vrf_id, ip->dst_addr, m->hash.rss);
+		o->nh = fib4_lookup(
+			o->iface->vrf_id,
+			ip->dst_addr,
+			gr_mbuf_flow_hash_get_l3(m, RTE_BE16(RTE_ETHER_TYPE_IPV4))
+		);
```
christophefontaine, line 99: (not introduced by this commit, but to open the discussion for a future PR)
The `fib4_lookup` gets the `nh` based on the hash. Yet, this hash is still the one computed by the hardware, pre-nat.
Shouldn't we update the RSS field to reflect the modified `dst_ip & dsp_port`?

modules/policy/datapath/dnat44_static.c
```diff
@@ -74,7 +75,11 @@ static uint16_t dnat44_static_process(
 		// We need the old address value to fixup the checksum properly.
 		ip->dst_addr = dnat->replace;
 
-		d->nh = fib4_lookup(d->iface->vrf_id, ip->dst_addr, mbuf->hash.rss);
+		d->nh = fib4_lookup(
+			d->iface->vrf_id,
+			ip->dst_addr,
+			gr_mbuf_flow_hash_get_l3(mbuf, RTE_BE16(RTE_ETHER_TYPE_IPV4))
+		);
```
christophefontaine, line 82: same comment as `dnat44_dynamic`

christophefontaine (commented):
With the introduction of  `gr_mbuf_flow_hash_get` to get the `mbuf->hash.rss`, we should also update `process_behav_decap` in `modules/srv6/datapath/srv6_local.c` to set the hash.
WDYT ?

And to update `modules/ip/datapath/icmp_local_send.c` as well as `icmp6_local_send.c` for consistency.

modules/l2/datapath/vxlan_input.c
david-marchand: @hcaldicott 

Hum, I did not look too much in detail at the full PR sorry.
I am trying to go back to the actual problem you are facing.

The issue seems to be that the rss hash is not a function of the inner packet.
This should be the case for routers implementing the recommendation in [RFC7348](https://datatracker.ietf.org/doc/html/rfc7348).

```
   Outer UDP Header:  This is the outer UDP header with a source port
      provided by the VTEP and the destination port being a well-known
      UDP port.
...
-  Source Port:  It is recommended that the UDP source port number
         be calculated using a hash of fields from the inner packet --
         one example being a hash of the inner Ethernet frame's headers.
         This is to enable a level of entropy for the ECMP/load-
         balancing of the VM-to-VM traffic across the VXLAN overlay.
         When calculating the UDP source port number in this manner, it
         is RECOMMENDED that the value be in the dynamic/private port
         range 49152-65535 [[RFC6335](https://datatracker.ietf.org/doc/html/rfc6335)].
```

If you are receiving traffic from another grout instance, the rss hash should have a source port that depends on the inner packet RSS (https://github.com/DPDK/grout/blob/main/modules/l2/datapath/vxlan_output.c#L90).

Is it that in your usecase the inner (meaning, before encapsulation on the transmitter side) packet rss 0 because it is received from a net/virtio (including virtio-user) port?

If this is the case, you may want to compute a RSS when it is not set at the RX side (see rx_virtio_process/rx_bond_virtio_process).

hcaldicott: @david-marchand yes, that is exactly the use case: the access ports here are virtio/vhost and TAP, so pre-encap packets carry no RSS and vxlan_output was reading `hash.rss == 0` for every flow - which is why all encapsulated flows shared one source port (the red test in the PR description).

On where to fix it: rather than computing eagerly in `rx_virtio_process`/`rx_bond_virtio_process`, the series computes the same software hash lazily at the first consumer (vxlan_output, bond_output, the fib lookups) through `gr_mbuf_flow_hash_get()`, which stores it in `m->hash.rss` and sets `RTE_MBUF_F_RX_RSS_HASH` exactly as a driver would. The end state is the same as computing it at RX, but packets that never reach a hash consumer do not pay for it. I measured the software hash at roughly 136 cycles/packet on Xeon D / X722; charged at RX it would tax every packet on virtio ports whether or not anything downstream needs a hash. If you would still rather see it in the RX path for uniformity I can move it, but the lazy placement was deliberate.

On the invalidation: fair point about RFC 7348. When the remote VTEP derives the source port from the inner flow (kernel vxlan, grout with this series), the outer RSS is already a function of the inner flow, so keeping it preserves both per-flow affinity and distribution after decap, and invalidating just forces a pointless software rehash - in our deployment that is every decapped packet that egresses over a bond. It also discards a hardware inner hash on NICs that parse the tunnel, per @christophefontaine's comment, without needing any packet_type check. The one thing the invalidation protected against is a remote VTEP using a fixed source port (which includes any pre-fix grout feeding from ports without RSS), but that traffic already polarizes across every underlay ECMP hop, so I agree it is not worth making the common case pay. The invalidation is dropped in v4; vxlan_input is now untouched by the series.

SRv6 decap is a different story (see Christophe's review comment on `process_behav_decap`): there the outer header carries no inner entropy at all, so that path does need to drop the stale hash. Handled in v4.

modules/infra/datapath/flow_hash.c
hcaldicott: Done in v4 (with `gr_macro.h` included directly — nothing in the datapath headers re-exports it).

modules/infra/datapath/flow_hash.c
hcaldicott: Agreed, follow-up material. Right now anything that is not immediately TCP/UDP falls through to the default arm and gets a L3-only hash: less entropy, but stable, so flows with extension headers degrade the same way fragments do instead of risking reordering. For a future PR I see two options: a bounded walk over the extension chain (stopping at a fragment header and staying L3-only in that case, same rule this series applies to IPv4 fragments), or using the flow label per RFC 6438 when it is non-zero, which trusts the sender to have derived it from the inner flow. The walk is probably the safer default. Happy to open that discussion separately.

modules/policy/datapath/dnat44_dynamic.c
hcaldicott: I think keeping the pre-NAT hash is actually correct: the hash is a stable per-flow identity for member selection, not a digest of the current header bytes. A given flow always produces the same pre-NAT tuple and distinct flows have distinct pre-NAT tuples, so both affinity and distribution hold; recomputing after the rewrite would spend cycles without changing either. 

One inconsistency worth noting for that future discussion: the software fallback in this series computes after the rewrite (the lookup sits below the NAT code), so RSS ports hash pre-NAT and non-RSS ports hash post-NAT. Each flow still sees one stable value since a port either has RSS or it does not - but if we ever want the two paths to agree, the cheap fix is seeding the software hash before the rewrite, not recomputing after it.

modules/policy/datapath/dnat44_static.c
hcaldicott: Same reasoning as on the dnat44_dynamic thread: the pre-NAT hash is still a stable per-flow identity, so I would leave both as they are and take the wider question to a follow-up discussion.

hcaldicott:
> With the introduction of `gr_mbuf_flow_hash_get` to get the `mbuf->hash.rss`, we should also update `process_behav_decap` in `modules/srv6/datapath/srv6_local.c` to set the hash. WDYT ?

@christophefontaine agreed, and it is a bit worse than an inconsistency. `hash.usr` aliases `hash.rss`, so the decap overwrites the value with the outer flow label while a hardware `RTE_MBUF_F_RX_RSS_HASH` flag (if any) stays set - and since our own encap sets the outer flow label to zero (`ip6_set_fields()`), every decapped flow ends up with a "valid" hash of 0. Unlike vxlan there is no inner entropy in the outer header either: no UDP source port, and RSS cannot parse past the routing header. 

I kept your flow-label approach for the case where the encapsulating node filled it in: a non-zero label is now stored through `gr_mbuf_flow_hash_set()` so the valid flag is coherent, and a zero label invalidates the hash so the first consumer computes one from the inner packet. New patch in v4, with a unit test covering both cases (it fails on the previous code with the stale outer hash still marked valid). Setting the outer flow label from the flow hash on encap (RFC 6438) would be a nice follow-up - it would also give the underlay ECMP something to work with.

> And to update `modules/ip/datapath/icmp_local_send.c` as well as `icmp6_local_send.c` for consistency.

Sure — the "fake RSS" writes there are exactly what `gr_mbuf_flow_hash_set()` does; converted in v4.

---

v4 is pushed (7 patches now). Summary of the changes since v3:

- Dropped the vxlan decap invalidation entirely; vxlan_input is untouched by the series and the patch 5 commit message carries the RFC 7348 rationale (see the resolved inline thread).
- New patch "srv6: fix flow hash validity on decap" as described above, with a `Fixes:` tag and a unit test.
- `icmp_local_send`/`icmp6_local_send` converted to `gr_mbuf_flow_hash_set()`.
- `PAYLOAD(eth)` in the VLAN parsing.

Revalidated in the AlmaLinux 9 container: full unit suite, vxlan, vxlan6 and ip_loadbalance smoke tests pass; the new srv6 test fails with only the fix reverted. PR description updated to match.

modules/infra/datapath/flow_hash.c
```diff
@@ -0,0 +1,219 @@
...
+	if (bit < 0)
+		ABORT("rte_mbuf_dynflag_register(grout_flow_hash_valid): %s",
+		      rte_strerror(rte_errno));
+	gr_mbuf_flow_hash_flag = RTE_BIT64(bit);
+}
+
+static struct module module = {
+	.name = "flow_hash",
+	.init = flow_hash_init,
+};
```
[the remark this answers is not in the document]
hcaldicott: Obsolete since v3: the dynfield registration was replaced by reusing `m->hash.rss` + `RTE_MBUF_F_RX_RSS_HASH`, so there is no init ordering left to race.

modules/l2/datapath/vxlan_input.c
david-marchand: > @david-marchand yes, that is exactly the use case: the access ports here are virtio/vhost and TAP, so pre-encap packets carry no RSS and vxlan_output was reading `hash.rss == 0` for every flow - which is why all encapsulated flows shared one source port (the red test in the PR description).

Ok, thanks for confirming.

> On where to fix it: rather than computing eagerly in `rx_virtio_process`/`rx_bond_virtio_process`, the series computes the same software hash lazily at the first consumer (vxlan_output, bond_output, the fib lookups) through `gr_mbuf_flow_hash_get()`, which stores it in `m->hash.rss` and sets `RTE_MBUF_F_RX_RSS_HASH` exactly as a driver would. The end state is the same as computing it at RX, but packets that never reach a hash consumer do not pay for it. I measured the software hash at roughly 136 cycles/packet on Xeon D / X722; charged at RX it would tax every packet on virtio ports whether or not anything downstream needs a hash. If you would still rather see it in the RX path for uniformity I can move it, but the lazy placement was deliberate.

There is always a tradeoff, unfortunately.
I am a bit scared at touching the stack to deal with a driver issue.

Do you have an idea of the performance impact of adding the branches in the RSS hash users?


> On the invalidation: fair point about RFC 7348. When the remote VTEP derives the source port from the inner flow (kernel vxlan, grout with this series), the outer RSS is already a function of the inner flow, so keeping it preserves both per-flow affinity and distribution after decap, and invalidating just forces a pointless software rehash - in our deployment that is every decapped packet that egresses over a bond. It also discards a hardware inner hash on NICs that parse the tunnel, per @christophefontaine's comment, without needing any packet_type check. The one thing the invalidation protected against is a remote VTEP using a fixed source port (which includes any pre-fix grout feeding from ports without RSS), but that traffic already polarizes across every underlay ECMP hop, so I agree it is not worth making the common case pay. The invalidation is dropped in v4; vxlan_input is now untouched by the series.

Ack.


> SRv6 decap is a different story (see Christophe's review comment on `process_behav_decap`): there the outer header carries no inner entropy at all, so that path does need to drop the stale hash. Handled in v4.

I had not considered SRv6 so far.
That seems problematic on the SRv6 receiving node.

https://learn.microsoft.com/en-us/windows-hardware/drivers/network/rss-hashing-types#ndis-hash-ipv6
https://learn.microsoft.com/en-us/windows-hardware/drivers/network/rss-hashing-types#ndis_hash_ipv6_ex

Nothing seems to look at the ipv6 flow label (that's me assuming grout sets it based on rss hash, @christophefontaine @rjarry ?).
On the other hand, I am not sure grout sets the described IPv6 options by Microsoft specification.

On DPDK side, it seems really few drivers support such rss on IPv6 extensions, and that seems to be for old hardware.

net/bnxt caught my eyes, as it seems to do something unspecified.. hashing ipv6 flow label (oh oh).

@christophefontaine had some funny idea, but this is something for another thread.

modules/infra/datapath/flow_hash.c
```diff
@@ -0,0 +1,145 @@
...
+		len = sizeof(tuple.v6);
+		break;
+	default:
+		return 0;
+	}
+
+	return rte_softrss_be(&tuple.u32, len / sizeof(uint32_t), rss_key);
+}
+
+uint32_t gr_mbuf_flow_hash(const struct rte_mbuf *m, gr_mbuf_flow_hash_mode_t mode) {
```
rjarry, line 110: I think it makes sense to move this function as `static inline` in `flow_hash.h` to benefit from the shortcut when the RSS hash has been computed by hardware. `flow_hash_l3` can remain in `flow_hash.c`.

modules/infra/datapath/flow_hash.c
```diff
@@ -0,0 +1,145 @@
...
+			rte_be16_t vlan_id;
+		} l2;
+	} tuple;
+	const struct rte_ether_hdr *eth;
+	const struct rte_vlan_hdr *vlan;
+	uint32_t l3_offset;
+	rte_be16_t eth_type;
+
+	if (mode == GR_MBUF_FLOW_HASH_RSS && (m->ol_flags & RTE_MBUF_F_RX_RSS_HASH))
+		return m->hash.rss;
```
rjarry, line 124:
```c
	if (likely(mode == GR_MBUF_FLOW_HASH_RSS && (m->ol_flags & RTE_MBUF_F_RX_RSS_HASH)))
		return m->hash.rss;
```

modules/infra/datapath/flow_hash.c
```diff
@@ -42,7 +42,8 @@ static uint32_t flow_hash_l3(const struct rte_mbuf *m, uint32_t l3_offset, rte_b
 		tuple.v4.dst_addr = l3.ip4->dst_addr;
 		switch (l3.ip4->next_proto_id) {
 		case IPPROTO_UDP:
-			if (l3.ip4->fragment_offset == 0) {
+			if ((rte_be_to_cpu_16(l3.ip4->fragment_offset) & RTE_IPV4_HDR_OFFSET_MASK)
+			    == 0) {
```
rjarry, line 46: Sometimes `clang-format` makes me sad :(

Do you think we could use [`rte_ipv4_frag_pkt_is_fragmented`](https://doc.dpdk.org/api/rte__ip__frag_8h.html#a20a18f3aadcc2f5bdd9ccdd306860ba5) instead?

```c
			if (!rte_ipv4_frag_pkt_is_fragmented(l3.ip4)) {
```

It checks for the MF flag too which looks correct.

modules/infra/datapath/flow_hash.c
```diff
@@ -56,7 +57,8 @@ static uint32_t flow_hash_l3(const struct rte_mbuf *m, uint32_t l3_offset, rte_b
 			}
 			break;
 		case IPPROTO_TCP:
-			if (l3.ip4->fragment_offset == 0) {
+			if ((rte_be_to_cpu_16(l3.ip4->fragment_offset) & RTE_IPV4_HDR_OFFSET_MASK)
+			    == 0) {
```
rjarry, line 61: Ditto.

modules/ip/datapath/ip_error.c
```diff
@@ -33,13 +34,15 @@ ip_error_process(struct rte_graph *graph, struct rte_node *node, void **objs, ui
 	struct rte_ipv4_hdr *ip;
 	struct rte_mbuf *mbuf;
 	ip4_addr_t src, dst;
+	uint32_t hash;
 	rte_edge_t edge;
 	unsigned len;
```
rjarry, line 39: Reverse x-mas tree please.

```diff
	ip4_addr_t src, dst;
	rte_edge_t edge;
+	uint32_t hash;
	unsigned len;
```

modules/ipip/datapath_out.c
```diff
@@ -32,6 +33,7 @@ ipip_output_process(struct rte_graph *graph, struct rte_node *node, void **objs,
 	struct rte_ipv4_hdr *outer;
 	const struct iface *iface;
 	struct rte_mbuf *mbuf;
+	uint32_t hash;
 	rte_edge_t edge;
```
rjarry, line 37: Reverse xmas tree too

modules/srv6/datapath/srv6_output.c
rjarry, line 60: Oh dear, this function is already looking bad. Could you add the new variables roughly in the correct spot for reverse xmas tree?

```diff
	struct rte_ipv6_routing_ext *srh;
	struct rte_ipv6_hdr *outer_ip6;
	const struct nexthop *nh;
	uint32_t optlen, plen;
+	rte_be16_t eth_type;
	struct rte_mbuf *m;
	uint8_t proto, n_omitted_segs;
	rte_edge_t edge;
+	uint32_t hash;
```

rjarry (changes requested):

maxime-leroy:
Regarding lazy hashing, I don't have an objection to the approach itself.

It avoids paying the software Toeplitz cost when the hash is never consumed. However, I'm not convinced yet that this translates into a net CPU gain.

With the eager approach, the fallback hash is computed at ingress.(See pr742), when the packet headers have just been accessed and are likely still hot in cache.

With lazy hashing, the first consumer can be several graph nodes later. At that point, computing the hash may require accessing packet data that is no longer in cache.

So the comparison is not simply:

software Toeplitz cost vs no Toeplitz cost

but potentially:

eager Toeplitz on hot packet data

versus:

lazy Toeplitz + later packet-data cache misses.

It would be interesting to benchmark cycles/packet together with cache misses for no-RSS traffic where the hash is unused, used immediately, and used later in the graph.

Without such measurements, I'm not convinced that lazy hashing provides a real CPU performance improvement.

The eager approach also has the advantage of being simpler: normalize the missing RSS at ingress and keep the existing "mbuf->hash.rss" contract everywhere else.

My main concern is the handling of an existing hardware RSS across tunnel decapsulation.

Removing an outer header does not imply that the hardware RSS has become useless. Tunnel-aware hardware can calculate RSS using information from the inner packet.

This is not theoretical for me: DPAA2 calculates RSS using both inner and outer flow information for tunneled packets.

Therefore, invalidating a valid DPAA2 hardware RSS after decapsulation and later recomputing the inner flow hash in software is a no-go for me.

The NIC has already calculated the inner-flow entropy. Throwing that result away and paying for a software Toeplitz calculation to obtain it again is both unnecessary and counterproductive.

Unfortunately, DPDK currently tells us that "mbuf->hash.rss" contains a valid RSS value, but does not provide enough per-packet information to tell Grout whether that RSS was derived from the outer packet, the inner packet, or both.

Therefore we cannot use decapsulation itself as a reason to invalidate an existing RSS hash.

As a workaround, we could consider a Grout port parameter describing the RSS behavior of the port, for example whether the hardware provides inner/tunnel-aware RSS.

This would allow us to preserve the hardware RSS on DPAA2 while still handling devices with different RSS behavior.

SRv6 makes this a blocker for me.

On DPAA2, discarding the hardware RSS during SRv6 decapsulation means discarding entropy already calculated from the inner flow and potentially paying to calculate it again in software.

Replacing the RSS directly with the SRv6 Flow Label is also questionable. RFC 8754 section 5.5 requires the Flow Label to be used according to RFC 6438 when calculating the ECMP hash; RFC 6438 defines it as an input to that calculation together with the IPv6 source and destination addresses, not as the final hash itself.

So I see two separate questions:

- whether lazy hashing provides a real CPU benefit once cache effects are taken into account;
- how hardware RSS is preserved across tunnel decapsulation.

The first one can be answered with more cache-aware benchmarking.

For the second one, invalidating the hardware RSS after decapsulation cannot be the default behavior.

I would not merge the SRv6 behavior as-is: on DPAA2 it discards useful hardware-computed inner-flow entropy and can then make us pay to calculate that entropy again in software.

hcaldicott:
Apologies - I will review these comments in the coming days as I've been quite busy.

christophefontaine:
Dumb suggestion: what about putting the queue id in the rss field for virtio interfaces in the reception node ?

That wouldn't replace the whole rss hash computation for more advanced use case, but the queue id could provide a spread good enough to push the packets to 2 (or more) ECMP peers or to select the proper link in an active-active bond ?
Eventually, this could be the `port_id + queue_id`, so even with 2 virtual interfaces with a unique queue, the 2 members of an active-active bond would be used ?

PR #702, opened 2026-08-17
infra: tolerate unsupported all-multicast mode

hcaldicott:
`rte_eth_allmulticast_enable()` returns `-ENOTSUP` for drivers that have
no all-multicast toggle — virtual PMDs in particular can be inherently
unfiltered and expose nothing to switch. `iface_port_reconfig()` treats
any error from it as fatal, so a port backed by such a driver cannot be
configured at all.

Extract the call into a `port_allmulticast_enable()` helper that follows
the same pattern as the existing link, promiscuous and MTU handlers:
tolerate `-ENOSYS`/`-EOPNOTSUPP`, derive `GR_IFACE_S_ALLMULTI` from
`rte_eth_allmulticast_get()` so the flag reflects the actual device
state, and log a warning when all-multicast could not be enabled. Real
failures from drivers that do support the toggle remain fatal.

Found during the same EVPN multihoming bring-up as #698, but reproducible
on any port whose PMD lacks the toggle since ec4cbdc7f18f made allmulti
unconditional.

## Testing

- Four unit tests in `iface_test` mock `rte_eth_allmulticast_enable()`
  and `rte_eth_allmulticast_get()`: success sets `GR_IFACE_S_ALLMULTI`,
  `-ENOTSUP` succeeds with the flag tracking the device state read-back
  (both off and already-on cases), and a real error (`-EIO`) still fails
  reconfiguration.
- A red test for the pre-fix behaviour was judged impractical: the
  failure sits inside `iface_port_reconfig()`, which needs the full port
  setup path mocked out. The failure mode is established by inspection —
  the DPDK API documents the `-ENOTSUP` return, and the old code returned
  it as fatal unconditionally.
- With the fix: full unit suite passes (AlmaLinux 9 container, arm64).

Related: #698

### DIFF
--- a/modules/infra/control/iface_test.c
+++ b/modules/infra/control/iface_test.c
@@ -57,6 +57,7 @@ mock_func(int, port_plug(struct iface_info_port *));
 mock_func(unsigned, worker_count(void));
 mock_func(int, worker_queue_distribute(const cpu_set_t *, vec struct iface_info_port **));
 mock_func(int, __wrap_rte_eth_allmulticast_enable(uint16_t));
+mock_func(int, __wrap_rte_eth_allmulticast_get(uint16_t));
 mock_func(int, __wrap_rte_eth_dev_mac_addr_add(uint16_t, struct rte_ether_addr *, uint32_t));
 mock_func(int, __wrap_rte_eth_dev_mac_addr_remove(uint16_t, struct rte_ether_addr *));
 mock_func(int, __wrap_rte_eth_promiscuous_disable(uint16_t));
@@ -185,6 +186,37 @@ static void port_mac_del_unicast(void **) {
 	assert_true(iface->macs[1].hardware);
 }
 
+static void port_allmulticast_supported(void **) {
+	iface->state &= ~GR_IFACE_S_ALLMULTI;
+	will_return(__wrap_rte_eth_allmulticast_enable, 0);
+	will_return(__wrap_rte_eth_allmulticast_get, 1);
+	assert_return_code(port_allmulticast_enable(iface), errno);
+	assert_true(iface->state & GR_IFACE_S_ALLMULTI);
+}
+
+static void port_allmulticast_unsupported(void **) {
+	iface->state &= ~GR_IFACE_S_ALLMULTI;
+	will_return(__wrap_rte_eth_allmulticast_enable, -ENOTSUP);
+	will_return(__wrap_rte_eth_allmulticast_get, 0);
+	assert_return_code(port_allmulticast_enable(iface), errno);
+	assert_false(iface->state & GR_IFACE_S_ALLMULTI);
+}
+
+static void port_allmulticast_unsupported_but_on(void **) {
+	iface->state &= ~GR_IFACE_S_ALLMULTI;
+	will_return(__wrap_rte_eth_allmulticast_enable, -ENOTSUP);
+	will_return(__wrap_rte_eth_allmulticast_get, 1);
+	assert_return_code(port_allmulticast_enable(iface), errno);
+	assert_true(iface->state & GR_IFACE_S_ALLMULTI);
+}
+
+static void port_allmulticast_error(void **) {
+	iface->state &= ~GR_IFACE_S_ALLMULTI;
+	will_return(__wrap_rte_eth_allmulticast_enable, -EIO);
+	assert_int_equal(port_allmulticast_enable(iface), -EIO);
+	assert_false(iface->state & GR_IFACE_S_ALLMULTI);
+}
+
 static void vlan_mac_add_multicast(void **) {
 	assert_return_code(iface_add_eth_addr(vlan_iface, &mcast1), errno);
 	assert_int_equal(vec_len(vlan_iface->macs), 0);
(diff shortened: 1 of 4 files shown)


rjarry:
I am curious, with which driver did you get this error? There are a very small number of drivers that do not support setting ALLMULTI.

Also, how will IPv6 work if your vdev does not allow receiving multicast frames?

hcaldicott:
I hit this with `net_virtio_user` backed by `/dev/vhost-net`. That backend never negotiates `VIRTIO_NET_F_CTRL_RX` (the kernel does not implement the control queue that I can see), so virtio_dev_allmulticast_enable() returns `-ENOTSUP` and port configuration fails.

We have been testing the use of `virtio-user` to give the host kernel stack access to the DPDK dataplane, carrying KVM live-migration traffic over the EVPN fabric we're designing around. `net_tap` works for this too and remains our fallback if this PR isn't accepted, but `virtio-user` with the `vhost-net` backend gives us better throughput.

Re v6: the failing call means the receive mode has no explicit toggle, not that multicast is filtered based on my testing - at least for this driver. Without `CTRL_RX` the device has no RX filtering path at all, so frames from the tap arrive unfiltered. My test assigns IPv6 addresses on the port's Linux peer and runs bidirectional `ping -6` with no static neighbours — resolution succeeds in both directions. That indicates (to me) that IPv6 ND is operational, and hence multicast is OK.

On the shape of the fix: I kept it generic rather than special-casing a driver because the error is conditional on feature negotiation, not driver identity (the same PMD succeeds under QEMU, which offers CTRL_RX), and because port.c already tolerates `-ENOSYS`/`-EOPNOTSUPP` the same way for set_link_up/down, promiscuous enable/disable and set_mtu. 

I am happy to rework it to match that pattern more closely - switch on the same errnos and derive `GR_IFACE_S_ALLMULTI` from `rte_eth_allmulticast_get()` like the promisc handler does? Not sure what you think would be cleaner?

Let me know if you want this closer to `rte_eth_allmulticast_get()` :)

hcaldicott:
> I hit this with `net_virtio_user` backed by `/dev/vhost-net`. That backend never negotiates `VIRTIO_NET_F_CTRL_RX` (the kernel does not implement the control queue that I can see), so virtio_dev_allmulticast_enable() returns `-ENOTSUP` and port configuration fails.
> 
> We have been testing the use of `virtio-user` to give the host kernel stack access to the DPDK dataplane, carrying KVM live-migration traffic over the EVPN fabric we're designing around. `net_tap` works for this too and remains our fallback if this PR isn't accepted, but `virtio-user` with the `vhost-net` backend gives us better throughput.
> 
> Re v6: the failing call means the receive mode has no explicit toggle, not that multicast is filtered based on my testing - at least for this driver. Without `CTRL_RX` the device has no RX filtering path at all, so frames from the tap arrive unfiltered. My test assigns IPv6 addresses on the port's Linux peer and runs bidirectional `ping -6` with no static neighbours — resolution succeeds in both directions. That indicates (to me) that IPv6 ND is operational, and hence multicast is OK.
> 
> On the shape of the fix: I kept it generic rather than special-casing a driver because the error is conditional on feature negotiation, not driver identity (the same PMD succeeds under QEMU, which offers CTRL_RX), and because port.c already tolerates `-ENOSYS`/`-EOPNOTSUPP` the same way for set_link_up/down, promiscuous enable/disable and set_mtu.
> 
> I am happy to rework it to match that pattern more closely - switch on the same errnos and derive `GR_IFACE_S_ALLMULTI` from `rte_eth_allmulticast_get()` like the promisc handler does? Not sure what you think would be cleaner?
> 
> Let me know if you want this closer to `rte_eth_allmulticast_get()` :)

It'd go something like this:

```
int port_allmulticast_enable(struct iface *iface) {
	struct iface_info_port *p = iface_info_port(iface);
	int ret;

	ret = rte_eth_allmulticast_enable(p->port_id);
	switch (ret) {
	case 0:
	case -ENOSYS:
	case -EOPNOTSUPP:
		break;
	default:
		return errno_log(-ret, "rte_eth_allmulticast_enable");
	}

	if (rte_eth_allmulticast_get(p->port_id) == 1)
		iface->state |= GR_IFACE_S_ALLMULTI;
	else
		iface->state &= ~GR_IFACE_S_ALLMULTI;

	return 0;
}
```

rjarry:
Understood, thanks for the details. I would like at least to log a warning when the operation isn't supported by the driver:

```c
int port_allmulticast_enable(struct iface *iface) {
	struct iface_info_port *p = iface_info_port(iface);
	int ret;

	ret = rte_eth_allmulticast_enable(p->port_id);
	switch (ret) {
	case 0:
	case -ENOSYS:
	case -EOPNOTSUPP:
		break;
	default:
		return errno_log(-ret, "rte_eth_allmulticast_enable");
	}

	if (rte_eth_allmulticast_get(p->port_id) == 1) {
		iface->state |= GR_IFACE_S_ALLMULTI;
	} else {
		LOG(WARNING, "%s: cannot enable ALLMULTI: %s", iface->name, strerror(-ret ?: EPERM));
		iface->state &= ~GR_IFACE_S_ALLMULTI;
	}

	return 0;
}
```

hcaldicott:
Reworked as suggested, thanks — the helper now matches the promisc/link/MTU pattern: tolerates `-ENOSYS`/`-EOPNOTSUPP`, derives `GR_IFACE_S_ALLMULTI` from `rte_eth_allmulticast_get()` and warns when all-multicast could not be enabled. Unit tests updated to mock the read-back as well, including the case where the toggle is unsupported but the device already reports all-multicast active. Full suite passes.

rjarry (approved):

PR #703, opened 2026-08-17
ip: preserve connected routes during protocol replay

hcaldicott:
During control-plane replay, a routing daemon can briefly advertise a
connected prefix before it has learned the interface address, then
withdraw it again once it reconciles. `route4_add()` replaces the
address-owned connected route with the protocol route, so the later
withdrawal removes the only route to the prefix while the address stays
configured — the connected network is unreachable until the address is
removed and re-added.

This change treats address-owned connected routes as authoritative:
protocol adds for such prefixes are accepted as a silent no-op, and API
deletes of internal-origin routes are refused the same way. Deletions now
resolve their nexthop with an exact-prefix lookup first, so the nexthop
type of a covering route is never used to delete a more specific prefix;
the previous longest-prefix lookup remains as the fallback when no exact
entry exists. Same handling for both address families.

Found while integrating FRR restart handling for EVPN multihoming (#698),
but the race only needs a routing daemon replaying routes over a
configured address.

## Testing

- New smoke test `connected_route_ownership_test.sh` replays the
  add/withdraw cycle over configured IPv4 and IPv6 addresses and asserts
  the connected routes survive with their `link` origin.
- Without the fix (test-only applied to `main`): the test fails at the
  first assertion — after add/withdraw, `172.16.0.0/24` is gone from the
  RIB entirely.
- With the fix: the new test passes, the full unit suite passes, and
  `ip_loadbalance_test.sh` confirms ordinary route add/del and ECMP
  behaviour is unchanged (AlmaLinux 9 container, arm64).

Related: #698

### DIFF
--- a/modules/ip/control/route.c
+++ b/modules/ip/control/route.c
@@ -198,6 +198,36 @@ struct nexthop *rib4_lookup_exact(uint16_t vrf_id, ip4_addr_t ip, uint8_t prefix
 	return nh_id_to_ptr(nh_id);
 }
 
+static bool is_addr_owned_route(uint16_t vrf_id, ip4_addr_t ip, uint8_t prefixlen) {
+	struct rte_fib *fib = get_fib(vrf_id);
+	struct nexthop_info_l3 *l3;
+	struct rte_rib_node *rn;
+	struct rte_rib *rib;
+	struct nexthop *nh;
+	gr_nh_origin_t *o;
+	uintptr_t nh_id;
+
+	if (fib == NULL)
+		return false;
+
+	rib = rte_fib_get_rib(fib);
+	rn = rte_rib_lookup_exact(rib, rte_be_to_cpu_32(ip), prefixlen);
+	if (rn == NULL)
+		return false;
+
+	o = rte_rib_get_ext(rn);
+	if (*o != GR_NH_ORIGIN_LINK && *o != GR_NH_ORIGIN_INTERNAL)
+		return false;
+
+	rte_rib_get_nh(rn, &nh_id);
+	nh = nh_id_to_ptr(nh_id);
+	if (nh->type != GR_NH_T_L3)
+		return false;
+
+	l3 = nexthop_info_l3(nh);
+	return (l3->flags & NH_LOCAL_ADDR_FLAGS) == NH_LOCAL_ADDR_FLAGS;
+}
+
 struct route4_event {
 	struct ip4_net dest;
 	uint16_t vrf_id;
@@ -339,6 +369,14 @@ static struct api_out route4_add(const void *request, struct api_ctx *) {
 	if (req->origin == GR_NH_ORIGIN_INTERNAL)
 		return api_out(EINVAL, 0, NULL);
 
+	// Address-owned connected routes are authoritative. During control-plane
+	// replay, a routing protocol can transiently offer the same prefix before
+	// Zebra has reconciled the interface address. Do not replace the connected
+	// route: the later protocol withdrawal would otherwise remove the only
+	// route while leaving the address configured.
+	if (is_addr_owned_route(req->vrf_id, req->dest.ip, req->dest.prefixlen))
+		return api_out(0, 0, NULL);
+
 	if (req->nh_id != GR_NH_ID_UNSET) {
 		nh = nexthop_lookup_id(req->nh_id);
 		if (nh == NULL)
@@ -387,7 +425,11 @@ static struct api_out route4_del(const void *request, struct api_ctx *) {
 	struct nexthop *nh = NULL;
 	int ret;
 
-	nh = rib4_lookup(req->vrf_id, req->dest.ip);
+	if (is_addr_owned_route(req->vrf_id, req->dest.ip, req->dest.prefixlen))
+		return api_out(EBUSY, 0, NULL);
+	nh = rib4_lookup_exact(req->vrf_id, req->dest.ip, req->dest.prefixlen);
+	if (nh == NULL)
+		nh = rib4_lookup(req->vrf_id, req->dest.ip);
 	ret = rib4_delete(
 		req->vrf_id, req->dest.ip, req->dest.prefixlen, nh ? nh->type : GR_NH_T_L3
 	);
--- a/modules/ip6/control/route.c
+++ b/modules/ip6/control/route.c
@@ -215,6 +215,44 @@ struct nexthop *rib6_lookup_exact(
 	return nh_id_to_ptr(nh_id);
 }
 
+static bool is_addr_owned_route(
+	uint16_t vrf_id,
+	uint16_t iface_id,
+	const struct rte_ipv6_addr *ip,
+	uint8_t prefixlen
+) {
+	struct rte_fib6 *fib6 = get_fib6(vrf_id);
+	const struct rte_ipv6_addr *scoped_ip;
+	struct nexthop_info_l3 *l3;
+	struct rte_rib6_node *rn;
+	struct rte_ipv6_addr tmp;
+	struct rte_rib6 *rib;
+	struct nexthop *nh;
+	gr_nh_origin_t *o;
+	uintptr_t nh_id;
+
+	if (fib6 == NULL)
+		return false;
+
+	scoped_ip = addr6_linklocal_scope(ip, &tmp, iface_id);
+	rib = rte_fib6_get_rib(fib6);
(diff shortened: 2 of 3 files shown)


modules/ip/control/route.c
```diff
@@ -332,13 +332,23 @@ int rib4_delete(uint16_t vrf_id, ip4_addr_t ip, uint8_t prefixlen, gr_nh_type_t
...
 		return api_out(EINVAL, 0, NULL);
 
+	// Address-owned connected routes are authoritative. During control-plane
+	// replay, a routing protocol can transiently offer the same prefix before
+	// Zebra has reconciled the interface address. Do not replace the connected
+	// route: the later protocol withdrawal would otherwise remove the only
+	// route while leaving the address configured.
+	existing = rib4_lookup_exact(req->vrf_id, req->dest.ip, req->dest.prefixlen);
+	if (existing != NULL && existing->origin == GR_NH_ORIGIN_INTERNAL)
+		return api_out(0, 0, NULL);
```
rjarry, line 378: You are making the decision based on the **nexthop** origin. Not the **route**'s. I would advise to add a small static function that gets the route origin and for good measure, also checks that the nexthop looks like a local address.

```c
static bool is_addr_owned_route(uint16_t vrf_id, ip4_addr_t ip, uint8_t prefixlen) {
	struct rte_fib *fib = get_fib(vrf_id);
	struct nexthop_info_l3 *l3;
	struct rte_rib_node *rn;
	struct rte_rib *rib;
	struct nexthop *nh;
	gr_nh_origin_t *o;
	uintptr_t nh_id;

	if (fib == NULL)
		return false;

	rib = rte_fib_get_rib(fib);
	rn = rte_rib_lookup_exact(rib, rte_be_to_cpu_32(ip), prefixlen);
	if (rn == NULL)
		return false;

	o = rte_rib_get_ext(rn);
	if (*o != GR_NH_ORIGIN_INTERNAL)
		return false;

	rte_rib_get_nh(rn, &nh_id);
	nh = nh_id_to_ptr(nh_id);
	if (nh->type != GR_NH_T_L3)
		return false;

	l3 = nexthop_info_l3(nh);
	return (l3->flags & NH_LOCAL_ADDR_FLAGS) == NH_LOCAL_ADDR_FLAGS;
}
```

modules/ip/control/route.c
```diff
@@ -387,7 +397,11 @@ static struct api_out route4_del(const void *request, struct api_ctx *) {
 	struct nexthop *nh = NULL;
 	int ret;
 
-	nh = rib4_lookup(req->vrf_id, req->dest.ip);
+	nh = rib4_lookup_exact(req->vrf_id, req->dest.ip, req->dest.prefixlen);
+	if (nh != NULL && nh->origin == GR_NH_ORIGIN_INTERNAL)
+		return api_out(0, 0, NULL);
+	if (nh == NULL)
+		nh = rib4_lookup(req->vrf_id, req->dest.ip);
```
rjarry, line 432: Same here. Simply reuse `is_addr_owned_route()`.

Shouldn't grout return an EBUSY error in that case?

modules/ip6/control/route.c
```diff
@@ -364,13 +364,22 @@ int rib6_delete(
...
 		return api_out(EINVAL, 0, NULL);
 
+	// Keep address-owned connected routes authoritative during routing daemon
+	// replay. See the IPv4 handler for the ordering race this prevents.
+	existing = rib6_lookup_exact(
+		req->vrf_id, GR_IFACE_ID_UNDEF, &req->dest.ip, req->dest.prefixlen
+	);
+	if (existing != NULL && existing->origin == GR_NH_ORIGIN_INTERNAL)
+		return api_out(0, 0, NULL);
+
```
rjarry, line 416: Same issue. You'll need to add an ipv6 version of is_addr_owned_route().

modules/ip6/control/route.c
```diff
@@ -428,7 +437,11 @@ static struct api_out route6_del(const void *request, struct api_ctx *) {
 	struct nexthop *nh = NULL;
 	int ret;
 
-	nh = rib6_lookup(req->vrf_id, GR_IFACE_ID_UNDEF, &req->dest.ip);
+	nh = rib6_lookup_exact(req->vrf_id, GR_IFACE_ID_UNDEF, &req->dest.ip, req->dest.prefixlen);
+	if (nh != NULL && nh->origin == GR_NH_ORIGIN_INTERNAL)
+		return api_out(0, 0, NULL);
+	if (nh == NULL)
+		nh = rib6_lookup(req->vrf_id, GR_IFACE_ID_UNDEF, &req->dest.ip);
```
rjarry, line 478: ditto.

rjarry (changes requested):
Hey Harrison,

Your check is a bit loose. See how to make it more robust below.

Also could you remove any mention that you were assisted by AI? It does not bring much value and we prefer to keep the git history leaner.

Thanks!

hcaldicott:
> Hey Harrison,
> 
> Your check is a bit loose. See how to make it more robust below.
> 
> Also could you remove any mention that you were assisted by AI? It does not bring much value and we prefer to keep the git history leaner.
> 
> Thanks!

Good pickups - I will address these and re-run my lab tests!

modules/ip/control/route.c
hcaldicott: Yes it should! Fixed.

modules/ip6/control/route.c
hcaldicott: Done.

modules/ip6/control/route.c
hcaldicott: Sorted!

modules/ip/control/route.c
hcaldicott: Yep, this is a much more eloquent solution. The local next-hop validation is a great addition. I have implemented this snippet and referenced it in the updated PR.

hcaldicott:
I have updated the PR. I am now rebuilding our physical lab on this to re-test the corrected behaviour. Converting back to a draft until my testing is completed.

rjarry:
Hey @hcaldicott, did you make progress with your testing?

hcaldicott:
> Hey @hcaldicott, did you make progress with your testing?

Sorry for the slow turnaround - I've been quite busy and had kind of forgotten to finish this off - thanks for the bump to remind me. I have now re-tested the reworked guards on the physical cluster. Converting back from draft.

hcaldicott:
Testing is done — the reworked guards are now validated on our physical lab (4x Lenovo SE350, X722 fabric, iBGP EVPN full mesh, FRR 10.6.1 with the grout dplane plugin).

Setup: the PR head cherry-picked onto our production branch, packaged through our staging pipeline (unit suite runs in the package build), and installed on one fabric node carrying live EVPN services. The node's address-owned routes are two fabric /31s (origin link), the /32 router-id (origin internal) and the per-interface IPv6 link-local /64s.

Results:

Ownership guards, exercised directly against the real prefixes: deleting any of the address-owned routes is refused with EBUSY; a protocol-style re-add of a connected /31 via the fabric peer is accepted as a no-op and the route keeps its link origin.
5 consecutive FRR restarts (full zebra replay through dplane_grout): all connected routes present after every cycle, BGP re-Established 3/3 in ~14s, all EVPN prefixes re-received, RIB size stable. No dplane errors logged.
One full grout restart: the address replay recreates the connected routes and the immediately following FRR replay does not displace them — this is exactly the race that originally motivated the patch.

rjarry (approved):
Thanks for testing again!

modules/ip6/control/route.c
```diff
@@ -371,6 +409,11 @@ static struct api_out route6_add(const void *request, struct api_ctx *) {
 	if (req->origin == GR_NH_ORIGIN_INTERNAL)
 		return api_out(EINVAL, 0, NULL);
 
+	// Keep address-owned connected routes authoritative during routing daemon
+	// replay. See the IPv4 handler for the ordering race this prevents.
+	if (is_addr_owned_route(req->vrf_id, GR_IFACE_ID_UNDEF, &req->dest.ip, req->dest.prefixlen))
+		return api_out(0, 0, NULL);
```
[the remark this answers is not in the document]
rjarry: @hcaldicott do you think this is a legit concern? I don't think FRR would try to replace a link local route. That does not make much sense to me. And we don't have any interface ID to use here anyway.

PR #708, opened 2026-08-18
fdb: fix cache-line thrashing in fdb_learn

rjarry:
Writing the last_seen timestamp field invalidates the related cache lines in all CPUs. This causes a systematic cache miss in any FDB object field access just after. This causes non-linear scaling with the number of datapath workers.

Update last_seen *after* testing if the iface_id or vtep fields have changed. And only update it if it hasn't changed in 1 second.

The results for a L2VPN VXLAN test (in packets per second):

              1c/2t     2c/4t     4c/8t
    Before     7.5M     12.0M     19.0M
    After      7.5M     14.8M     29.7M

### DIFF
--- a/modules/l2/control/fdb.c
+++ b/modules/l2/control/fdb.c
@@ -118,6 +118,7 @@ void fdb_learn(
 	const struct l3_addr *vtep
 ) {
 	const struct fdb_key key = {bridge_id, vlan_id, *mac};
+	gr_clock_ns_t now = gr_clock_ns();
 	struct fdb_entry *fdb;
 	void *data;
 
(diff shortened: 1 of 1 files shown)


modules/l2/control/fdb.c
```diff
@@ -118,6 +118,7 @@ void fdb_learn(
 	const struct l3_addr *vtep
 ) {
 	const struct fdb_key key = {bridge_id, vlan_id, *mac};
+	gr_clock_ns_t now = gr_clock_ns();
```
christophefontaine: Should we move from `gr_clock_ns` to `rdtsc` (+ a helper ?)

maxime-leroy (approved):

PR #710, opened 2026-08-24
ip,ip6,frr: flush routes like zebra on address and iface teardown

maxime-leroy:
zebra drops routes from its RIB on its own, without sending anything to the
dataplane, in two situations:

  - an interface goes administratively down, on every supported FRR version
  - the last IPv4 address of an interface is deleted, since 10.7
    (FRRouting/frr#19564)

It assumes the dataplane already dropped them, as Linux does
(fib_disable_ip, fib_del_ifaddr), and calls rib_delete() with
fromkernel=true, so nothing reaches the provider. grout keeps them, so the
two tables disagree for good: grout still forwards on routes zebra no
longer knows, and zebra never re-learns them until the plugin resyncs.

Measured with grout + zebra:

    grcli address del <last IPv4 addr>   10.5: kept in both   10.7: grout only
    grcli interface set port p0 down     10.5 and 10.7: grout only, v4 and v6

grout now flushes those routes too and notifies, so both tables converge on
every FRR version. The address case is unconditional: without a source
address the remaining nexthops can never be resolved, so keeping them was a
defect of its own. The interface case is a compatibility concession, a down
interface being a reversible state, so it is optional
(flush-on-iface-down, off by default) and the plugin enables it. Carrier
loss is not a trigger, the interface stays configured up. Routes installed
by a routing daemon are never touched, their owner reevaluates them.

The notifications for the routes deleted with an interface are dropped as
well (skip-events-on-iface-down, off by default, enabled by the plugin):
zebra has already purged them, and each one costs a full rib_delete() plus
a failed reinstall attempt for the routes zebra owns.

Two new smoke tests. Both assert that grout and zebra hold the same view
rather than a fixed outcome, so they are version agnostic, and both are red
without the series:

      frr-10.5.4   addr_del_kernel_route / iface_down_kernel_route   OK
      frr-10.7.0   addr_del_kernel_route / iface_down_kernel_route   OK

### DIFF
--- a/frr/zebra_dplane_grout.c
+++ b/frr/zebra_dplane_grout.c
@@ -713,6 +713,10 @@ static void dplane_grout_connect(struct event *) {
 		&grout_ctx.dg_t_dplane_update
 	);
 
+	// redone on every connection: a restarted grout comes back with its defaults
+	struct gr_iface_config_set_req req = {.flush_routes_on_iface_down = true};
+	grout_client_send_recv(GR_IFACE_CONFIG_SET, sizeof(req), &req, NULL);
+
 	gr_log_notice("connected, monitoring iface/ip events");
 }
 
--- a/frr/zebra_dplane_grout.c
+++ b/frr/zebra_dplane_grout.c
@@ -714,7 +714,11 @@ static void dplane_grout_connect(struct event *) {
 	);
 
 	// redone on every connection: a restarted grout comes back with its defaults
-	struct gr_iface_config_set_req req = {.flush_routes_on_iface_down = true};
+	struct gr_iface_config_set_req req = {
+		.flush_routes_on_iface_down = true,
+		.skip_route_events_on_iface_down = true,
+		.set_attrs = GR_IFACE_CONFIG_SET_FLUSH_ROUTES | GR_IFACE_CONFIG_SET_SKIP_EVENTS,
+	};
 	grout_client_send_recv(GR_IFACE_CONFIG_SET, sizeof(req), &req, NULL);
 
 	gr_log_notice("connected, monitoring iface/ip events");
--- a/main/grout.default
+++ b/main/grout.default
@@ -45,6 +45,11 @@ GROUT_OVERRIDE_RP_FILTER=true
 # nexthop and come back when the interface goes up again.
 #GROUT_FLUSH_ROUTES_ON_IFACE_DOWN=false
 
+# Do not emit route events for the routes deleted with an interface. Only
+# useful for a control plane which already evicts them on the interface
+# event (default: false).
+#GROUT_SKIP_ROUTE_EVENTS_ON_IFACE_DOWN=false
+
 # Maximum Transmission Unit (default: 1800).
 #GROUT_MAX_MTU=1800
 
--- a/main/grout.default
+++ b/main/grout.default
@@ -40,6 +40,11 @@ GROUT_OVERRIDE_RP_FILTER=true
 # Print all ingress/egress packets (default: false).
 #GROUT_TRACE_PACKETS=false
 
+# Delete the routes of an interface when it goes administratively down, like
+# zebra and Linux do. Off by default: the routes are kept with an unresolved
+# nexthop and come back when the interface goes up again.
+#GROUT_FLUSH_ROUTES_ON_IFACE_DOWN=false
+
 # Maximum Transmission Unit (default: 1800).
 #GROUT_MAX_MTU=1800
 
--- a/modules/infra/api/iface.c
+++ b/modules/infra/api/iface.c
@@ -333,6 +333,7 @@ static struct api_out iface_config_get(const void *, struct api_ctx *) {
 		return api_out(ENOMEM, 0, NULL);
 
 	resp->flush_routes_on_iface_down = gr_config.flush_routes_on_iface_down;
+	resp->skip_route_events_on_iface_down = gr_config.skip_route_events_on_iface_down;
 
 	return api_out(0, sizeof(*resp), resp);
 }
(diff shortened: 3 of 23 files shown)


maxime-leroy:
The API check failure is not a real incompatibility. Both
`gr_iface_config` and `gr_iface_config_set_req` are added by this series:
the second commit creates them, the fourth extends them, and nothing
outside the series ever saw the intermediate layout. Against main the
whole series is additive, only new requests and new types, so any client
built before it stays compatible.

The check trips because its baseline is the parent commit:

    prev_revision=${PREV_REVISION:-HEAD^}

which means every breaking commit needs its own bump. `GR_API_VERSION` is
a handshake token though, and what it has to describe accurately is the
compatibility between two *released* binaries, not between two commits of
a development cycle. Comparing against the last release tag instead would
express that, and would leave the commit layout of a series free.

So I would rather not bump here: it would put every client and daemon in
lockstep for a struct layout that never shipped. If you prefer the check
green on every commit as it stands, I can declare the final layout in the
commit which introduces the requests, or squash the two commits together.

rjarry:
See my comment on #711.

I understand that in this specific case: adding a new message and modifying it in the same patch series should *not* cause check-api.sh to fail. **However**, there is no way to differentiate between a new message being modified mid pull request from other real ABI breaking changes. 

I really want to keep all API breaking changes self contained (i.e. with their own GR_API_VERSION bump attached).

If you can find a way to determine the "correct" reference commit mid pull request, we can drop the hard-coded `HEAD^`. But I don't think it will be trivial, and most likely impossible.

rjarry:
Commenting on the code here, why do we need a knob to configure these behaviors? We could just mimic what Linux does (even if it is silly). The point here is to be transparent.

If we really want to have this configurable at runtime, I think it would make more sense to use environment variables like for other opt-in things.

https://github.com/DPDK/grout/blob/819435ff2e048fa696c6e57bcef43626ce398021/main/grout.default#L17-L26

maxime-leroy:
> Commenting on the code here, why do we need a knob to configure these behaviors? We could just mimic what Linux does (even if it is silly). The point here is to be transparent.
> 
> If we really want to have this configurable at runtime, I think it would make more sense to use environment variables like for other opt-in things.
> 
> https://github.com/DPDK/grout/blob/819435ff2e048fa696c6e57bcef43626ce398021/main/grout.default#L17-L26

An env var does not work for the only consumer there is: the plugin cannot
set one in grout's process, so it would have to go in grout.default, and
installing the plugin without editing the unit file leaves the divergence
in place. Hence an API request.

    // redone on every connection: a restarted grout comes back with its defaults
    struct gr_iface_config_set_req req = {.flush_routes_on_down = true};
    if (gr_api_client_send_recv(grout_ctx.client, GR_IFACE_CONFIG_SET, sizeof(req), &req, NULL) < 0)
            gr_log_err("GR_IFACE_CONFIG_SET: %s", strerror(errno));


On mimicking Linux: Linux did make this configurable, at runtime and per
device.

    net.ipv6.conf.<dev>.keep_addr_on_down
    net.ipv4.conf.<dev>.ignore_routes_with_linkdown
    net.ipv6.conf.<dev>.ignore_routes_with_linkdown

keep_addr_on_down decides whether the addresses, and the routes derived
from them, survive an admin down. ignore_routes_with_linkdown decides
whether routes out of a carrier-down device are used at all. The only
piece that is genuinely unconditional is the IPv4 flush in
fib_disable_ip().

The last point is that the current grout behaviour is better than the
kernel's here, not merely different. Bring a Linux interface down and up
and the static routes through it are gone for good; the connected ones
come back only because the addresses were kept. grout keeps the route and
leaves the nexthop unresolved, so it comes back with the interface. The
flush matches origin <= GR_NH_ORIGIN_STATIC, so making it unconditional
would silently drop routes a user typed into grcli. For anyone running
grout without a control plane that is a regression, and the kernel
behaviour it aligns with is a historical wart rather than a design.

frr/zebra_dplane_grout.c
```diff
@@ -710,6 +710,12 @@ static void dplane_grout_connect(struct event *) {
 		&grout_ctx.dg_t_dplane_update
 	);
 
+	// redone on every connection: a restarted grout comes back with its defaults
+	struct gr_iface_config_set_req req = {.flush_routes_on_down = true};
+	if (gr_api_client_send_recv(grout_ctx.client, GR_IFACE_CONFIG_SET, sizeof(req), &req, NULL)
+	    < 0)
+		gr_log_err("GR_IFACE_CONFIG_SET: %s", strerror(errno));
```
rjarry: Why not use grout_client_send_recv()?

```c
	// redone on every connection: a restarted grout comes back with its defaults
	struct gr_iface_config_set_req req = {.flush_routes_on_down = true};
	grout_client_send_recv(GR_IFACE_CONFIG_SET, sizeof(req), &req, NULL);
```

It already logs errors.

modules/infra/api/iface.c
```diff
@@ -339,7 +339,12 @@ static struct api_out config_get(const void *, struct api_ctx *) {
 
 static struct api_out config_set(const void *request, struct api_ctx *) {
 	const struct gr_iface_config_set_req *req = request;
-	iface_config = req->base;
+
+	if (req->set_attrs & GR_IFACE_CONFIG_SET_FLUSH_ROUTES)
+		iface_config.flush_routes_on_down = req->flush_routes_on_down;
+	if (req->set_attrs & GR_IFACE_CONFIG_SET_SKIP_EVENTS)
+		iface_config.skip_route_events_on_iface_down = req->skip_route_events_on_iface_down;
```
rjarry: Instead of a private struct iface_config, could you add bool fields to gr_config:

https://github.com/DPDK/grout/blob/b1e59afaead0acf99815ac4322ea044ac690741b/main/config.h#L26-L27

This way, the object is visible globally via `gr_config.skip_route_events_on_iface_down`.

Also, add env vars to tune these settings on startup (in addition to the API messages for the frr plugin).

rjarry (changes requested):

main/grout.default
```diff
@@ -45,6 +45,11 @@ GROUT_OVERRIDE_RP_FILTER=true
 # nexthop and come back when the interface goes up again.
 #GROUT_FLUSH_ROUTES_ON_IFACE_DOWN=false
 
+# Do not emit route events for the routes deleted with an interface. Only
+# useful for a control plane which already evicts them on the interface
+# event (default: false).
+#GROUT_SKIP_ROUTE_EVENTS_ON_IFACE_DOWN=false
```
rjarry, line 51: Please add these two new settings to docs/grout.8.scd

rjarry (changes requested):

rjarry (approved):

PR #711, opened 2026-08-24
api: compare the ABI against the last release

maxime-leroy:
The check compared the API headers with the parent commit, which required a GR_API_VERSION bump from every commit changing a type, including a commit changing a type another commit of the same series had just introduced. Nothing outside the series ever saw that intermediate layout, and bumping for it puts clients and daemons in lockstep for something which never shipped.

GR_API_VERSION is negotiated in the hello handshake between two binaries, so what it has to describe is the compatibility of released versions. Compare against the last release tag instead, which needs one bump per release cycle at most and leaves the commit layout of a series free.

What this gives up is the mixing of binaries built from two different commits of the same cycle: they announce the same version while the API has changed in between. PREV_REVISION is still honoured for anyone wanting the previous behaviour.

### DIFF
--- a/devtools/check_api.sh
+++ b/devtools/check_api.sh
@@ -8,19 +8,32 @@ abidiff=abidiff
 output=.check_api.stamp
 src_dir=.
 cc_cmd=cc
-prev_revision=${PREV_REVISION:-HEAD^}
+base_revision=${BASE_REVISION:-}
 
-while getopts "d:o:c:p:s:" opt; do
+while getopts "b:d:o:c:s:" opt; do
 	case $opt in
+	b) base_revision="$OPTARG" ;;
 	d) abidiff="$OPTARG" ;;
 	o) output="$OPTARG" ;;
 	c) cc_cmd="$OPTARG" ;;
-	p) prev_revision="$OPTARG" ;;
 	s) src_dir="$OPTARG" ;;
 	*) echo "error: invalid arguments" >&2; exit 1 ;;
 	esac
 done
 shift $((OPTIND - 1))
+
+# The base is the commit this series forks from.
+if [ -z "$base_revision" ]; then
+	# A git rebase -x, as used by CI, already knows where the series starts.
+	onto=$(git -C "$src_dir" rev-parse --absolute-git-dir)/rebase-merge/onto
+	[ ! -f "$onto" ] || base_revision=$(cat "$onto")
+fi
+if [ -z "$base_revision" ]; then
+	# Otherwise rely on @{u}, like REVISION_RANGE in the GNUmakefile.
+	base_revision=$(git -C "$src_dir" merge-base HEAD "@{u}" 2>/dev/null) || true
+fi
+# Without a base, use the parent commit where nothing is exempt.
+[ -n "$base_revision" ] || base_revision=HEAD^
 
 dir=$(dirname "$output")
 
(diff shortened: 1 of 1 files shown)


rjarry:
The point of GR_API_VERSION is twofold:

1. It prevents incompatible client/servers from attempting to communicate.
2. It makes every ABI breaking change obvious since the check-api.sh script fails if GR_API_VERSION was not updated.

I understand your point about pointless enforcing for things that will never be shipped. But think about the following issues with your approach:

* If we want to backport an ABI breaking change to a maintenance release, having the GR_API_VERSION bump self contained makes it easier to spot breaking changes. If we decouple GR_API_VERSION update from the commits that introduce breaking changes, there is a good chance we will forget to advertise them.
* If we run pre-release or mid-release clients against an incompatible server (or vice-versa), the ABI incompatibility will be silent and cause mayhem.

I would prefer keeping the current per-commit ABI compatibility check.

devtools/check_api.sh
rjarry, line 19: There is no need to have two different arguments here. Just rename prev_version to base_revision (default empty).

devtools/check_api.sh
```diff
@@ -22,23 +24,54 @@ while getopts "d:o:c:p:s:" opt; do
...
+	onto=$(git -C "$src_dir" rev-parse --absolute-git-dir)/rebase-merge/onto
+	[ ! -f "$onto" ] || base_revision=$(cat "$onto")
+fi
+if [ -z "$base_revision" ]; then
+	# Otherwise rely on @{u}, like REVISION_RANGE in the GNUmakefile.
+	base_revision=$(git -C "$src_dir" merge-base HEAD "@{u}" 2>/dev/null) || true
+fi
+# Without a base, fall back to the parent commit where nothing is exempt.
+git -C "$src_dir" merge-base --is-ancestor "$base_revision" "$prev_revision" \
+	2>/dev/null || base_revision=$prev_revision
```
rjarry: And if auto-detection fails, just set base_revision=HEAD^

devtools/check_api.sh
```diff
@@ -22,23 +24,54 @@ while getopts "d:o:c:p:s:" opt; do
...
-done | tr ' ' '\n' | sort -u |
-xargs git -C "$src_dir" archive "$prev_revision" |
-tar -C "$dir/check_api/a" -x --transform='s|.*/||'
+install_headers "$prev_revision" "$dir/check_api/a"
+
+# Same for the base, unless that is the previous revision. Kept out of
+# check_api/ since these headers are only listed, never compiled.
+base_headers=
+if [ "$(git -C "$src_dir" rev-parse "$base_revision")" != \
+	"$(git -C "$src_dir" rev-parse "$prev_revision")" ]
```
rjarry: You will need to adjust this to `git -C "$src_dir" rev-parse HEAD^` literally.

devtools/check_api.sh
```diff
@@ -99,24 +132,53 @@ while read -r func; do
...
+if [ $((status & 3)) -ne 0 ]; then
+	cat "$dir/abidiff.log" >&2
+	echo "error: abidiff failed" >&2
+	exit 1
+fi
+
+if [ "$status" -ne 0 ] || [ "$inline_change" = true ]; then
 	grep -vE '((Functions|Variables) changes|Unreachable types) summary:' "$dir/abidiff.log"
-	if grep -q '^  \[[DC]\]' "$dir/abidiff.log" || [ "$inline_change" = true ]; then
+	if [ $((status & 8)) -ne 0 ] || [ "$inline_change" = true ]; then
```
rjarry: I remember that breaking changes in "unreachable" types (e.g. types not referenced in any public function signature) do not affect bit 8 in the exit status.

Most types should be covered by dummy public functions:

https://github.com/DPDK/grout/blob/98cf3540fa0a14c15efa11e3cdf5cf88aedbceb2/devtools/check_api.sh#L54-L66

But there may be some of them which are not. And this would silently miss breaking changes in such "unreachable" types.

Could you go over all existing commits that did bump GR_API_VERSION and ensure that the detection is preserved?

devtools/check_api.sh
```diff
@@ -99,24 +132,53 @@ while read -r func; do
...
+		change="code changed"
+	else
+		continue
+	fi
+	if [ -z "$base_headers" ] || grep -qwF -e "$func" $base_headers/*.h; then
 		inline_change=true
-		echo "inline API function $func: code changed"
 	fi
+	echo "inline API function $func: $change"
 done < $dir/check_api/inline_funcs
```
rjarry, line 138: I don't understand why you need to change this.

rjarry (changes requested):
I must say I am not super comfortable with this patch. There is a non-zero chance that it will cause some actual breaking changes to be missed.

rjarry (approved):

PR #713, opened 2026-08-25
srv6: fix SRH generation for multi-segment and reduced encapsulation

maxime-leroy:
Two independent bugs in the SRv6 headend, both from 7d88a1443587
("srv6: add srv6 module"). Every test so far used a single segment with
the default encap behavior, which is why neither showed up.

`struct rte_ipv6_routing_ext` has a 32 bit `flags` field aliasing
`last_entry`, so clearing `flags` after assigning `last_entry` wipes it.
Encapsulated packets always advertise a last entry of 0, and Linux drops
such an SRH beyond two segments. This hits `h.encaps`, FRR's default, so
a three segment policy does not forward at all.

In `h.encaps.red` the SRH was sized for the full segment list while the
fill loop skips the first segment. The last slot was never written, so 16
bytes of stale mbuf headroom went on the wire:

```
RT6 (len=4, type=4, segleft=1, last-entry=0,
     [0]fd00:202:900::, [1]fd00:a68f:c1ce:4839:2e3:ad74:509b:800)
```

Those trailing bytes are the MAC addresses of previously handled packets.
Forwarding still works, hence the wire level assertion in the tests
rather than a ping.

The four combinations of encap behavior and segment count are now
covered, including the reduced single segment case where no SRH is
emitted at all. Each fix ships with the test section it makes pass, so
every commit is green on its own.

### DIFF
--- a/modules/srv6/datapath/srv6_output.c
+++ b/modules/srv6/datapath/srv6_output.c
@@ -92,7 +92,7 @@ srv6_output_process(struct rte_graph *graph, struct rte_node *node, void **objs,
 		optlen = 0;
 		reduc = d->encap == SR_H_ENCAPS_RED ? 1 : 0;
 		if (d->n_seglist > reduc)
-			optlen += sizeof(*srh) + (d->n_seglist * sizeof(d->seglist[0]));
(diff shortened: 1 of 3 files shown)


christophefontaine (approved):
Good catch !

PR #714, opened 2026-08-26
api: reduce syscall overhead in client and server

rjarry:
Reduce the number of syscalls per API transaction on both client and server sides.

On the server, use BEV_OPT_DEFER_CALLBACKS so that libevent accumulates header and payload writes in the evbuffer and flushes them in a single writev(). Remove the no-op bufferevent_flush() calls and the manual read_cb re-trigger which libevent already handles.

On the client, replace two separate send() calls with a single sendmsg() using an iovec, and add a BUFSIZ read-ahead buffer so that recv_all() can serve multiple small reads from one recv() syscall. This is particularly effective during stream iterations where many small messages arrive back-to-back.

A GR_PING request type and accompanying ping_perf benchmark tool are added to measure the resulting transaction rate.

Different approach from https://patches.dpdk.org/project/grout/patch/20260720224924.3299328-1-mb@smartsharesystems.com/

Cc: @MortenBroerup

### DIFF
--- a/api/gr_api_client_impl.h
+++ b/api/gr_api_client_impl.h
@@ -171,30 +171,6 @@ int gr_api_client_disconnect(struct gr_api_client *client) {
 	return ret;
 }
 
-static ssize_t send_all(const struct gr_api_client *c, const void *buf, size_t len) {
-	size_t remaining = len;
-	const char *ptr = buf;
-	ssize_t n;
-
-	while (remaining > 0) {
-		n = send(c->sock_fd, ptr, remaining, MSG_NOSIGNAL);
-		if (n < 0) {
-			if (remaining < len && (errno == EAGAIN || errno == EWOULDBLOCK)) {
-				// Blocking mid-transmission, wait for socket to be writable again.
-				struct pollfd pfd = {.fd = c->sock_fd, .events = POLLOUT};
-				poll(&pfd, 1, -1);
-				continue;
-			}
-			return n;
-		}
-
-		ptr += n;
-		remaining -= n;
-	}
-
-	return len;
-}
-
 static ssize_t recv_all(const struct gr_api_client *c, void *buf, size_t len) {
 	size_t remaining = len;
 	char *ptr = buf;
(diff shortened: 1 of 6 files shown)


MortenBroerup (commented):
Deferred bufferevent callbacks look like good solution.
I have a concern about huge responses...
For reference, PHP has functions to enable/disable output buffering.
Is there any means to limit libevent buffering for huge responses? Otherwise, they may consume huge amounts of memory in the server (Grout), instead of streaming in chunks to the client.
I'm asking for either a configurable threshold in libevent, or a function to temporarily disable/enable deferring the callbacks, or a flush-like function that invokes the callback.

main/api.c
```diff
@@ -254,6 +257,15 @@ void api_send(struct api_ctx *ctx, uint32_t len, const void *payload) {
 		LOG(ERR, "pid=%d cannot write header", ctx->pid);
 	if (bufferevent_write(ctx->bev, payload, len) < 0)
 		LOG(ERR, "pid=%d cannot write payload", ctx->pid);
+
+	// Force flush when buffer grows past HIGH_WATERMARK to avoid runaway memory use
+	struct evbuffer *output = bufferevent_get_output(ctx->bev);
+	if (evbuffer_get_length(output) >= HIGH_WATERMARK) {
+		evutil_socket_t fd = bufferevent_getfd(ctx->bev);
+		evbuffer_unfreeze(output, true);
+		evbuffer_write_atmost(output, fd, -1);
```
MortenBroerup: When crossing the HIGH_WATERMARK, the flush calls deeply inside libevent, to replicate this:
https://github.com/libevent/libevent/blob/master/bufferevent_sock.c#L301
But the error handling is missing here.
Would it be possible to call the libevent write callback instead?

rjarry: I have double checked and we don't need this flag after all. All bufferevent_write() calls made synchronously (without yielding back to the event loop) are coalesced into a **single** EV_WRITE event. When the read callback returns (after writing any number of times), up to 512K bytes will be written in a single system call (using writev() + iovecs).

I have removed the commit that added BEV_OPT_DEFER_CALLBACKS entirely.

api/gr_api_client_impl.h
```diff
@@ -171,13 +174,38 @@ int gr_api_client_disconnect(struct gr_api_client *client) {
...
+			n = remaining < c->recv_buf_len ? remaining : c->recv_buf_len;
+			memcpy(ptr, c->recv_buf + c->recv_buf_off, n);
+			c->recv_buf_off += n;
+			c->recv_buf_len -= n;
+			ptr += n;
+			remaining -= n;
+			continue;
+		}
+
+		if (remaining < sizeof(c->recv_buf)) {
```
MortenBroerup: With recv_buf[sizeof(struct gr_api_response) + GR_API_MAX_MSG_LEN]], this branch will always be taken. (Except when receiving a max-size message.)
In other words: We could consider changing the algorithm from conditional read-ahead, to just always read via the recv_buf.

main/api.c
MortenBroerup: OK, sounds good.
Out of curiosity, what is then the purpose of BEV_OPT_DEFER_CALLBACKS?

rjarry: It only affects **user** callbacks (i.e. the ones you set via bufferevent_set_callbacks()).

api/gr_api_client_impl.h
rjarry: Good point. I've simplified the code accordingly.

main/api.c
MortenBroerup: OK. That's confusing.
For anyone else following the discussion... I triple checked that the data written remains buffered (and not immediately sent to the client), by logging a debug message before returning from read_cb().
https://github.com/DPDK/grout/blob/819435ff2e048fa696c6e57bcef43626ce398021/main/api.c#L361
```
	LOG(DEBUG, "output buffer len %zu at return",
		evbuffer_get_length(bufferevent_get_output(bev)));
	return;
```
Still learning about Grout, and - as a bonus - also learning about libevent. :-)

MortenBroerup (approved):
I haven't looked at the FRR changes.
Everything else looks good now.

PR #716, opened 2026-08-27
main: add high-performance clock

rjarry:
Reading the common clock, gr_clock_ns() -- implemented with clock_gettime() -- is quite slow, even though the kernel exposes it as a vDSO.

For improved performance in the dataplane, introduce a high-performance clock: clock_ns(), based on snapshotting the common clock, for use where a snapshot is sufficiently accurate.

Immediately before walking the graph in the dataplane worker thread's main loop, mark the clock as trusted and update it, so it is safe to use in the "process" functions of the graph nodes.

Define GR_CLOCK_SOURCE (as CLOCK_MONOTONIC_RAW), to highlight which clock source is used by Grout.

Add definitions like GR_NS_PER_S to help convert between nanoseconds and microseconds (GR_NS_PER_US) respectively milliseconds (GR_NS_PER_MS).

Replace all direct calls to gr_clock_ns() in grout control and data plane with clock_ns() and update <gr_clock.h> include to "clock.h" accordingly.

Respin of: https://patches.dpdk.org/project/grout/patch/20260827094041.520166-1-mb@smartsharesystems.com/
Cc: @MortenBroerup

### DIFF
--- a/main/clock.h
+++ b/main/clock.h
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026 SmartShare Systems
+
+#pragma once
+
+#include <gr_clock.h>
+
+#include <rte_branch_prediction.h>
+
+#include <assert.h>
+
+// Per-thread snapshot of gr_clock_ns().
+extern __thread gr_clock_ns_t clock_snapshot_ns;
+
+// Update the clock for the current thread.
+static inline void clock_update(void) {
+	clock_snapshot_ns = gr_clock_ns();
+}
+
+// When true, assume that the clock is updated for the current thread.
+extern __thread bool clock_trusted;
+
+// Update clock_trusted for the current thread.
+static inline void clock_set_trusted(bool trusted) {
+	clock_trusted = trusted;
+}
+
+// Get the current value of clock_snapshot_ns if clock_trusted is true.
+// Otherwise, fallback to gr_clock_ns().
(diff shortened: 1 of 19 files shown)


MortenBroerup (approved):

main/clock.h
```diff
@@ -0,0 +1,36 @@
...
+static inline void clock_update(void) {
+	clock_snapshot_ns = gr_clock_ns();
+}
+
+// When true, assume that the clock is updated for the current thread.
+extern __thread bool clock_trusted;
+
+// Update clock_trusted for the current thread.
+static inline void clock_set_trusted(bool trusted) {
+	clock_trusted = trusted;
```
maxime-leroy, line 25: What is the goal of trusted ? When trusted is false ? For control thread ? In this case, could we just fallback to gr_clock_ns() when the thread is not a worker thread  ?

maxime-leroy (changes requested):

main/clock.h
MortenBroerup: Trusted is also false when the worker thread is doing housekeeping.

In theory, control threads could also use the trusted clock.
E.g. in ageing loops, they currently snapshot the clock in a local variable named "now", and use that.
As an alternative, they could update the per-thread clock, trust it, and use that snapshot instead of the "now" local variable. (And then untrust it again when done with the ageing loop.)

MortenBroerup: It's only trusted when the clock has been very recently updated.

maxime-leroy (approved):

PR #718, opened 2026-08-31
Extend packet logging to show VXLAN & SRv6 payload

rjarry:


modules/infra/datapath/trace.c
```diff
@@ -579,6 +602,21 @@ trace_eth_payload(char *buf, size_t len, const struct rte_ether_hdr *eth, size_t
...
+	if (pkt_len < sizeof(*eth))
+		return n;
+
+	SAFE_BUF(snprintf, len, " / Ethernet " ETH_F " > " ETH_F, &eth->src_addr, &eth->dst_addr);
+	SAFE_BUF(trace_eth_payload, len, eth, pkt_len - sizeof(*eth));
+
+	return n;
+err:
+	return -1;
+}
```
christophefontaine, line 660: This is nearly identical to `trace_log_packet`, maybe we could merge the 2 functions ?

rjarry: I have looked at merging them but there is very little in common. The only duplicate code is

```c
SAFE_BUF(trace_eth_payload, len, eth, pkt_len - sizeof(*eth));
```

The rest is specific to this function.

christophefontaine (approved):

PR #720, opened 2026-09-01
infra: add diagnostics (for developers)

MortenBroerup:
Add API and CLI command to dump DPDK objects.

Intended for developers only, so the object types are not enumerated for the CLI.

### DIFF
--- a/api/gr_api.h
+++ b/api/gr_api.h
@@ -11,7 +11,7 @@
 #include <stdlib.h>
 
 // Must be bumped when making non-backward compatible changes in API headers
-#define GR_API_VERSION 11
+#define GR_API_VERSION 10
 
 // API request header.
 struct gr_api_request {
@@ -164,6 +164,12 @@ const char *gr_api_message_name(uint32_t type);
 #endif
 
 struct gr_empty { };
+
+// Unstructured text, not zero-terminated.
+struct gr_text {
+	uint32_t len; // Limited by GR_API_MAX_MSG_LEN.
+	char text[/* len */]; // Text format.
+};
 
 #define GR_MAIN_MODULE 0xcafe
 
--- a/api/gr_api.h
+++ b/api/gr_api.h
@@ -11,7 +11,7 @@
 #include <stdlib.h>
 
 // Must be bumped when making non-backward compatible changes in API headers
-#define GR_API_VERSION 10
+#define GR_API_VERSION 11
 
 // API request header.
 struct gr_api_request {
@@ -57,14 +57,14 @@ gr_api_client_send(struct gr_api_client *, uint32_t req_type, size_t tx_len, con
 
 // Receive an API response with minimum payload size validation.
 // Caller must free(*rx_data) after use.
-// Returns 0 on success, negative errno on failure.
+// Returns response payload size on success, negative errno on failure.
 // Returns -EMSGSIZE if payload is non-empty but smaller than min_resp_size.
 int gr_api_client_recv(struct gr_api_client *, uint32_t req_type, uint32_t for_id, void **rx_data);
 
 // Send a request and receive the response.
 // Validates response payload size against GR_REQ-declared type.
 // Caller must free(*rx_data) after use.
-// Returns 0 on success, negative errno on failure.
+// Returns response payload size on success, negative errno on failure.
 static inline int gr_api_client_send_recv(
 	struct gr_api_client *client,
 	uint32_t req_type,
@@ -84,7 +84,7 @@ int __gr_api_client_stream_drain(struct gr_api_client *, uint32_t req_type, uint
 // Send a request and iterate over the received stream of responses.
 //
 // @param obj Iterator variable (const pointer to response object type).
-// @param ret Final return code of the operation.
+// @param ret Final return code of the operation. (Inside the loop: Response payload size.)
 // @param client API client handle.
 // @param req_type Request type code.
 // @param tx_len Request payload size (0 if tx_data is NULL).
@@ -165,12 +165,6 @@ const char *gr_api_message_name(uint32_t type);
 
 struct gr_empty { };
 
-// Unstructured text, not zero-terminated.
-struct gr_text {
-	uint32_t len; // Limited by GR_API_MAX_MSG_LEN.
-	char text[/* len */]; // Text format.
-};
-
 #define GR_MAIN_MODULE 0xcafe
 
 enum gr_main_requests : uint32_t {
@@ -247,5 +241,5 @@ struct gr_api_event {
 
 // Receive an event notification.
 // Caller must free(*event) after use.
-// Returns 0 on success, negative errno on failure.
+// Returns event payload size on success, negative errno on failure.
 int gr_api_client_event_recv(const struct gr_api_client *, struct gr_api_event **);
--- a/modules/infra/api/diagnos.c
+++ b/modules/infra/api/diagnos.c
@@ -1,0 +1,119 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026 SmartShare Systems
+
+#include "module.h"
+
+#include <gr_api.h>
+#include <gr_infra.h>
+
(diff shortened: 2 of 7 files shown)


modules/infra/api/gr_infra.h
```diff
@@ -483,6 +483,17 @@ struct gr_packet_trace_set_req {
...
+
+// Dump an object.
+struct gr_diagnos_obj_dump_req {
+#define GR_DIAGNOS_OBJ_DUMP_ID_SIZE 128
+	char type[GR_DIAGNOS_OBJ_DUMP_ID_SIZE];
+	char name[GR_DIAGNOS_OBJ_DUMP_ID_SIZE]; // All if empty.
+};
+
+GR_REQ_STREAM(GR_DIAGNOS_OBJ_DUMP, struct gr_diagnos_obj_dump_req, struct gr_text);
+
```
rjarry, line 503: What do you think about renaming to:

```c
GR_REQ_STREAM(GR_DEBUG_OBJ_DUMP, struct gr_debug_obj_dump_req, struct gr_debug_obj_dump_resp);
```

modules/infra/api/gr_infra.h
```diff
@@ -247,6 +246,7 @@ enum gr_infra_requests : uint32_t {
 	GR_PACKET_TRACE_CLEAR,
 	GR_PACKET_TRACE_DUMP,
 	GR_PACKET_TRACE_SET,
+	GR_DIAGNOS_OBJ_DUMP,
```
rjarry: Append at the end of the enum to avoid having to bump the API version.

modules/infra/api/gr_infra.h
```diff
@@ -84,7 +84,6 @@ struct __gr_iface_base {
 	uint16_t vrf_id;
 	uint16_t domain_id; // Link domain interface ID (!GR_IFACE_MODE_VRF).
 	uint32_t speed; // Link speed in Megabit/sec.
-	uint32_t cp_id; // Linux control plane interface index.
```
rjarry, line 87: Probably a rebase glitch.

api/gr_api.h
```diff
@@ -11,7 +11,7 @@
 #include <stdlib.h>
 
 // Must be bumped when making non-backward compatible changes in API headers
-#define GR_API_VERSION 11
+#define GR_API_VERSION 10
```
rjarry: This should not be needed.

api/gr_api.h
```diff
@@ -165,6 +165,12 @@ const char *gr_api_message_name(uint32_t type);
 
 struct gr_empty { };
 
+// Unstructured text, not zero-terminated.
+struct gr_text {
+	uint32_t len; // Limited by GR_API_MAX_MSG_LEN.
+	char text[/* len */]; // Text format.
+};
+
```
rjarry: Prefer a specific struct like:

https://github.com/DPDK/grout/blob/790fb503fd2acbd9f26e831b31e6bfcb00aaff5d/modules/infra/api/gr_infra.h#L438-L443

modules/infra/api/diagnos.c
```diff
@@ -0,0 +1,119 @@
...
+	if (obj != NULL)
+		helper->dump_fn(f, obj);
+	else
+		helper->dump_list_fn(f);
+	fclose(f);
+
+	for (pos = 0; pos < len; pos += resp->len) {
+		resp->len = RTE_MIN(GR_API_MAX_MSG_LEN - sizeof(*resp), len - pos);
+		memcpy(resp->text, buf + pos, resp->len);
+		api_send(ctx, sizeof(*resp) + resp->len, resp);
```
rjarry: Use NUL terminated strings to make printing easier.

modules/infra/cli/diagnos.c
```diff
@@ -0,0 +1,64 @@
...
+	    || strlcpy(req.type, type, sizeof(req.type)) >= sizeof(req.type))
+		return CMD_ERROR;
+
+	*req.name = 0;
+	if ((name = arg_str(p, "NAME")) != NULL
+	    && strlcpy(req.name, name, sizeof(req.name)) >= sizeof(req.name))
+		return CMD_ERROR;
+
+	gr_api_client_stream_foreach(resp, ret, c, GR_DIAGNOS_OBJ_DUMP, sizeof(req), &req)
+		fwrite(resp->text, 1, resp->len, stdout);
```
rjarry: If resp is just a NUL terminated string, just use printf()

rjarry (changes requested):

api/gr_api.h
MortenBroerup: I didn't change that. I guess something went wrong during cloning.
Will fix.
For me, the GitHub workflow is learning by doing... expect mistakes along the way. :-)

api/gr_api.h
MortenBroerup: I thought a generic text type would be useful for other/future modules too, like the gr_empty type.
I intentionally did not mention charset/encoding for the text type. It could be ASCII, UTF-8, or even ISO-8859-1. By describing it as "text", it seems obvious that binary data (including the NUL character) is not permitted.
I can make the type specific to the feature, if preferred?
Alternatively, I can generalize on a common gr_text type and use that for the gr_graph_dump_resp too?

The gr_graph_dump_resp is described as a NUL terminated char stream. Shouldn't that be "a NUL terminated **text** stream"? NUL is a perfectly valid char.

You also mention using a NUL terminated string instead of providing the length.
In string operations in the source code, functions with a size field are preferred, so I assumed it applied here too. I'm happy to change to NUL terminated text instead, if you still prefer?

modules/infra/api/gr_infra.h
MortenBroerup: Good point. Will do.

modules/infra/api/gr_infra.h
MortenBroerup: I considered "dump" and "debug", but I settled for a broader spectrum name, covering other developer purposes too.
Debug is associated with removing bugs.
Diagnostics is associated with analysis of what is going on.
In the future, profiling may also be part of the diagnostics commands.
In other words: "diagnostics" is a superset of "debug", and I expect much more than "debug" to go into this (developer-targeted feature) in the future.

modules/infra/cli/diagnos.c
MortenBroerup: Some dumps are larger than the max message size, so their reply comprises of a stream of messages.
Unlike recv(), gr_api_client_recv() does not return the length of the received message, and the client has no other means for determining the reply length, so NUL terminated strings would require adding a NUL character at the end of each message in the stream of messages.
So NUL cannot be used for indicating the end of the text stream; maybe two consecutive NUL characters could be used?
IMO, gr_api_client_recv() should return the length of the received message instead of 0 on success.
But I didn't dare change that, so I went with the length field instead.

modules/infra/api/gr_infra.h
MortenBroerup: Yeah, something was definitely out of sync when I submitted this.
Will try to fix it.

MortenBroerup: Turns out I copied some files from another directory which was not up to date. _PEBCAK!_
PS:
I had to lookup how to translate the Danish expression _"Fejl 40"_ (directly translated to English: _Error 40_), which is short for: Error occurred 40 cm from the computer.
Google came up with PEBCAK: Problem Exists Between Chair And Keyboard.

modules/infra/cli/diagnos.c
rjarry: I had not realized that you had responses that were larger than the max message size. Keep it that way then.

modules/infra/api/diagnos.c
rjarry: nevermind, I had missed the detail about large responses.

api/gr_api.h
rjarry: Keep the size if you have segmented large text blobs that need to be returned.

PR #723, opened 2026-09-01
l2: install the bridge SVI MAC as a local FDB entry

rjarry:
For inbound unicast to a bridge's own IP to work, the SVI MAC must be resolvable in the FDB. So far grout only got that entry by accident: bridge_input also acts as the SVI output node, so a packet sent by the L3 stack out the bridge re-entered bridge_input with the bridge MAC as source and was learned pointing back at the bridge. Being a learned entry, it aged out after ageing_time (300s) of silence. Once gone, traffic to the SVI MAC turned into unknown unicast: flooded to every member with GR_BRIDGE_F_FLOOD, or dropped without it. Linux avoids this by keeping the bridge MAC as a permanent BR_FDB_LOCAL entry.

Do the same with a GR_FDB_F_LOCAL flag. bridge_set_eth_addr() installs and refreshes the entry with iface_id set to the bridge itself, so bridge_input steers matching frames to local delivery. It never ages, cannot be deleted or flushed over the API, and learning leaves it alone (fdb_learn only touches GR_FDB_F_LEARN entries).

The entry is also the sole owner of the SVI MAC in member HW Rx filters: fdb_event_cb programs it on a MAC change, fdb_sync_hardware on member attach, fdb_iface_del_cb removes it on teardown. This drops the old bridge_all_member_add_mac() path so the MAC is programmed once.

Finally, keep this entry out of FRR. grout_macfdb_change() would push it to zebra as an EVPN type-2 route, which is wrong: the SVI MAC already reaches zebra as the interface hw_addr and is the type-5 router MAC. Skip GR_FDB_F_LOCAL there, both in the event stream and the GR_FDB_LIST resync. The upstream FRR kernel dplane behaves the same, since zebra_neigh_macfdb_update() ignores entries on the bridge device and NUD_PERMANENT ones.

### DIFF
--- a/modules/l2/control/fdb.c
+++ b/modules/l2/control/fdb.c
@@ -176,6 +176,55 @@ void fdb_purge_iface(uint16_t iface_id) {
 	}
 }
 
+int fdb_add_local(uint16_t bridge_id, const struct rte_ether_addr *mac) {
+	const struct fdb_key key = {bridge_id, 0, *mac};
+	struct fdb_entry *fdb;
+	void *data;
+	int ret;
+
+	if (rte_hash_lookup_data(fdb_hash, &key, &data) >= 0)
+		return errno_set(EEXIST);
+
+	if (rte_mempool_get(fdb_pool, &data) < 0)
+		return errno_set(ENOMEM);
+
+	fdb = data;
+	fdb->prev_iface_id = GR_IFACE_ID_UNDEF;
+	fdb->bridge_id = bridge_id;
+	fdb->vlan_id = 0;
+	fdb->mac = *mac;
+	fdb->flags = GR_FDB_F_LOCAL;
+	fdb->iface_id = bridge_id;
+	memset(&fdb->vtep, 0, sizeof(fdb->vtep));
+	fdb->last_seen = gr_clock_ns();
+
+	if ((ret = rte_hash_add_key_data(fdb_hash, &key, fdb)) < 0) {
+		rte_mempool_put(fdb_pool, fdb);
+		return errno_set(-ret);
+	}
+
+	event_push(GR_EVENT_FDB_ADD, fdb);
+
+	return 0;
+}
+
+int fdb_del_local(uint16_t bridge_id, const struct rte_ether_addr *mac) {
+	const struct fdb_key key = {bridge_id, 0, *mac};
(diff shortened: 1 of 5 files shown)


modules/l2/control/fdb.c
```diff
@@ -274,6 +334,8 @@ static struct api_out fdb_flush(const void *request, struct api_ctx *) {
 	int ret;
 
 	while (rte_hash_iterate(fdb_hash, &key, &data, &next) >= 0) {
+		if (((struct fdb_entry *)data)->flags & GR_FDB_F_LOCAL)
```
david-marchand, line 342: Should we not update fdb_match too?

Wrt fdb_list, for example.

modules/l2/control/fdb.c
```diff
@@ -176,6 +176,55 @@ void fdb_purge_iface(uint16_t iface_id) {
...
+	const struct fdb_key key = {bridge_id, 0, *mac};
+	struct fdb_entry *fdb;
+	void *data;
+
+	if (rte_hash_lookup_data(fdb_hash, &key, &data) < 0)
+		return errno_set(ENOENT);
+
+	fdb = data;
+	if (!(fdb->flags & GR_FDB_F_LOCAL))
+		return errno_set(EPERM);
```
david-marchand, line 221: General comment for the patch.

Why EPERM? Other refused operation wrt flags end up with a EINVAL.

david-marchand (commented):
I have a general question.

Why only care about the bridge interface?
Should this optimisation (because I don't think anything is broken atm) be applied to vlan ifaces on top of bridges for example?

modules/l2/control/fdb.c
rjarry: EINVAL seemed a bit vague. The arguments are "valid" from a typing perspective but the FDB entry itself cannot be modified. I hesitated between EPERM and EBUSY.

modules/l2/control/fdb.c
rjarry: Maybe. I'll have a look.

rjarry: I have added LOCAL filtering in fdb_match for consistency.

david-marchand (approved):

PR #725, opened 2026-09-03
Diagnostics for developers

MortenBroerup:
This series comprises of two patches:

**1. Update the Client API to make the response payload size available to the client**
The response payload size is already present in the response header, but the response header is not exposed to the client.
Without this Client API update, it would be necessary for variable size responses to include a length field in the response payload. Such a redundant length field can now be avoided, which makes the Grout code for both the api and the cli much cleaner for responses of this kind.

**2. Add API and CLI for developer diagnostics**
Initially, add message/command to dump various DPDK objects.
More messages/commands can be added later.

### DIFF
--- a/modules/infra/api/diagnos.c
+++ b/modules/infra/api/diagnos.c
@@ -1,0 +1,121 @@
+// SPDX-License-Identifier: BSD-3-Clause
+// Copyright (c) 2026 SmartShare Systems
+
+#include "module.h"
+
+#include <gr_api.h>
+#include <gr_infra.h>
+
+#include <rte_bus.h>
+#include <rte_bus_pci.h>
+#include <rte_graph.h>
+#include <rte_lcore.h>
+#include <rte_log.h>
+#include <rte_malloc.h>
+#include <rte_mbuf_dyn.h>
+#include <rte_memory.h>
+#include <rte_mempool.h>
+#include <rte_memzone.h>
+#include <rte_ring.h>
+#include <rte_tailq.h>
+#include <rte_trace.h>
+
+static void graph_dump(FILE *f, struct rte_graph *graph) {
+	rte_graph_obj_dump(f, graph, true);
+}
+
+static void malloc_dump_stats(FILE *f) {
+	rte_malloc_dump_stats(f, NULL);
+}
+
+static void node_dump(FILE *f, void *id) {
+	rte_node_dump(f, (rte_node_t)(uintptr_t)id);
+}
+
+static void *node_from_name(const char *name) {
+	rte_node_t id = rte_node_from_name(name);
+	return id != RTE_NODE_ID_INVALID ? (void *)(uintptr_t)id : NULL;
+}
+
+typedef void (*dump_list_fn_t)(FILE *f);
+typedef void (*dump_fn_t)(FILE *f, void *obj);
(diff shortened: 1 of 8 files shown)


modules/infra/api/diagnos.c
```diff
@@ -0,0 +1,121 @@
...
+	if (obj != NULL)
+		helper->dump_fn(f, obj);
+	else
+		helper->dump_list_fn(f);
+	fclose(f);
+
+	for (pos = 0; pos < len; pos += payload_len) {
+		payload_len = RTE_MIN((size_t)GR_API_MAX_MSG_LEN, len - pos);
+		api_send(ctx, payload_len, buf + pos);
+	}
```
MortenBroerup, line 112: Notice how beautifully we can stream directly from the dump buffer, without the need for copying the chunks to an intermediate chunk message type (with a payload length field).

modules/infra/cli/diagnos.c
```diff
@@ -0,0 +1,64 @@
...
+	    || strlcpy(req.type, type, sizeof(req.type)) >= sizeof(req.type))
+		return CMD_ERROR;
+
+	*req.name = 0;
+	if ((name = arg_str(p, "NAME")) != NULL
+	    && strlcpy(req.name, name, sizeof(req.name)) >= sizeof(req.name))
+		return CMD_ERROR;
+
+	gr_api_client_stream_foreach (resp, ret, c, GR_DIAGNOS_OBJ_DUMP, sizeof(req), &req)
+		fwrite(resp->text, 1, ret, stdout);
```
MortenBroerup, line 30: Look how simple this is, when the client has access to the message length.

modules/infra/api/gr_infra.h
```diff
@@ -483,6 +484,24 @@ struct gr_packet_trace_set_req {
...
+};
+
+// Stream of unstructured text.
+// Not NUL-terminated.
+// This struct only serves for API type validation and compatibility.
+struct gr_diagnos_obj_dump_resp {
+	char text[1];
+};
+
+GR_REQ_STREAM(GR_DIAGNOS_OBJ_DUMP, struct gr_diagnos_obj_dump_req, struct gr_diagnos_obj_dump_resp);
```
rjarry, line 503: I am probably nit picking here. But "diagnos" is a bit odd. I would expect either "diagnostics" or "diag". Any rationale behind this name choice? What do you think about this?

```
grout# diagnostics object dump ...
...
```

rjarry (changes requested):

modules/infra/api/gr_infra.h
MortenBroerup: I discarded "diag" because it could be short for "diagram".
I chose "diagnos" because it could be short for both the noun ("diagnostics") and the verb ("diagnose").
If you prefer the longer "diagnostics", I have no issues with that.

rjarry: Yes, I would prefer diagnostics. We have command completion in grcli so it will not be longer to type anyway.

MortenBroerup: Just in the CLI command, or also in all the code (gr_diagnostics_obj_dump_req etc.)?

PR #730, opened 2026-09-03
srv6: forward the last uA CSID to its adjacency

rjarry:
An End.X with the NEXT-C-SID flavor (a uA) is a cross-connect: it must hand the packet to its L3 adjacency regardless of whether more CSIDs remain in the container. When the uA is the last CSID, csid_shift() sees an empty argument and returns false, so process_behav_end() falls through to the regular End path and delivers the packet locally instead of forwarding it.

This stayed hidden because IS-IS reported csid_bits as the locator node length (16) instead of node plus function (32). The stray function bits kept the shift argument non-zero, so csid_shift() always succeeded. The referenced FRR fix now reports the correct 32 bit width, which exposes the bug: the adjacency SID becomes the sole CSID and the packet is answered locally rather than crossing the adjacency.

Consume the last CSID for End.X, leaving the bare locator block, and forward to the adjacency. Plain End (uN) keeps delivering locally when the container is exhausted.

Version-gate the smoke test csid_bits assertion and the expected shifted destination, since the released 10.5, 10.6 and 10.7 tags still carry the old 16 bit value.

Fixes: c720487b6970 ("srv6: add support for END.X behavior")
Link: https://github.com/FRRouting/frr/commit/515987912bbf2fc3cf12

### DIFF
--- a/smoke/isis_srv6_adj_sid_frr_test.sh
+++ b/smoke/isis_srv6_adj_sid_frr_test.sh
@@ -128,22 +128,36 @@ assert_nexthop "$nh_id" '.behavior == "end.x"' || {
 	fail "No SRv6-local END.X nexthops found - ISIS SRv6 integration failed"
 }
 
+case "$frr_version" in
+10.5.*|10.6.*|10.7.*)
+	csid_bits=16
+	# 5f00:100:0:1:: shifts to 5f00:100:1::
+	shifted_dst=5f00:100:1::
+	;;
+*)
+	csid_bits=32
+	# the last CSID is consumed, only the block remains
+	shifted_dst=5f00:100::
+	;;
+esac
+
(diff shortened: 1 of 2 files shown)


smoke/isis_srv6_adj_sid_frr_test.sh
```diff
@@ -152,6 +166,8 @@ for _ in $(seq 50); do
 	sleep 0.1
 done
 
+grcli trace clear
+
```
christophefontaine: do we need to clear the trace ?

rjarry: Not really, this was a leftover from my debug session. I'll drop it.

christophefontaine (approved):

PR #736, opened 2026-09-04
ip, ip6: answer arp/ndp using the strong host model

rjarry:
An ARP request or neighbor solicitation was answered whenever the target address was local to the receiving VRF, regardless of which interface owned it. This is the weak host model and causes ARP/NDP flux. Reply only when the target address is owned by the interface the request came in on, as mandated by [RFC 4861 section 7.2.3](https://www.rfc-editor.org/rfc/rfc4861#section-7.2.3) for IPv6 and recommended by [RFC 1122](https://www.rfc-editor.org/rfc/rfc1122) for IPv4.

Loopback addresses now become unreachable from any port. Add `address {un,}expose ADDR iface IFACE` as a deliberate exception to answer a local address on other interfaces of the same VRF.

Also fix several related issues: send neighbor advertisements from the input interface with its MAC, drop replies with an explicit error instead of a garbage source MAC when the interface has none, refresh address nexthops on MAC change, and stop sending gratuitous ARPs for loopback addresses or during interface teardown.


modules/infra/cli/cli_l3.h
```diff
@@ -4,11 +4,41 @@
...
+	size_t n = 0;
+	int ret;
+
+	gr_api_client_stream_foreach (id, ret, c, req_type, req_len, req)
+		SAFE_BUF(
+			snprintf, sizeof(buf), "%s%s", n > 0 ? "," : "", iface_name_from_id(c, *id)
+		);
+err:
+	gr_table_cell(table, col, "%s", n > 0 ? buf : "");
+}
```
MortenBroerup, line 40: This function is much more generic that its description says.
Maybe it should be exposed as a generic function instead? (The function name should indicate that it generates a GR_DISP_STR_ARRAY of interface names.)
Its description describes the current use case for the caller of the function.

Alternatively, make it specific to this use case and strip a bunch of parameters.

rjarry: I might duplicate the code seeing how specific this is. This will not be the end of the world.

MortenBroerup:
> > This should also be added to L2 bridge ARP/ND suppression [...]
> 
> The ARP/ND suppression is only effective on "remote" nexthops (e.g. installed by FRR BGP).
> The changes in this pull request should not affect the feature, they only modify how grout responds to _local_ addresses.

OK. Get it!
(I had misread the L2 bridge ARP/ND suppression code, and interpreted GR_NH_F_REMOTE as GR_NH_F_NEIGH, so I thought the L2 bridge ARP/ND suppression covered much broader.)

Suggestion retracted.

christophefontaine (approved):

PR #740, opened 2026-09-07
ICMP: deliver unclaimed messages to the kernel, drop the builtin ping

maxime-leroy:
Depends on #742 . Its 5 commits show here until it merges.

  grout hands punted echo replies and ICMP errors to whichever builtin ping
  session asked for them, and frees the rest. So ping(8) through a control plane
  TAP always timed out, and an ICMP error about a TCP or UDP flow never reached
  the socket it concerned, breaking PMTU discovery for FRR sessions on top of
  grout.

  Deliver them through `l4_loopback_output` instead, then remove `grcli ping`,
  `traceroute` and the API behind them, since ping(8) and traceroute(8) now work.
  
  ## Bugs this turned up

  Placed first so they apply independently:

  - `ip_local` / `ip6_local` derived the payload length without bounding it,
    wrapping a `uint16_t` to ~64k on a crafted header. `icmp_input` fed that to
    `rte_raw_cksum()`.
  - ICMPv6 had no length check and its checksum was verified nowhere, while four
    `ndp_*_input` nodes documented relying on it.
  - The IPv4 echo body length was unchecked before reading the timestamp, the
    IPv6 one checked away from where it is read.
  - The ICMP input nodes accepted a quoted datagram without validating it.
  - A bond MAC never reached its TAP nor its VLAN sub-interfaces.

  ## Testing

  New unit test blocks for `ip_local`, `ip6_local`, `icmp_input`, `icmp6_input`,
  each checked to fail without its fix. `smoke/icmp_kernel_punt_test.sh` covers
  the punt. `ip_icmp_test.sh` and `ip6_icmp_test.sh`, renamed from
  `*_builtin_icmp_test.sh`, now drive the same paths from the kernel. Each of the
  14 commits builds and passes unit tests alone.


modules/infra/datapath/flow_hash.h
```diff
@@ -67,7 +70,7 @@ flow_hash_l3l4(const struct rte_mbuf *m, uint32_t l3_offset, rte_be16_t eth_type
 			}
 			break;
 		case IPPROTO_TCP:
-			if (l3.ip4->fragment_offset == 0) {
+			if (!frag) {
 				l4.tcp = rte_pktmbuf_mtod_offset(
```
christophefontaine, line 88: The previous commit and this one seems to have similarities with https://github.com/DPDK/grout/pull/701 
Could review this PR ?

maxime-leroy: You are right, thanks. The five flow hash commits are now a separate PR, #742.
This one still depends on it and shows them until it merges.

 #701 does the same extraction and fixes the same DF bug. I have written the
 overlap and where the two differ in the description of #742, and I will follow
 up with a review on #701 itself.

PR #745, opened 2026-09-11
Avoid excessive use of void *

MortenBroerup:
Grout is built with -Wstrict-aliasing=2, which requires type casting to the generic "void *", like this:
```
	struct conn *conn;
	// create a new connection object
	if (rte_mempool_get(conn_pool, (void *)&conn) < 0)
		return NULL;
```
Change to -Wstrict-aliasing=3, so we can use more accurate type casting, like this:
```
	struct conn *conn;
	// create a new connection object
	if (rte_mempool_get(conn_pool, (void **)&conn) < 0)
		return NULL;
```
Grout currently compiles just fine with -Wstrict-aliasing=3.

For reference, rte_mempool_get is declared as:
```
static __rte_always_inline int
rte_mempool_get(struct rte_mempool *mp, void **obj_p);
```

In addition to changing the compiler flags, improve the code as follows:

Replace all (void *) blanket casts with more appropriate ones.

Get rid of the now unnecessary intermediate void *data variables, since we can cast to (void **).

**Bonus**
Fixing the above for the l2 module revealed that the fdb_entry structure was declared in the implementation, but should be declared in the l2 header file, so the first patch in the series fixes this.


modules/infra/control/l3_nexthop.c
```diff
@@ -87,17 +87,17 @@ static inline void set_nexthop_key(
...
-	void *data;
+	struct nexthop *nh;
 
 	if (af == AF_UNSPEC)
 		return NULL;
 
 	set_nexthop_key(&key, af, vrf_id, iface_id, addr);
 
-	if (rte_hash_lookup_data(l3_hash, &key, &data) < 0)
+	if (rte_hash_lookup_data(l3_hash, &key, (void **)&nh) < 0)
```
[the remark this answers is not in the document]
MortenBroerup: Type conversion has to happen somewhere.
This:
```
struct sometype *value;
get((void **)&value);
use(value);
```
is better and shorter than using a temporary variable:
```
struct sometype *value;
void *temp;
get(&temp);
value = temp;
use(value);
```
Alternatively, we could take the long and silly road, and use a union:
```
union {
  struct sometype *value;
  void *temp_inner;
} temp_outer;
get(&temp_outer.temp_inner);
use(temp_outer.value);
```
Let's stick to the short version.

christophefontaine (approved):

modules/infra/control/vlan.c
```diff
@@ -25,12 +25,15 @@ struct vlan_key {
...
 
 struct iface *vlan_get_iface(uint16_t parent_id, uint16_t vlan_id) {
-	void *data;
+	struct iface *iface;
 
-	if (rte_hash_lookup_data(vlan_hash, &(struct vlan_key) {parent_id, vlan_id}, &data) < 0)
+	if (rte_hash_lookup_data(
+		    vlan_hash, &(struct vlan_key) {parent_id, vlan_id}, (void **)&iface
+	    )
+	    < 0)
```
rjarry: Just cosmetic, but could you add an inline variable to make the code less ugly? (thx clang-format :trollface:)

```c
struct iface *vlan_get_iface(uint16_t parent_id, uint16_t vlan_id) {
	const struct vlan_key key = {parent_id, vlan_id};
	struct iface *iface;

	if (rte_hash_lookup_data(vlan_hash, &key, (void **)&iface) < 0)
```

modules/infra/datapath/trace.c
```diff
@@ -842,12 +837,10 @@ int gr_trace_dump(
 ) {
 	uint32_t n = 0;
 	uint16_t p = 0;
-	void *data;
+	struct gr_trace_item *head;
 	int s;
```
rjarry, line 841: Reverse xmas tree please.

```diff
+	struct gr_trace_item *head;
	uint32_t n = 0;
	uint16_t p = 0;
-	void *data;
	int s;
```

modules/l2/control/fdb.c
```diff
@@ -327,16 +313,16 @@ static struct api_out fdb_flush(const void *request, struct api_ctx *) {
 	const struct gr_fdb_flush_req *req = request;
 	uint32_t next = 0;
 	const void *key;
-	void *data;
+	struct fdb_entry *fdb;
 	int ret;
```
rjarry, line 317: Reverse xmas tree please:

```diff
	const struct gr_fdb_flush_req *req = request;
+	struct fdb_entry *fdb;
	uint32_t next = 0;
	const void *key;
-	void *data;
	int ret;
```

modules/l2/control/vxlan.c
```diff
@@ -405,10 +405,9 @@ static int vtep_flood_list(uint16_t vrf_id, struct api_ctx *ctx) {
 	const struct iface_info_vxlan *vxlan;
 	uint32_t next = 0;
 	const void *key;
-	void *data;
+	struct iface *iface;
```
rjarry: Reverse xmas tree please:

```diff
	struct gr_flood_entry entry = {.type = GR_FLOOD_T_VTEP};
	const struct iface_info_vxlan *vxlan;
+	struct iface *iface;
	uint32_t next = 0;
	const void *key;
-	void *data;
```

rjarry (changes requested):

modules/infra/control/vlan.c
MortenBroerup: Will do.
I was wondering why a local variable was sometimes used, and sometimes the struct was built and passed directly as the function parameter.
Your comment here explains it!
:-)

rjarry (approved):

modules/infra/datapath/trace.c
MortenBroerup: Done.

modules/l2/control/fdb.c
MortenBroerup: Done.

modules/l2/control/vxlan.c
MortenBroerup: Done.

modules/infra/control/l3_nexthop.c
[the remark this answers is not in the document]
MortenBroerup: This is intended, and the purpose of the patch.

rjarry (approved):

PR #748, opened 2026-09-14
bench: add performance test scenarios

rjarry:
Add a `bench/` directory holding self-contained performance test scenarios for grout. Each scenario is a pair of files sharing a base name: a `grcli` configuration applied on top of a partially configured grout (CPU affinity and the two physical ports `p0`/`p1` already created), and a matching TRex stateless traffic profile. Together they fully describe what grout and a back-to-back TRex must do, so an external harness can discover them, configure grout, load the profile and run a bidirectional RFC 2544 throughput search.

Keep everything static with hardcoded MAC addresses, so the scenarios run without FRR and without ARP or NDP. This keeps them independent of the physical NIC addresses and prevents a top-of-rack switch from getting confused.

Cover plain IPv4 and IPv6 forwarding, a dual-stack mix, 802.1Q VLAN, an IMIX distribution, the SRv6 datapath (transit `End`, L3VPN with `H.Encaps` and `End.DT4`, reduced `H.Encaps.Red`, and uSID `uDT4`) and VXLAN in bridged and routed modes over both IPv4 and IPv6 underlays.

Since the traffic generation harness lives elsewhere, also ship a small dummy harness under `bench/harness/` that starts grout in test mode, creates the two ports as taps and replays a profile with scapy through a mock of the TRex API. It only checks that traffic is forwarded, but it lets a scenario be smoke-tested locally without TRex (`make bench-validate`).

Include one small CLI change as an enabler: `fdb add` gains an optional `vtep` address, so static VXLAN L2VPN forwarding entries can be created without relying on learning.


bench/srv6_l3vpn.py
```diff
@@ -0,0 +1,63 @@
...
+#
+# Asymmetric SRv6 L3VPN profile (modeled on smoke/srv6_test.sh):
+#   direction 0 (TRex port 0 -> grout p0): plain IPv4 packet toward the remote
+#     site; grout H.Encaps it into SRv6 and sends the IPv6 packet out p1;
+#   direction 1 (TRex port 1 -> grout p1): SRv6 packet destined to grout's local
+#     End.DT4 SID with an inner IPv4 packet; grout decapsulates and routes the
+#     inner IPv4 out p0.
+#
+# Ingress direction 0 is a 64-byte IPv4 frame; ingress direction 1 is
+# Ether(14) + IPv6(40) + SRH(8 + 16) + IPv4(20) + UDP(8) = 106 bytes.
```
aharivel, line 14: The code only show this:
Ether(14) + IPv6(40) + IPv4(20) + UDP(8) = 82 bytes

GNUmakefile
```diff
@@ -74,6 +74,25 @@ $(smoke_scripts):
...
+		else \
+			fail=$$((fail + 1)); \
+			failed="$$failed $$name"; \
+		fi; \
+	done; \
+	echo "PASS: $$pass FAIL: $$fail"; \
+	if [ -n "$$failed" ]; then \
+		echo "FAILED: $$failed"; \
+	fi; \
+	exit "$$fail"
```
aharivel, line 94: exit $$(( fail > 0 )) is a bit more secure even if I don't know if we are going to have more than 125 scenarios :D

aharivel (approved):

